PR #1, opened 2018-01-06
Feature/some refactoring

guusvw:


mrIncompetent:
@guusvw rebase please

Has this been tested?

guusvw:
(Rebase )Will do it and I will also write an test for `EnsureSSHKeypairSecret` to ensure the same behaviour as before.

mrIncompetent:
We indeed are missing some tests

pkg/client/clientset/versioned/scheme/register.go
```diff
@@ -24,13 +24,15 @@ import (
...
+)
 
 func init() {
-	v1.AddToGroupVersion(Scheme, schema.GroupVersion{Version: "v1"})
-	AddToScheme(Scheme)
+	v1.AddToGroupVersion(scheme, schema.GroupVersion{Version: "v1"})
+	AddToScheme(scheme)
 }
 
 // AddToScheme adds all types of this clientset into the given scheme. This allows composition
```
mrIncompetent: everything under pkg/client is generated code

pkg/ssh/configmap.go
```diff
@@ -20,48 +20,59 @@ const (
 	secretName = "machine-controller-ssh-key"
 )
 
+// EnsureSSHKeypairSecret
 func EnsureSSHKeypairSecret(client kubernetes.Interface) (*rsa.PrivateKey, error) {
+	if client == nil {
+		return nil, fmt.Errorf("got an nil k8s client")
+	}
```
mrIncompetent: Do we really wan to do a nil check?

pkg/ssh/configmap.go
```diff
@@ -20,48 +20,59 @@ const (
...
-	return keyFromSecret(secret)
 }
 
 func keyFromSecret(secret *v1.Secret) (*rsa.PrivateKey, error) {
 	b, exists := secret.Data[privateKeyDataIndex]
 	if !exists {
 		return nil, fmt.Errorf("key data not found in secret '%s/%s' (secret.data['%s']). remove it and a new one will be created", secret.Namespace, secret.Name, privateKeyDataIndex)
 	}
+	if len(b) == 0 {
+		return nil, fmt.Errorf("key data not found in secret '%s/%s' (secret.data['%s']). remove it and a new one will be created", secret.Namespace, secret.Name, privateKeyDataIndex)
```
mrIncompetent: i would rather state something like `invalid key data found`

pkg/ssh/configmap.go
guusvw: Yes ;) make the code more stable

pkg/ssh/configmap.go
guusvw: fixed

PR #3, opened 2018-01-10
add circle ci config

guusvw:


.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+  build_and_test:
+    jobs:
+      - checkout_code:
+          filters:
+            tags:
+              only: /v.*/
+      - install-dependencies:
+          requires:
+            - checkout_code
+          filters:
```
mrIncompetent: not sure if this filter is needed. we used it at kube-machine, but no idea why

.circleci/config.yml
```diff
@@ -0,0 +1,69 @@
...
+      - run: go build -o machine-controller cmd/controller/main.go
+
+workflows:
+  version: 2
+  build_and_test:
+    jobs:
+      - checkout_code:
+          filters:
+            tags:
+              only: /v.*/
```
mrIncompetent: not sure if this filter is needed. we used it at kube-machine, but no idea why

PR #18, opened 2018-01-13
fix some linting issues

mrIncompetent:


pkg/controller/machine.go
```diff
@@ -424,7 +426,11 @@ func (c *Controller) updateMachineStatus(machine *machinev1alpha1.Machine, node
 }
 
 func parseContainerRuntime(s string) (runtime, version string, err error) {
-	re, _ := regexp.Compile(`(\w*)://(.*)`)
+	regex := `(\w*)://(.*)`
+	re, err := regexp.Compile(regex)
```
metalmatze: This is one of the rare cases where I actually put things into globals. This way they won't be compiled every time the func is called.

guusvw: yeah and in this case you can use `regexp.MustCompile` and assign the result in an var-block

pkg/controller/machine.go
```diff
@@ -455,16 +461,35 @@ func (c *Controller) getNode(instance instance.Instance, provider string) (node
...
+	}
+	newJSON, err := json.Marshal(newMachine)
+	if err != nil {
+		return fmt.Errorf("failed to marshal updated machine to json: %v", err)
+	}
+	patch, err := jsonmergepatch.CreateThreeWayJSONMergePatch(oldJSON, newJSON, currentJSON)
+	if err != nil {
+		return fmt.Errorf("failed to create three-way-json-merge-patch: %v", err)
+	}
+	if string(patch) == "{}" {
```
guusvw: `"{}"` would be nice as a const like `const empyJSON`

PR #19, opened 2018-01-14
use new to create instances of config

guusvw:


pkg/userdata/coreos/userdata.go
```diff
@@ -21,15 +21,15 @@ type config struct {
 }
 
 func getConfig(r runtime.RawExtension) (*config, error) {
-	p := config{}
+	p := new(config)
```
metalmatze: I'd really discourage you from this. New hides the fact that a points is created. If you want to do this then make it explicitly: `p := &config{}`.

pkg/userdata/coreos/userdata.go
```diff
@@ -21,15 +21,15 @@ type config struct {
...
 	if len(r.Raw) == 0 {
-		return &p, nil
+		return p, nil
 	}
 
-	if err := json.Unmarshal(r.Raw, &p); err != nil {
+	if err := json.Unmarshal(r.Raw, p); err != nil {
```
metalmatze: This also decreases readability with the fact that you pass a pointer. The old version displays this very clearly... :confused:

pkg/userdata/coreos/userdata.go
guusvw: `new` has a defined behaviour, it always returns a value of type `*T`, from my point of view, it's way more clearer than `p := &config{}`

guusvw: Both are valid solutions

metalmatze: Yes. Both are valid solutions. The `&config{}` one seems to be used by more people in the community. To summarize this PR: I wouldn't create a pointer in this func right away, especially due to the fact that we can more explicitly pass the pointer to `json.Unmarshal` :relaxed:

guusvw:
Discussed it on slack and decide to go the non `new`-way

PR #23, opened 2018-01-22
test get node

mrIncompetent:


pkg/controller/machine_test.go
```diff
@@ -0,0 +1,167 @@
...
+	node3 = getTestNode("3", "")
+
+	nodeList = &corev1.NodeList{
+		Items: []corev1.Node{
+			node1,
+			node2,
+			node3,
+		},
+	}
+)
```
metalmatze: Correct me if I'm wrong, but this can be inlined into the `TestController_GetNode` func.

PR #26, opened 2018-01-24
Add docker version support

mrIncompetent:


pkg/userdata/ubuntu/docker.go
```diff
@@ -0,0 +1,45 @@
+package ubuntu
+
+import "errors"
+
+func getDockerInstallCandidate(desiredVersion string) (pkg string, version string, err error) {
```
metalmatze: I don't see a reason to have these named return values. Is there one?

mrIncompetent: I prefer named returns when using more than 2 return values. 
Otherwise its not obvious what return parameter is what

metalmatze: I think in that case it's better to create a decent comment for the func. IMO it's maybe better a the declaration of the func, but not in the code itself further down, when or what is returned. It's not that bad in this case, because you only have one return and give it the return types explicitly. 
So I'm good with merging this.

PR #28, opened 2018-01-24
Fix check for available Openstack floating IPs

### LINKED ISSUE #21: Openstack: Floating IPs are not reused which may result in FIP exhaustion
Basically title, from the the machine controller log:

```
E0120 13:47:35.431740       1 machine.go:162] machine-controller failed with: failed to create machine at cloudprovider: failed to allocate a floating ip: Expected HTTP response code [201 202] when accessing [POST http://192.168.0.39:9696/v2.0/floatingips], but got 409 instead
{"NeutronError": {"message": "No more IP addresses available on network 06fb6e98-4e98-4320-9f00-34e028ed53cb.", "type": "IpAddressGenerationFailure", "detail": ""}}
```

I'd expect the machine-controller to reuse already assigned but unused FIPs instead of requesting a new one.

alvaroaleman:
Resolves #21

### DIFF
--- a/pkg/cloudprovider/provider/openstack/helper.go
+++ b/pkg/cloudprovider/provider/openstack/helper.go
@@ -30,6 +30,8 @@ var (
 // Protects creation of public key
 var publicKeyCreationLock = &sync.Mutex{}
 
+const openstackFloatingIPErrorStatusName = "ERROR"
+
 func getRegion(client *gophercloud.ProviderClient, id string) (*osregions.Region, error) {
 	idClient, err := goopenstack.NewIdentityV3(client, gophercloud.EndpointOpts{Availability: gophercloud.AvailabilityPublic})
 	if err != nil {
@@ -302,7 +304,7 @@ func getFreeFloatingIPs(client *gophercloud.ProviderClient, region string, float
 
 	var freeFIPs []osfloatingips.FloatingIP
 	for _, f := range allFIPs {
-		if f.Status == "ACTIVE" && f.PortID == "" {
+		if f.Status != openstackFloatingIPErrorStatusName && f.PortID == "" {
 			freeFIPs = append(freeFIPs, f)
 		}
 	}


pkg/cloudprovider/provider/openstack/helper.go
```diff
@@ -302,7 +302,7 @@ func getFreeFloatingIPs(client *gophercloud.ProviderClient, region string, float
 
 	var freeFIPs []osfloatingips.FloatingIP
 	for _, f := range allFIPs {
-		if f.Status == "ACTIVE" && f.PortID == "" {
+		if f.Status == "DOWN" && f.PortID == "" {
```
mrIncompetent: I'm not sure about this one.
When does a floatingip get the status "DOWN"? 
In the installation i have access to, the IP stays active but its just not assigned to something

guusvw: I don't know the correct behaviour, I trust you ;) but I would love to see the different states as a `const` - that would be amazing

alvaroaleman: So, the Openstack docs are not very clear about this, they only tell you about what status exist but not what they mean[0].
According to the change that introduced the floating ip status field[1], the default status is "DOWN":
> By the way, DOWN is the default status for a new floating IP[...]

In my(tm) Openstack cloud, FIPs that are allocated but unassigned have the status DOWN:
```bash
openstack floating ip list --status DOWN
+--------------------------------------+---------------------+------------------+------+--------------------------------------+----------------------------------+
| ID                                   | Floating IP Address | Fixed IP Address | Port | Floating Network                     | Project                          |
+--------------------------------------+---------------------+------------------+------+--------------------------------------+----------------------------------+
| 3578e95f-cfb7-461c-8802-0415d7673f1b | 192.168.0.79        | None             | None | 06fb6e98-4e98-4320-9f00-34e028ed53cb | 68625d66215f4830a8b197c6b8dca2ef |
```

And FIPs that are allocated and assigned have the status ACTIVE:

```bash
openstack floating ip list --status ACTIVE
+--------------------------------------+---------------------+------------------+--------------------------------------+--------------------------------------+----------------------------------+
| ID                                   | Floating IP Address | Fixed IP Address | Port                                 | Floating Network                     | Project                          |
+--------------------------------------+---------------------+------------------+--------------------------------------+--------------------------------------+----------------------------------+
| 3ba400b6-64be-49a4-aab0-cc34bd928714 | 192.168.0.82        | 10.0.0.3         | 763ab121-6525-42af-9778-1b99c98d46e4 | 06fb6e98-4e98-4320-9f00-34e028ed53cb | 68625d66215f4830a8b197c6b8dca2ef |
```

If your cloud behaves differently, maybe we should alter the condition to `if f.Status != "ERROR" && f.PortID == "" {`?

[0]https://developer.openstack.org/api-ref/network/v2/#floating-ips-floatingips
[1]https://review.openstack.org/#/c/66866/

mrIncompetent: Yes, then i would say: `if f.Status != "ERROR" && f.PortID == "" {`
Thanks for the docs!

pkg/cloudprovider/provider/openstack/helper.go
```diff
@@ -302,7 +304,7 @@ func getFreeFloatingIPs(client *gophercloud.ProviderClient, region string, float
 
 	var freeFIPs []osfloatingips.FloatingIP
 	for _, f := range allFIPs {
-		if f.Status == "ACTIVE" && f.PortID == "" {
+		if f.Status != openstackFloatingIPErrorStatusName && f.PortID == "" {
```
mrIncompetent: nit: as you are already in the go package "openstack" theres no need to prefix the constant with "openstack*"
I'll change it upstream

PR #29, opened 2018-01-24
Expose public address of Openstack machines

alvaroaleman:


mrIncompetent:
There are two reasons why i'm against this PR:
- We cannot modify the machine struct
  - The struct comes from the machine-api discussion from: https://github.com/kubernetes/kube-deploy/pull/298
  - except for the `machine.spec.providerConfig` its equal. (The providerConfig has a different type and the kube-deploy version will adapt soon. See https://github.com/kubernetes/kube-deploy/issues/504)
  - If you have good reasons why `addresses` should be added to the `machine.status` please create an issue in https://github.com/kubernetes/kube-deploy
- The instance interface already provides a function to return the addresses. The primary usage is to find the matching k8s node. I would rather add the floatingip address to the addresses returned there.
Especially because in some environments we don't know actually if the ip is publicly available or not - i think some aws nodes might not be public at all.

pkg/controller/machine.go
```diff
@@ -485,6 +499,21 @@ func (c *Controller) patchMachine(newMachine, oldMachine *machinev1alpha1.Machin
...
+	if err != nil {
+		fmt.Errorf("Failed to create a map from mergepatch: '%v'", err)
+	}
+	if _, ok := patchMap["metadata"]; ok {
+		delete(patchMap, "metadata")
+		patch, err = json.Marshal(patchMap)
+		if err != nil {
+			return fmt.Errorf("failed to marshal patch with removed metadata: '%v'", err)
+		}
+	}
```
mrIncompetent: I've seen that some patches fail due to unstructured data send to the api.
Though i think this here will break the Finalizer handling.

I'm not sure whats the status on StrategicMergePatch and CRD's. I'll check

pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -318,6 +318,11 @@ func (d *doInstance) ID() string {
 	return strconv.Itoa(d.droplet.ID)
 }
 
+func (d *doInstance) PublicAddress() string {
+	//TODO: Implement this
+	return "Not implemented"
```
guusvw: As long as we don't return an error here, can you do a 

`panic("not implemented")`

just in case?

alvaroaleman:
To give some context, the underlying reason is that right now it is not possible to get the address of a node which is behind a NAT from the k8s api. IMHO that is a useful info for debugging. What do you generally think about having a feature for that in the machine-controller? I just found the machine spec the most intuitive place, but e.G. a simple node annotation would work as well.

The reason for not using the existing `Addresses()` part of the instance interface is that it returns a slice which contains in case of NAT at least two addresses, so I have to guess which one to connect to.

pkg/controller/machine.go
alvaroaleman: It does indeed break the finalizer. Not sure what the expected behaviour is, but as long as the patch contained a metadata attribute with just a `resourceVersion` property, patching failed with a 
```
E0124 21:57:53.314219       1 status.go:64] apiserver received an error that is not an metav1.Status: unable to find api field in struct Unstructured for the json field "metadata"
```
from the api server.

pkg/cloudprovider/provider/digitalocean/provider.go
alvaroaleman: Hm, don't you think killing off the whole controller just because of that would be a bit over the top? Maybe it would be smarter to return an empty string, thought.

pkg/controller/machine.go
mrIncompetent: I see.
I'll test if we are safe to use update (put) instead

alvaroaleman:
Clarified in Slack that this not actually needed because the kubelet already exposes the floating IP when the cloud provider is configured correctly, thus closing.

PR #31, opened 2018-01-26
use index to to fix machine loading

mrIncompetent:


mrIncompetent:
force merge, as i only rebased

PR #33, opened 2018-01-27
fix error when no install candidate can be found for the requested cri-o version

mrIncompetent:
cc @alvaroaleman


alvaroaleman:
:+1:

PR #34, opened 2018-01-27
Improve crio usability

alvaroaleman:


pkg/userdata/ubuntu/userdata.go
```diff
@@ -129,7 +129,7 @@ write_files:
     set -xeuo pipefail
     mkdir -p /opt/bin /opt/cni/bin /etc/cni/net.d /var/run/kubernetes /var/lib/kubelet /etc/kubernetes/manifests /var/log/containers
     if [ ! -f /opt/bin/kubelet ]; then
-      curl -L -o /opt/bin/kubelet https://storage.googleapis.com/kubernetes-release/release/v{{ .MachineSpec.Versions.Kubelet }}/bin/linux/amd64/kubelet
+      curl -L -o /opt/bin/kubelet https://storage.googleapis.com/kubernetes-release/release/{{ .MachineSpec.Versions.Kubelet }}/bin/linux/amd64/kubelet
```
mrIncompetent: We should parse the entered version and accept both from the user:
- `v1.9.2`
- `1.9.2`

I think using a samver library should solve that. I can do a separate pr afterwards

pkg/userdata/ubuntu/userdata.go
```diff
@@ -116,7 +116,7 @@ ssh_authorized_keys:
 write_files:
 - path: "/etc/kubernetes/cloud-config"
   content: |
-{{ .CloudConfig | indent 4 }}
+{{ if ne .CloudConfig "" }}{{ .CloudConfig | indent 4 }}{{ end }}
```
mrIncompetent: Beautiful!

pkg/userdata/ubuntu/crio.go
```diff
@@ -7,6 +7,9 @@ func getCRIOInstallCandidate(desiredVersion string) (pkg string, version string,
 	case "1.9", "1.9.0":
 		pkg = "cri-o"
 		version = "1.9.0-1~ubuntu16.04.2~ppa1"
+	case "":
+		pkg = "cri-o"
+		version = "1.9.0-1~ubuntu16.04.2~ppa1"
```
mrIncompetent: As mentioned in slack, we should have a dedicated logic which will find the best container-runtime version for the given kubernetes version.
But can be done in a separate pr as well. For now that is alright

pkg/userdata/ubuntu/userdata.go
mrIncompetent: issue: #36

pkg/userdata/ubuntu/crio.go
mrIncompetent: issue: #35

PR #37, opened 2018-01-29
add version defaulting for the container runtime

### LINKED ISSUE #35: Make container runtime version optional
Based upon the entered Kubernetes version and the selected OS we should default to a docker/cri-o version.

For now the logic should be:
- cri-o
  - Kubernetes v1.8 + Ubuntu 16.04 -> error as theres no cri-o 1.8 package in the repos
  - Kubernetes v1.9 + Ubuntu 16.04 -> cri-o 1.9
  - Kubernetes v1.8 + Container Linux -> error as theres no cri-o for coreos
  - Kubernetes v1.8 + Container Linux -> error as theres no cri-o for coreos
- docker
  - Kubernetes v1.8 + Ubuntu 16.04 -> docker 1.13
  - Kubernetes v1.9 + Ubuntu 16.04 -> docker 1.13
  - Kubernetes v1.8 + Container Linux -> docker 1.12
  - Kubernetes v1.8 + Container Linux -> docker 1.12

mrIncompetent:
Adds sane defaulting for the container runtime.
Also it handles versions defined with and without `v` as prefix.

fixes #35

cc @alvaroaleman


docs/container_runtime.md
```diff
@@ -0,0 +1,75 @@
+# Container Runtimes
+
+## Defaulting
+The machine-controller is able to default to a supported container runtime in case no runtime was specified in the machine-spec.
+Also in case only the version is missing only that will get defaulted.
```
metalmatze: What is this sentence supposed to say?

pkg/containerruntime/crio/supported_versions.go
```diff
@@ -0,0 +1,26 @@
...
+
+// GetOfficiallySupportedVersions returns the officially supported cri-o version for the given kubernetes version
+func GetOfficiallySupportedVersions(kubernetesVersion string) ([]string, error) {
+	v, err := semver.NewVersion(kubernetesVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	majorMinorString := fmt.Sprintf("%d.%d", v.Major(), v.Minor())
+	switch majorMinorString {
```
metalmatze: Wouldn't be a `strings.HasPrefix(kubernetesVersion, "1.8")` be sufficient? But I don't have a problem with using this lib.

docs/container_runtime.md
mrIncompetent: You can simply define `machine.Spec.ContainerRuntime.Name` without defining `machine.Spec.ContainerRuntime.Version`

mrIncompetent: I'll clarify that sentence

pkg/containerruntime/crio/supported_versions.go
mrIncompetent: It prevents breaking the code when someone enters `v1.9.2` or `1.9.2`

pkg/errors/errors.go
```diff
@@ -0,0 +1,8 @@
+package errors
+
+import "errors"
+
+var (
+	NoSupportedVersionAvailableErr = errors.New("no supported version available")
+	VersionNotAvailableErr         = errors.New("version not available")
+)
```
metalmatze: Can't we find a more well suited place for this file to be placed in? I'm hesitant with such files, as they just introduce more little helper packages. Either this error is returned by a package and also in that package, or it is consumed by a package and inside that package. 

For example `VersionNotAvailableErr` is only used in the `package ubuntu` so place it there.

pkg/containerruntime/crio/supported_versions.go
metalmatze: Fair enough. Then using that lib is probably better then starting your own regex magic. :sparkles:

pkg/errors/errors.go
mrIncompetent: fixed in 3e8b884

PR #39, opened 2018-01-30
Unique ssh key name

mrIncompetent:
Currently we create a public ssh key on the cloud providers with the same name.
As this breaks when you have multiple clusters, we need to make the name customizable.

### DIFF
--- a/pkg/cloudprovider/provider.go
+++ b/pkg/cloudprovider/provider.go
@@ -8,23 +8,26 @@ import (
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider/provider/digitalocean"
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider/provider/openstack"
 	"github.com/kubermatic/machine-controller/pkg/providerconfig"
+	machinessh "github.com/kubermatic/machine-controller/pkg/ssh"
 )
+
+type providerInitializer = func(key *machinessh.PrivateKey) cloud.Provider
 
 var (
 	// ErrProviderNotFound tells that the requested cloud provider was not found
 	ErrProviderNotFound = errors.New("cloudprovider not found")
 
-	providers = map[providerconfig.CloudProvider]cloud.Provider{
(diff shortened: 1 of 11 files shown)


pkg/cloudprovider/provider.go
```diff
@@ -8,23 +8,26 @@ import (
...
+
 var (
 	// ErrProviderNotFound tells that the requested cloud provider was not found
 	ErrProviderNotFound = errors.New("cloudprovider not found")
 
-	providers = map[providerconfig.CloudProvider]cloud.Provider{
-		providerconfig.CloudProviderDigitalocean: digitalocean.New(),
-		providerconfig.CloudProviderAWS:          aws.New(),
-		providerconfig.CloudProviderOpenstack:    openstack.New(),
+	providers = map[providerconfig.CloudProvider]providerInitializer{
```
metalmatze: Nitpick: You define `providerInitializer` as new type and then use it in exactly one place. Additionally one line later you simply create funcs with its signature.

What about deleting the type and simply use?
```diff
- providers = map[providerconfig.CloudProvider]providerInitializer{
+ providers = map[providerconfig.CloudProvider]func(key *machinessh.PrivateKey) cloud.Provider{
```

I guess I would do it like:
```go
var providers = make(map[providerconfig.CloudProvider]func(key *machinessh.PrivateKey) cloud.Provider)
providers[providerconfig.CloudProviderDigitalocean] = func(key *machinessh.PrivateKey) cloud.Provider {
	return digitalocean.New(key)
}
...
```

Not 100% sure though :thinking:

mrIncompetent: Yeah, thought about this as well, i thought declaring it in the map might be a bit confusing.
But it doesnt look bad what you wrote.
I'll adapt!

mrIncompetent: I decied to simply remove the interface and stick with the static map.
Setting the keys using your provided option seemed a bit noisy - creating 3 lines for each cloud-provider

PR #40, opened 2018-01-31
Use PUT instead of PATCH to eliminate the api errors

mrIncompetent:
We got errors from the apiserver on some PATCH requests.
As we are the only controller operating on machines, we can use PUT instead.


mrIncompetent:
cc @alvaroaleman as mentioned in https://github.com/kubermatic/machine-controller/pull/29

PR #42, opened 2018-01-31
make cluster dns flag on the kubelet configurable

mrIncompetent:
Make `cluster-dns` flag on the kubelet configurable

### DIFF
--- a/pkg/template/functions.go
+++ b/pkg/template/functions.go
@@ -9,11 +9,6 @@ import (
 	"github.com/Masterminds/sprig"
 )
 
-// FuncMap defines the available functions to kubermatic templates.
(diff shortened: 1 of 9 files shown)


pkg/template/functions.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+	"net"
+	"strings"
+	"text/template"
+
+	"github.com/Masterminds/sprig"
+)
+
+// FuncMap defines the available functions to kubermatic templates.
+var funcs = template.FuncMap{
```
metalmatze: This can be inlined into `TxtFuncMap` again?

mrIncompetent: good point

Issue #47, opened 2018-02-01
simple e2e test tool


p0lyn0mial:
having a simple command line tool that would verify whether a node has been created serves not only as a good warm up exercise but also as a handy test tool.

the idea is that we would have a list of predefined machine manifests that would need some customisation in terms of credentials. The credentials could be accepted as a command line arguments and passed all the way down to the manifests. After POST'ing the given manifests to the `kube-api` server the test tool would read the current cluster state in order to determine the correctness of `machine-controller`

the test tool would use the standard `client-go` library to talk to the `api` server and would read the `kubeconfig` configuration file to discover where the cluster is actually located.

assumptions:
- cluster was created manually
- `kube config` is accessible
- there is a list of predefined machine manifests

for example, running the following command: `verify -input path_to_manifest -parameters key=value, key2=value` would print `a machine "node-docker" has been crated` to stdout.

mrIncompetent:
lgtm

PR #51, opened 2018-02-02
initial scaffolding of the test tool app.

p0lyn0mial:
this PR introduces the initial scaffolding of the test tool app. At the moment it reads command line arguments and modifies the given manifest accordingly. 

**Which issue(s) this PR fixes** : see more: #47

### DIFF
--- a/test/tools/verify/cmd/main.go
+++ b/test/tools/verify/cmd/main.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+func main() {
+	var manifestPath string
+	var parameters string
+
+	flag.StringVar(&manifestPath, "input", "", "a path to the machine's manifest.")
+	flag.StringVar(&parameters, "parameters", "", "a list of comma-delimited key value pairs i.e key=value,key1=value2")
+	flag.Parse()
+
+	if len(manifestPath) == 0 || len(parameters) == 0 {
+		fmt.Println("please specify input and parameters flags")
+		flag.PrintDefaults()
+		os.Exit(1)
+	}
+	keyValuePairs := strings.Split(parameters, ",")
+	if len(keyValuePairs) == 0 {
+		fmt.Println("incorrect value of parameters flag")
+		flag.PrintDefaults()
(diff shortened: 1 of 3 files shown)


test/tools/verify/cmd/main.go
```diff
@@ -0,0 +1,55 @@
...
+		return "", err
+	}
+	content := fmt.Sprintf("%s", contentRaw)
+
+	for _, keyValuePair := range keyValuePairs {
+		kv := strings.Split(keyValuePair, "=")
+		if len(kv) != 2 {
+			continue
+		}
+		content = strings.Replace(content, kv[0], kv[1], -1)
```
mrIncompetent: Can we use golang templating here?

test/tools/verify/cmd/main.go
```diff
@@ -0,0 +1,55 @@
...
+	contentRaw, err := ioutil.ReadFile(pathToManifest)
+	if err != nil {
+		return "", err
+	}
+	content := fmt.Sprintf("%s", contentRaw)
+
+	for _, keyValuePair := range keyValuePairs {
+		kv := strings.Split(keyValuePair, "=")
+		if len(kv) != 2 {
+			continue
```
p0lyn0mial: alternatively we could fail on error

test/tools/verify/cmd/main.go
p0lyn0mial: I think we could. But doesn't that mean we would have to maintain a separate struct with predefined fields ? As the golang's template engine applies a template to the given data object (template.Execute). Given the fact that each manifest has slightly different set of dynamic fields replacing a substring in the given strings seems to be much easier. What do you think :) ?

mrIncompetent: Very good point. Using a simple string replace seems better!

test/tools/verify/cmd/main.go
mrIncompetent: I would prefer failing early

p0lyn0mial: sounds good to me.

p0lyn0mial:
@mrIncompetent comments addressed.

p0lyn0mial:
yay!

PR #60, opened 2018-02-04
Use run.Group to manage main components’ lifecycle

metalmatze:
Right now we're running everything in goroutines from the main. Some parts are not really able to shutdown gracefully. Using github.com/oklog/run enables us to shutdown then internal http handler gracefully too. 
In the future its usage can be extended.

Let's discuss!

As I am on my mobile phone's internet I did not  yet run `dep ensure add` for both new libs.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"net"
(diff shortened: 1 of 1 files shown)


cmd/controller/main.go
```diff
@@ -34,6 +35,8 @@ import (
 	"github.com/kubermatic/machine-controller/pkg/machines"
 	"github.com/kubermatic/machine-controller/pkg/signals"
 	"github.com/kubermatic/machine-controller/pkg/ssh"
+	"github.com/oklog/run"
+	"github.com/pkg/errors"
```
mrIncompetent: why are we introducing this library?
Whats the benefit of this compared against a simple `fmt.Errorf` ?

mrIncompetent:
@metalmatze are you ok with closing this PR?

PR #63, opened 2018-02-05
implements verify method

p0lyn0mial:
implements verify method, at the moment this includes checking the cluster state before and after a new machine has been created.

**Which issue(s) this PR fixes** : see more #47

### DIFF
--- a/test/tools/verify/cmd/main.go
+++ b/test/tools/verify/cmd/main.go
@@ -1,14 +1,25 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	machineclientset "github.com/kubermatic/machine-controller/pkg/client/clientset/versioned"
+	machinev1alpha1 "github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
 	"io/ioutil"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 	"os"
 	"strings"
+	"time"
+)
+
+const (
+	machineReadyCheckPeriod  = 2 * time.Second
+	machineReadyCheckTimeout = 5 * time.Minute
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


test/tools/verify/cmd/main.go
```diff
@@ -51,19 +57,64 @@ func main() {
...
+		manifestReader := strings.NewReader(manifest)
+		manifestDecoder := yaml.NewYAMLToJSONDecoder(manifestReader)
+		err := manifestDecoder.Decode(newMachine)
+		if err != nil {
+			return err
+		}
+	}
+
+	// we expect to find no nodes within the cluster
+	err := assureNodeCount(0, kubeClient)
```
p0lyn0mial: this is a assumption I would like to have, no nodes within the cluster, what do you think ?

test/tools/verify/cmd/main.go
```diff
@@ -51,19 +57,64 @@ func main() {
...
+		if err == nil {
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		return fmt.Errorf("falied to created the new machine, err = %v", err)
+	}
+
+	// TODO(lukasz): implement clean-up logic
```
p0lyn0mial: I'd like to clean up the mess we have made. Bonus we could get is that we could validate if our controller knows how to remove unwanted machines.

test/tools/verify/cmd/main.go
mrIncompetent: I agree, we should expect 0 nodes within a cluster

test/tools/verify/cmd/main.go
mrIncompetent: :+1:

test/tools/verify/cmd/main.go
```diff
@@ -51,19 +57,64 @@ func main() {
...
+	// TODO(lukasz): add dep
```
p0lyn0mial: this is optional, not sure if we want to add dependencies.

mrIncompetent: Could you explain what you mean exactly?
What kind of dependencies?

p0lyn0mial: oh, I was thinking that we could use the `dep` tool. So that we could lock imported(k8s) stuff in.

mrIncompetent: Maybe then we should move the test command to https://github.com/kubermatic/machine-controller/tree/master/cmd ?
Then we would use the same vendor as the controller itself

test/tools/verify/cmd/main.go
```diff
@@ -51,19 +57,64 @@ func main() {
...
+	}
+	err = wait.Poll(500*time.Millisecond, 60*time.Second, func() (bool, error) {
+		err := assureNodeCount(1, kubeClient)
+		if err == nil {
+			return true, nil
+		}
+		return false, nil
+	})
+	if err != nil {
+		return fmt.Errorf("falied to created the new machine, err = %v", err)
```
p0lyn0mial: at this point we could grab the machine's `Status` from the  kube-api server  and put it into the logs. That would facilitate troubleshooting.

test/tools/verify/cmd/main.go
p0lyn0mial: sounds good to me.

p0lyn0mial:
@mrIncompetent is there anything that block this pull? I will follow up clean-up, moving the binary to different directory and printing the `Status` to logs.

test/tools/verify/cmd/main.go
```diff
@@ -51,19 +57,64 @@ func main() {
...
+	if err != nil {
+		return fmt.Errorf("unable to perform the verification, incorrect cluster state detected %v", err)
+	}
+
+	fmt.Printf("creating a new \"%s\" machine\n", newMachine.Name)
+	_, err = machineClient.MachineV1alpha1().Machines().Create(newMachine)
+	if err != nil {
+		return err
+	}
+	err = wait.Poll(500*time.Millisecond, 60*time.Second, func() (bool, error) {
```
mrIncompetent: Should be moved into constants.
Also the timeout is too low. It can take up to 5min

p0lyn0mial: good point.

p0lyn0mial:
@mrIncompetent comments addressed.

p0lyn0mial:
@mrIncompetent just a rebase.

PR #66, opened 2018-02-05
print out statuses to facilitate troubleshooting

p0lyn0mial:
**What this PR does / why we need it**: prints out statuses to facilitate troubleshooting
**Which issue(s) this PR fixes**: #47


p0lyn0mial:
@mrIncompetent that was super quick !

PR #67, opened 2018-02-05
Build docker image in circleci pipeline

### LINKED ISSUE #57: Extend circle pipeline
Whats missing:
Building a docker image

- On push using the commit hash as docker tag
- On tag using the git tag as docker tag & latest

alvaroaleman:

Automatically publish a docker image via CircleCI

Fixes #57 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,14 +59,15 @@ jobs:
           go/src/github.com/kubermatic/machine-controller &&
           sudo mkdir -p /go/src/github.com/kubermatic/machine-controller && sudo chmod 777 -R
           /go/src/github.com/kubermatic/machine-controller &&
-          sudo chown -R circleci:circleci /go/src/github.com/kubermatic/machine-controller
+          sudo chown -R circleci:circleci /go/src/github.com/kubermatic/machine-controller &&
+          cd /go/src/github.com/kubermatic/machine-controller
       - restore_cache:
           key: machine-controller-{{ .Revision }}
       - restore_cache:
           key: vendor-{{ checksum "Gopkg.lock" }}
       - restore_cache:
           key: repo-{{ .Environment.CIRCLE_SHA1 }}
-      - run: cd /go/src/github.com/kubermatic/machine-controller && make docker-image
+      - run: make docker-image
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,7 +62,11 @@ jobs:
           key: machine-controller-{{ .Revision }}
       - restore_cache:
           key: vendor-{{ checksum "/go/src/github.com/kubermatic/machine-controller/Gopkg.lock" }}
-      - run: cd /go/src/github.com/kubermatic/machine-controller && make docker-image-nodep
+      - run: >
+          cd /go/src/github.com/kubermatic/machine-controller &&
+          make docker-image-nodep &&
+          docker login -u $DOCKERHUB_USER -p $DOCKERHUB_PASS &&
+          make push-nodep
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -63,6 +63,8 @@ jobs:
       - restore_cache:
           key: machine-controller-{{ .Revision }}
       - restore_cache:
+          key: vendor-{{ checksum "Gopkg.lock" }}
+      - restore_cache:
           key: repo-{{ .Environment.CIRCLE_SHA1 }}
       - run: cd /go/src/github.com/kubermatic/machine-controller && make docker-image
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,15 +59,14 @@ jobs:
           go/src/github.com/kubermatic/machine-controller &&
           sudo mkdir -p /go/src/github.com/kubermatic/machine-controller && sudo chmod 777 -R
           /go/src/github.com/kubermatic/machine-controller &&
-          sudo chown -R circleci:circleci /go/src/github.com/kubermatic/machine-controller &&
-          cd /go/src/github.com/kubermatic/machine-controller
+          sudo chown -R circleci:circleci /go/src/github.com/kubermatic/machine-controller
       - restore_cache:
           key: machine-controller-{{ .Revision }}
       - restore_cache:
-          key: vendor-{{ checksum "Gopkg.lock" }}
+          key: vendor-{{ checksum "/go/src/github.com/kubermatic/machine-controller/Gopkg.lock" }}
       - restore_cache:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -20,6 +20,13 @@ machine-controller: $(shell find cmd pkg -name '*.go') vendor
 
 docker-image: machine-controller
 	docker build -t $(IMAGE_NAME) .
+	if git describe --tags $(shell git rev-parse HEAD)|grep -v -- '-g'; then \
+		$(eval IMAGE_TAG = $(shell git describe --abbrev=0 --tags)) \
```
mrIncompetent: You can get the tag from the circle ci environment variables: https://circleci.com/docs/2.0/env-vars/#build-details

mrIncompetent: nevermind - i'm an idiot. Obviously those are not suitable for the Makefile :man_facepalming:

Makefile
```diff
@@ -19,7 +21,26 @@ machine-controller: $(shell find cmd pkg -name '*.go') vendor
 				github.com/kubermatic/machine-controller/cmd/controller
 
 docker-image: machine-controller
+	make docker-image-nodep
+
+docker-image-nodep:
```
mrIncompetent: what means nodep?

Makefile
```diff
@@ -19,7 +21,26 @@ machine-controller: $(shell find cmd pkg -name '*.go') vendor
...
 	docker build -t $(IMAGE_NAME) .
+	if git describe --tags $(shell git rev-parse HEAD)|grep -v -- '-g'; then \
+		$(eval IMAGE_TAG = $(shell git describe --abbrev=0 --tags)) \
```
mrIncompetent: So there are 2 things for me here:
- I would rather have 1 task called `push` which builds the image and pushes it. - Why have 2 tasks for it?
- I would rather use an environment variable for the tag. Parsing some output always makes me cringe

.circleci/config.yml
```diff
@@ -45,6 +45,31 @@ jobs:
       - restore_cache:
           key: vendor-{{ checksum "Gopkg.lock" }}
       - run: go build -o machine-controller github.com/kubermatic/machine-controller/cmd/controller
+      - save_cache:
+          key: machine-controller-{{ .Revision }}
+          paths:
+            - machine-controller
+
+  publish:
+    machine: true
```
mrIncompetent: why machine instead of
```
    docker:
      - image: docker:stable
```
?

.circleci/config.yml
```diff
@@ -45,6 +45,31 @@ jobs:
...
+    steps:
```
mrIncompetent: Specify:
`working_directory: /go/src/github.com/kubermatic/machine-controller`
On the publish level, so you dont have to do a change-directory

.circleci/config.yml
```diff
@@ -45,6 +45,31 @@ jobs:
...
+      - restore_cache:
+          key: vendor-{{ checksum "/go/src/github.com/kubermatic/machine-controller/Gopkg.lock" }}
+      - run: >
+          cd /go/src/github.com/kubermatic/machine-controller &&
+          make docker-image-nodep &&
+          docker login --help &&
+          docker version &&
+          set -u &&
+          docker login -u "${DOCKERHUB_USER}" -p "${DOCKERHUB_PASS}" &&
+          make push-nodep
```
mrIncompetent: why `&&` here?

Why not simply:
```
      - run: |
          make docker-image-nodep 
          ...
```

.circleci/config.yml
```diff
@@ -45,6 +45,31 @@ jobs:
...
+      - restore_cache:
+          key: repo-{{ .Environment.CIRCLE_SHA1 }}
+      - restore_cache:
+          key: machine-controller-{{ .Revision }}
+      - restore_cache:
+          key: vendor-{{ checksum "/go/src/github.com/kubermatic/machine-controller/Gopkg.lock" }}
+      - run: >
+          cd /go/src/github.com/kubermatic/machine-controller &&
+          make docker-image-nodep &&
+          docker login --help &&
```
mrIncompetent: whats that for?

.circleci/config.yml
```diff
@@ -45,6 +45,31 @@ jobs:
...
+      - save_cache:
+          key: machine-controller-{{ .Revision }}
+          paths:
+            - machine-controller
+
+  publish:
+    machine: true
+    steps:
+      - run: >
+          sudo mkdir -p /go/src/github.com/kubermatic/machine-controller &&
```
mrIncompetent: sudo should not be required. If its required, we a producing this binary with wrong permissions

.circleci/config.yml
alvaroaleman: Because that's what I found in the docs: https://circleci.com/docs/2.0/building-docker-images/#example

.circleci/config.yml
alvaroaleman: Yeah I tried that, makes the build instantly fails because that dir doesn' exist

.circleci/config.yml
alvaroaleman: It is not required because of the binary, it is required to create the folders

.circleci/config.yml
alvaroaleman: Debugging ;) Will remove

.circleci/config.yml
alvaroaleman: Yeah, combined with a `set -x` that will do the same, I can change it

Makefile
alvaroaleman: Yeah, thats rather non-obvious. The reason is that make uses timestamps on files to determine if it has to rebuild something. CircleCI uses caches from any build, which means that sometimes it restores the `vendor` folder from an older build. The result is:

* Make sees `Gopkg.{toml,lock}` are newere than the vender folder
* -> Have to build the vendor target
* -> Vendor target is newer than the binary ->Have to rebuild the binary

To skip all that (which makes perfect sense locally btw), I added that `-nodep` target

Makefile
alvaroaleman: I guess we could merge the targets, but I dislike relying on CI-Implementiation-Specific env vars

.circleci/config.yml
mrIncompetent: Can you try again?
In https://github.com/kube-node/kube-machine/blob/master/.circleci/config.yml
We didn't had to do it

Makefile
mrIncompetent: Regarding the env var, i use TAG or something within the makefile and in the pipeline i would set ENV=CIRCLE_SHA1

alvaroaleman: Done that

alvaroaleman:
@mrIncompetent PTAL, all comments should be addressed now. You can check a successfully push [in th build for 900ecc5](https://circleci.com/gh/kubermatic/machine-controller/943)

Issue #58, opened 2018-02-02
add leaderelection


mrIncompetent:
Use the leaderelection component from the kubernetes go-client

PR #73, opened 2018-02-07
factors out controller context and moves synchronisation logic

p0lyn0mial:
**What this PR does / why we need it**: the idea is that we will create the controller from the context in a "loop". In addition this pull also moves synchronisation logic down to the controller itself.

**Which issue(s) this PR fixes**: see #58

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"reflect"
 	"strings"
 	"time"
 
@@ -38,6 +37,7 @@ import (
 	apiextclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 )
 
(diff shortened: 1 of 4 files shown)


p0lyn0mial:
we can hold this pull, I'll issue a few more so that we can see where this all is going and we can decide if we like it or not.

mrIncompetent:
/assign @metalmatze 
I'm a bit swamped right now

pkg/controller/machine.go
```diff
@@ -94,15 +95,27 @@ type MetricsCollection struct {
...
+	metrics MetricsCollection,
+	stopCh <-chan struct{}) *Controller {
+
+	go kubeInformerFactory.Start(stopCh)
+	go machineInformerFactory.Start(stopCh)
+
+	for _, syncsMap := range []map[reflect.Type]bool{kubeInformerFactory.WaitForCacheSync(stopCh), machineInformerFactory.WaitForCacheSync(stopCh)} {
+		for key, synced := range syncsMap {
+			if !synced {
+				glog.Fatalf("unable to sync %s", key)
```
p0lyn0mial: I would prefer to return an error :)

mrIncompetent:
Very important thing: please rebase first. The arguments for NewController changed.
We stopped passing in the informerFactories and instead passed in, the required informers

p0lyn0mial:
> Very important thing: please rebase first. The arguments for NewController changed.

grr.

cmd/controller/main.go
```diff
@@ -50,6 +50,63 @@ var (
...
+
+	// TODO: desc
+	machineInformerFactory machineinformers.SharedInformerFactory
+
+	// TODO: desc
+	stopCh <-chan struct{}
+}
+
+// TODO: desc
+func createClientsOrDie(cfg *rest.Config) (kubernetes.Interface, apiextclient.Interface, machineclientset.Interface) {
```
metalmatze: I think it's fine to create these clients simply in the main func. We don't need to wrap these things.
Plus: Do we really want to return interfaces? You should always return concrete types. Other funcs can then ask for Interfaces.

cmd/controller/main.go
```diff
@@ -50,6 +50,63 @@ var (
 	workerCount   int
 )
 
+// TODO:desc
+type controllerContext struct {
```
metalmatze: It's weird calling this context. Not really sure if that's a common thing within Kubernetes, but these days, when I see something called context I mostly expect it to use the `context` package. :thinking:

pkg/controller/machine.go
```diff
@@ -94,15 +95,27 @@ type MetricsCollection struct {
 	NodeJoinDuration    metrics.Histogram
 }
 
-// NewMachineController returns a new machine controller
-func NewMachineController(
+// NewMachineControllerOrDie returns a new machine controller
+func NewMachineControllerOrDie(
```
metalmatze: Why don't we simply make this a `func NewMachineController(...) (*Controller, error) {` and let the caller decide what they want to do?
The caller can still decide if they want to fail (log something and then exit != 0) or if they want to maybe backoff and retry, or simply continue.

p0lyn0mial: yes, good idea.

cmd/controller/main.go
p0lyn0mial: sure we can change the name, I don't have strong naming preferences.

PR #74, opened 2018-02-07
decouples readiness checks from controller

p0lyn0mial:
**Which issue(s) this PR fixes**: see #58


pkg/utils/utils.go
```diff
@@ -0,0 +1,56 @@
+package utils
```
p0lyn0mial: we can find a better place for the following function - just a proposition.

cmd/controller/main.go
```diff
@@ -172,7 +174,7 @@ func main() {
 
 	health := healthcheck.NewHandler()
 	health.AddReadinessCheck("apiserver-connection", machinehealth.ApiserverReachable(kubeClient))
-	for name, c := range c.ReadinessChecks() {
+	for name, c := range utils.ReadinessChecks(createConfigMapInformer(kubeClient)) {
```
p0lyn0mial: I want to run `ReadinessChecks` outside the "loop"

pkg/utils/utils.go
mrIncompetent: We should avoid a `utils` package. Instead we should define explicit packages.

For `GetClusterInfoKubeconfig` i would create a new package called `clusterinfo`
For `ReadinessChecks` i would move it into the `health` package

p0lyn0mial: why not, sounds good to me.

mrIncompetent: The reason against using a `utils` is mostly because it normally becomes a mess where people put code it, they don't know where to put else. It becomes unorganized and messy.

A good read on that is: https://blog.golang.org/package-names

cmd/controller/main.go
```diff
@@ -182,6 +184,13 @@ func main() {
 	}
 }
 
+// TODO: desc
+func createConfigMapInformer(kubeClient kubernetes.Interface) listerscorev1.ConfigMapLister {
```
metalmatze: I don't necessarily see a reason this needs to be in another func. 😉

pkg/utils/utils.go
metalmatze: I think that we don't actually need another package at all. 
I'd put the `ReadinessChecks` simply into the main.go file. That's the only place this is used. There's no need to put it into another package. 

With the `GetClusterInfoKubeconfig` I'm not entirely sure, because it's use in `ReadinessChecks` and somewhere in the controller AFAIK.

Thoughts?

mrIncompetent: @metalmatze i agree

PR #77, opened 2018-02-07
run machine controller through leader election

p0lyn0mial:


p0lyn0mial:
this pull demonstrates the whole concept, just wanted to show you where we are heading for.
I think that it will take a few more pulls to get it done.

Oh, I also need to concentrate on the hardest part ... testing :)

PR #79, opened 2018-02-09
add fallback for kubeconfig retrieval via endpoints and client config ca

### LINKED ISSUE #78: Do not use cluster-info configmap anymore
Right now the `machine-controller` uses the `cluster-info` configmap to get the CACert and the endpoint for the apiserver.

Instead it should get the CACert from its `kubeconfig` or from `/run` and the apiserver endpoints from its `kubeconfig` or from the endpoints of the `kubernetes` service when running in-cluster.

This will reduce the configuration overhead and help ppl to get started faster.

mrIncompetent:
Adds a defaulting for the case the cluster-info configmap does not exist.
In this case, we fetch the first IP from end Kubernetes endpoints and the ca from the pod passed kubeconfig

Fixes #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,16 +36,20 @@ kubectl create -f examples/machine.yaml
 ## Advanced usage
 
 ### Specifying the apiserver endpoint
-By default the controller looks for a cluster-info ConfigMap within the kube-public Namespace.
+By default the controller looks for a `cluster-info` ConfigMap within the `kube-public` Namespace.
 If one is found which contains a minimal kubeconfig (kubeadm cluster have them by default), this kubeconfig will be used for the node bootstrapping.
-The kubeconfig only needs to container only two things:
+The kubeconfig only needs to contain two things:
 - CA-Data
 - The public endpoint for the Apiserver 
 
 If no ConfigMap can be found:
-**CA-data** 
+
+**CA-data**
+ 
 The CA will be loaded from the passed kubeconfig when running outside the cluster or from `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` when running inside the cluster.
+
 **Apiserver endpoint**
+
 The first endpoint from the kubernetes endpoints will be taken. `kubectl get endpoints kubernetes -o yaml`
   
 #### Example cluster-info ConfigMap
--- a/README.md
+++ b/README.md
@@ -69,3 +69,9 @@ make machine-controller
 ```bash
 ./controller -logtostderr -v=8 -kubeconfig=/path/to/kubeconfig
 ```
+
+## End-to-end testing
+
+For a simple e2e-testing using Hetzner cloud, just run `make test-e2e`.
+This requires the environment variable `HZ_TOKEN` to be filled with a
+valid token.
--- a/README.md
+++ b/README.md
@@ -86,7 +86,13 @@ make machine-controller
 ./controller -logtostderr -v=8 -kubeconfig=/path/to/kubeconfig
 ```
 
-## End-to-end testing
+## Testing
+
+### Unittests
+
+Simply run `make test-unit`
+
+### End-to-End
 
 For a simple e2e-testing using Hetzner cloud, just run `make test-e2e`.
 This requires the environment variable `HZ_TOKEN` to be filled with a
--- a/README.md
+++ b/README.md
@@ -21,12 +21,34 @@
 ## What does not work
 - Master creation (Not planned at the moment)
 
-# Running
+# Quickstart
 
-## Requirements
-The controller expects a cluster-info configmap to exist within the kube-public namespace.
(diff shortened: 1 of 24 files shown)


examples/machine-controller.yaml
```diff
@@ -105,6 +105,7 @@ rules:
   - ""
   resources:
   - configmaps
+  - endpoints
```
alvaroaleman: This is the wrong namespace, the `endpoints` are in the `default` namespace. This means we need another `clusterRole{,Binding}`

alvaroaleman:
Thinking about the issue some more, maybe it would be smart to create a `cluster-info` configmap and popupate it with the info from `endpoints`/`kubeconfig` instead of getting it ad-hoc?

mrIncompetent:
Though about that as well, but that would mean that we set a specific endpoint IP as "final".
I would prefer to take whatever ip is there at the moment.

alvaroaleman:
Okay, you have a point there.

What I find a little annoying, is that it now logs on every healthcheck that the configmap couldn't be found. It would be way better if we only logged that once and then have a log message when we use a new endpoint address or data from the configmap.

mrIncompetent:
Its getting logged with level 6, which shouldn't be used in production.
And as we check everytime if we can find a configmap, i would like to keep it that way

mrIncompetent:
@alvaroaleman PTAL

README.md
```diff
@@ -21,12 +21,34 @@
...
+## Creating a machine
+```bash
+# edit examples/machine.yaml & create the machine
+kubectl create -f examples/machine.yaml
+```
+
+## Advanced usage
+
+### Specifying the apiserver endpoint
+By default the controller looks for a cluster-info ConfigMap within the kube-public Namespace.
```
alvaroaleman: Please put the `cluster-info` and `kube-public` in backticks to mark as code

README.md
```diff
@@ -21,12 +21,34 @@
...
+If one is found which contains a minimal kubeconfig (kubeadm cluster have them by default), this kubeconfig will be used for the node bootstrapping.
+The kubeconfig only needs to container only two things:
+- CA-Data
+- The public endpoint for the Apiserver 
+
+If no ConfigMap can be found:
```
alvaroaleman: You need to insert newlines before and after `**CA-data**`/`**Apiserver endpoint**`, otherwise the markdown rendering puts "If no ConfigMap can be found: **CA-data** " into the same line. (I usually just go to the repo and select the branch to see the rendered markdown and verify it looks as intended).

examples/machine-controller.yaml
```diff
@@ -98,7 +98,7 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: Role
 metadata:
-  name: machine-controller
+  name: configmap-reader-kube-public
```
alvaroaleman: I really dislike putting meta infos and content of something into its name instead of simply the purpose because:

* It is redundant
* It means the name has to be changed everytime the contents of that object change
* It is not not easily possible to find out afterwards what for this was created

examples/machine-controller.yaml
```diff
@@ -111,14 +111,43 @@ rules:
...
+  verbs:
+  - get
+  - watch
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: RoleBinding
 metadata:
-  name: machine-controller
+  name: machine-controller-kube-public
```
alvaroaleman: This does not work, the `RoleBinding` has to be in the same namespace as the `Role`

examples/machine-controller.yaml
```diff
@@ -111,14 +111,43 @@ rules:
...
+subjects:
+- kind: ServiceAccount
   name: machine-controller
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
+metadata:
+  name: machine-controller-default
+  namespace: kube-system
```
alvaroaleman: This does not work, the `RoleBinding` has to be in the same namespace as the `Role`

README.md
```diff
@@ -21,12 +21,34 @@
...
+# edit examples/machine.yaml & create the machine
+kubectl create -f examples/machine.yaml
+```
+
+## Advanced usage
+
+### Specifying the apiserver endpoint
+By default the controller looks for a cluster-info ConfigMap within the kube-public Namespace.
+If one is found which contains a minimal kubeconfig (kubeadm cluster have them by default), this kubeconfig will be used for the node bootstrapping.
+The kubeconfig only needs to container only two things:
```
alvaroaleman: `s/container/contain/g`, also the seconds `two` must vanish

alvaroaleman: The second `only` not the second `two`

mrIncompetent:
@alvaroaleman PTAL

PR #90, opened 2018-02-12
finalizes the work on the leader election

p0lyn0mial:
**What this PR does / why we need it**: this PR finalizes the work on the leader election


p0lyn0mial:
@mrIncompetent , @metalmatze I have also addressed your previous comments. Please take another look and tell me what do you think.

cmd/controller/main.go
```diff
@@ -147,16 +108,33 @@ func main() {
 		glog.Fatalf("invalid cluster dns specified: %v", err)
 	}
 
-	// set up signals so we handle the first shutdown signal gracefully
-	// TODO: fix stopCh
-	//stopCh := signals.SetupSignalHandler()
+	// TODO: add graceful shutdown, propagate stopCh to run method and to http server
```
p0lyn0mial: graceful shutdown is broken - we need to fix it. I think I could add an issue to our backlog.

mrIncompetent:
ci is failing.
But lgtm from my side.
@metalmatze needs to approve as well before merging

mrIncompetent:
@metalmatze Not sure if the kubernetes components support something other than a stop channel

p0lyn0mial:
> Can't we use context.Context instead of the pre context days hack with stopCh?

as @mrIncompetent  already pointed out, kubernetes doesn't accept `context` but that doesn't stop us from using `context` :D. Actually my initial plan was to use `context` and derive other synchronisation primitives including `stopCh`and pass them to `startUtilHttpServerOrDie` and `startControllerViaLeaderElectionOrDie`. 

Perhaps a little more challenging can be synchonization between HTTP server and the controller itself i.e. exit the app when both are down. Also I haven't figured yet what to do when the controller haven't entered `run` routine - haven't became a leader. How can we tell the leader election library to stop ?
 
I think that gracefull shutdown takes a little scrutiny to get it right and make it elegant. That's why I suggested to create a separate issue unless the lack of proper shutdown blocks this PR ?

mrIncompetent:
No, the missing proper shutdown should not be part of this PR.
@metalmatze opened a PR which should take care of managing multiple go-routines & gracefully shutting them down.

p0lyn0mial:
@mrIncompetent, @metalmatze I'll merge all my PRs, rebase against the master and I think I'll need to issue one more pull.
@metalmatze you still have a chance to say "no way" :)

PR #94, opened 2018-02-13
run machine controller via leader election

### LINKED ISSUE #58: add leaderelection

p0lyn0mial:
**What this PR does / why we need it**: runs the machine controller via leader election this essentially means that we can have multiple instances and at the same time only one is operational.
before we acquire a lock we actually warm up caches mirroring the state of the API server.
machine controller will be run only if a proper lock was acquired.
note that the program terminates when the leadership was lost.

**Which issue(s) this PR fixes**: Fixes #58

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -17,18 +17,35 @@ limitations under the License.
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"net"
 	"net/http"
+	"os"
 	"reflect"
 	"strings"
 	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	apiextclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
+	kubeinformers "k8s.io/client-go/informers"
+	corev1informers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/leaderelection"
+	"k8s.io/client-go/tools/leaderelection/resourcelock"
+	"k8s.io/client-go/tools/record"
 
 	"github.com/golang/glog"
 	"github.com/heptiolabs/healthcheck"
 	machineclientset "github.com/kubermatic/machine-controller/pkg/client/clientset/versioned"
 	machineinformers "github.com/kubermatic/machine-controller/pkg/client/informers/externalversions"
+	machinesv1alpha1 "github.com/kubermatic/machine-controller/pkg/client/informers/externalversions/machines/v1alpha1"
 	"github.com/kubermatic/machine-controller/pkg/clusterinfo"
 	"github.com/kubermatic/machine-controller/pkg/controller"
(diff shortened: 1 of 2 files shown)


cmd/controller/main.go
```diff
@@ -25,23 +26,33 @@ import (
 	"strings"
 	"time"
 
+	"os"
```
mrIncompetent: should be appended to the imports above.

We normally separate:
- Standard Go SDK imports
- k8s.io
- everything else

cmd/controller/main.go
```diff
@@ -25,23 +26,33 @@ import (
...
+
 	"github.com/golang/glog"
 	"github.com/heptiolabs/healthcheck"
 	machineclientset "github.com/kubermatic/machine-controller/pkg/client/clientset/versioned"
 	machineinformers "github.com/kubermatic/machine-controller/pkg/client/informers/externalversions"
+	"github.com/kubermatic/machine-controller/pkg/client/informers/externalversions/machines/v1alpha1"
```
mrIncompetent: alias
`machinesv1alpha1`

cmd/controller/main.go
```diff
@@ -25,23 +26,33 @@ import (
...
 	"github.com/kubermatic/machine-controller/pkg/clusterinfo"
 	"github.com/kubermatic/machine-controller/pkg/controller"
 	machinehealth "github.com/kubermatic/machine-controller/pkg/health"
 	"github.com/kubermatic/machine-controller/pkg/machines"
 	"github.com/kubermatic/machine-controller/pkg/signals"
 	"github.com/kubermatic/machine-controller/pkg/ssh"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
-
+	"k8s.io/api/core/v1"
```
mrIncompetent: alias
`corev1`

cmd/controller/main.go
```diff
@@ -25,23 +26,33 @@ import (
...
 	apiextclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	kubeinformers "k8s.io/client-go/informers"
+	corev1informers "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
```
mrIncompetent: alias
`typedcorev1`

`corev1` is normally used for `"k8s.io/api/core/v1"`

cmd/controller/main.go
```diff
@@ -134,18 +188,150 @@ func main() {
...
+
+	leaderelection.RunOrDie(leaderelection.LeaderElectionConfig{
+		Lock:          &rl,
+		LeaseDuration: defaultLeaderElectionLeaseDuration,
+		RenewDeadline: defaultLeaderElectionRenewDeadline,
+		RetryPeriod:   defaultLeaderElectionRetryPeriod,
+		Callbacks: leaderelection.LeaderCallbacks{
+			OnStartedLeading: run,
+			OnStoppedLeading: func() {
+				glog.Infof("leaderelection lost, closing the app")
```
mrIncompetent: should this be a `fatal`? or is there any other callback which exists?

p0lyn0mial: yes, should be `fatal` - thanks! To the best of my knowledge only `OnStoppedLeading` exits.

p0lyn0mial:
@metalmatze I'm going to merge this PR, in case you would like to change something just let me know.

PR #96, opened 2018-02-13
make default securitygroup creation a fallback when none is set

### LINKED ISSUE #85: Make security-groups creation on openstack a fallback
We should only create a security-group on Openstack when none is defined.
Also we need to log this with a loglevel of 2 probably

mrIncompetent:
make default securitygroup creation a fallback when none is set

Fixes #85

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -301,9 +301,14 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 		}
 	}
 
-	err = ensureKubernetesSecurityGroupExist(client, c.Region, securityGroupName)
-	if err != nil {
-		return nil, fmt.Errorf("failed to ensure that the kubernetes security group %q exists: %v", securityGroupName, err)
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -301,17 +301,22 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 		}
 	}
 
-	err = ensureKubernetesSecurityGroupExist(client, c.Region, securityGroupName)
-	if err != nil {
-		return nil, fmt.Errorf("failed to ensure that the kubernetes security group %q exists: %v", securityGroupName, err)
+	securityGroups := c.SecurityGroups
+	if len(c.SecurityGroups) == 0 {
```
alvaroaleman: This doesn't make any actual difference, but wouldn't it it be more consistent to only use `securityGroups` instead of `c.SecurityGroups` after assigning the latter to the former?

mrIncompetent: good point!

mrIncompetent:
addresses the comment and rebased

PR #106, opened 2018-02-15
simplify if statement

mrIncompetent:


pkg/controller/machine.go
```diff
@@ -274,10 +274,10 @@ func (c *Controller) syncHandler(key string) error {
 		} else if machineControllerName != "" && c.name == "" {
```
p0lyn0mial: I think we could remove this else as well :) ?

mrIncompetent: Nope, we have 2 if statements here,
first: controller is running with a given name but the machine has no name specified.
second: other way around: controller is running without a name but the machine has a name specified.

I wanted to log the difference explcitly

PR #110, opened 2018-02-20
adds graceful shutdown capability

### LINKED ISSUE #101: Add graceful shutdown

p0lyn0mial:
**What this PR does / why we need it**: adds graceful shutdown capability

Fixes #101

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"errors"
 	"flag"
 	"fmt"
@@ -25,6 +26,7 @@ import (
 	"os"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
@@ -53,6 +55,7 @@ import (
 	"github.com/kubermatic/machine-controller/pkg/machines"
 	"github.com/kubermatic/machine-controller/pkg/signals"
 	"github.com/kubermatic/machine-controller/pkg/ssh"
+	"github.com/oklog/run"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
@@ -118,6 +121,9 @@ type controllerRunOptions struct {
 
 	// name of the controller. When set the controller will only process machines with the annotation "machine.k8s.io/controller": name
 	name string
+
+	// ctx carries a cancelation signal
+	ctx context.Context
 }
 
 func main() {
@@ -136,7 +142,6 @@ func main() {
 		glog.Fatalf("invalid cluster dns specified: %v", err)
 	}
 
-	// TODO: add graceful shutdown, propagate stopCh to run method and to http server
 	stopCh := signals.SetupSignalHandler()
 
 	cfg, err := clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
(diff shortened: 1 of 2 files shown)


cmd/controller/main.go
```diff
@@ -211,14 +216,44 @@ func main() {
 		}
 	}
 
-	startUtilHttpServerOrDie(kubeClient, kubeconfigProvider, stopCh)
-	startControllerViaLeaderElectionOrDie(runOptions)
+	ctx, ctxDone := context.WithCancel(context.Background())
+	runOptions.ctx = ctx
```
metalmatze: Remove this line here.

cmd/controller/main.go
```diff
@@ -211,14 +216,44 @@ func main() {
...
+				return errors.New("parent context has been closed - propagating the request")
+			}
+		}, func(err error) {
+			ctxDone()
+		})
+
+	}
+	{
+		g.Add(func() error {
+			return startControllerViaLeaderElection(runOptions)
```
metalmatze: Before this line do: `runOptions.ctx = ctx`

cmd/controller/main.go
```diff
@@ -211,14 +216,44 @@ func main() {
...
 	}
 
-	startUtilHttpServerOrDie(kubeClient, kubeconfigProvider, stopCh)
-	startControllerViaLeaderElectionOrDie(runOptions)
+	ctx, ctxDone := context.WithCancel(context.Background())
+	runOptions.ctx = ctx
+	var g run.Group
+	{
+		g.Add(func() error {
+			return startUtilHttpServer(kubeClient, kubeconfigProvider, ctx)
```
metalmatze: Take `startUtilHttpServer` out of that function and create the `http.Server{}` directly before adding this group. Then there's no need to pass the context. ☺️

cmd/controller/main.go
```diff
@@ -211,14 +216,44 @@ func main() {
...
+		}, func(err error) {
+			ctxDone()
```
metalmatze: Instead of `ctxDone()` run `server.Shutdown()` on the server you created above in the same group.

cmd/controller/main.go
```diff
@@ -211,14 +216,44 @@ func main() {
...
+		})
+	}
+	{
+		g.Add(func() error {
```
metalmatze: I think we don't need this group. 😉

cmd/controller/main.go
```diff
@@ -261,49 +305,82 @@ func startControllerViaLeaderElectionOrDie(runOptions controllerRunOptions) {
...
 			OnStoppedLeading: func() {
-				glog.Fatalf("leaderelection lost, closing the app")
+				newCtxDone()
 			},
 		},
 	})
+
+	var g run.Group
+	{
+		g.Add(func() error {
```
metalmatze: This group func looks pretty weird. I think we can find a better way to handle this. 😊

cmd/controller/main.go
p0lyn0mial: I think that having a separate function that encloses and manages creation of an HTTP server is actually nice and readable. In addition I also think that passing a context to dependent function is actually okay as context was designed for that purpose ? (To propagate a cancellation).

cmd/controller/main.go
p0lyn0mial: this group listens for Ctrl-C, for example a user requested an app to be closed.

cmd/controller/main.go
p0lyn0mial: yeah, I'm open to suggestion. One thing worth noting is the fact that there is no way to tell the leader election library to stop.

cmd/controller/main.go
p0lyn0mial: If we will remove this group how can we handle such request ?

mrIncompetent:
My 2 cents:
I would prefer having only a single `run.Group` instead of multiple.
It would simplify the general logic.

`startUtilHttpServer` would need to return a `*http.Server` -> `getUtilHttpServer`
`startControllerViaLeaderElection` would need to return the `*leaderelection.LeaderElector` -> `getControllerLeaderElection`

Also, can we move the controller instantiation outside of the run function for the leader election?

p0lyn0mial:
I'm closing this pull, see: https://github.com/kubermatic/machine-controller/pull/112

Issue #107, opened 2018-02-16
Allow to specify ssh-key via flag


mrIncompetent:
Current state:
On initial start, we check if a secret with a private ssh key exists.
If no secret is found, we generate a secret with a private key.

This ssh key will be later used when creating instances at cloud-providers.
This was made so the user does not have to specify a ssh public key in the machine-manifest, as some cloud providers require to specify a public key when creating a instance (aws).

All public keys from the machine manifest are getting deployed via cloud-init.

Desired state:
The controller should accept a path to a private key via a command line flag.
If the flag is specified and a valid key got found, this key should be taken.
If no flag was specified or the key was not found, the old logic with the secret should apply.

alvaroaleman:
Thinking about this some more, I dislike that idea. We should instead remove this whole ssh key deployment functionality everywhere where it is not absolutely required (AWS) and let users pass in their ssh key via the machineSpec if they so chose.

Allowing to pass a ssh key via flag is just an incentive to not put the key in the place it belongs, the machineSpec.

mrIncompetent:
Pro:
- We can get rid of ssh key handling code in all providers - except aws

Cons:
- We would need to extend the key handling code in AWS
  - Check if any of the specified ssh keys exists in the account
  - If none of the specified keys exists, we would need to create it -> which one and which name to take?
- Different behavior on the different cloud-providers -> bad user-experience
- We would introduce a hard coupling of the `sshPublicKeys` property and the cloud-provider

alvaroaleman:
No, leave the code for AWS as-is, only remove the handling from the other cloud providers.

mrIncompetent:
So we maintain the initial creation of the ssh-key + secret but we simply ignore it for everyone except AWS?
I'm pretty sure this will create the most confusion - as users now see a secret containing a ssh key secret in their cluster although it'll never get used

alvaroaleman:
I don't think so, the machine-controller itself is cloud-agnostic, it doesn't have a flag `--cloud-provider=<my-cloud-provider>` thus it has to do everything that any of the supported clouds may need.

We can just add a note in the Readme why this is needed and also add a comment in the code. I find this much less confusing than creating a ssh key for every cloud provider just because AWS needs it.

mrIncompetent:
Need to check if the ssh key is still necessary

mrIncompetent:
Digitalocean requires us to specify an ssh key. Otherwise the api will respond with:
[...]

PR #115, opened 2018-02-24
E2E testing in CI

### LINKED ISSUE #87: Add end-to-end testing to CircleCI pipeline
To better know if PRs add bugs, we should add the existing end-to-end testing to the CircleCI-Pipeline.

This requires:

* Configuring the ssh keypair name at the cloudprovider with a random prefix
* Deleting the ssh keypair before ending the e2e test
* Adding the `test-e2e` target to circleci

alvaroaleman:

Add end2end testing to CI pipeline

**Which issue(s) this PR fixes** :
Will eventually resolve #87 


alvaroaleman:
This won't clean up a testnode in case it was created and an error happened afterwards, however I think if this ever happens its fine to manually clean it up.

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -133,38 +131,6 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 	return nil
 }
 
-func ensureSSHKeysExist(ctx context.Context, client hcloud.SSHKeyClient, key *machinessh.PrivateKey) (*hcloud.SSHKey, error) {
```
mrIncompetent: see my points in https://github.com/kubermatic/machine-controller/issues/107

mrIncompetent:
Blocked by https://github.com/kubermatic/machine-controller/issues/119

alvaroaleman:
IMHO this is done, PTAL @mrIncompetent

.circleci/config.yml
```diff
@@ -62,6 +62,77 @@ jobs:
...
+      - restore_cache:
+          key: vendor-{{ checksum "Gopkg.lock" }}
+      - setup_remote_docker
+      - run: apk update && apk add make bash git
+      - run: |
+          set -e
+          docker create -v /go/src/github.com/kubermatic/machine-controller --name vol-container alpine /bin/true
+          docker cp . vol-container:/go/src/github.com/kubermatic/machine-controller/
+          export USE_VOLUME_CONTAINER=true && cd test/tools/verify && make verify
+          docker cp vol-container:/go/src/github.com/kubermatic/machine-controller/test/tools/verify/verify .
```
mrIncompetent: Can you quickly explain why we do this instead of a a simple `docker build` ?

mrIncompetent:
I'm not a big fan of using docker container in the makefile in general.
It leads to overly complex Makefiles & build-files.

I would rather prefer the way Kubernetes is doing it. They have a wrapper script `run.sh` which is responsible for running make commands in a special docker container.
So you do `./hack/run.sh make`

.circleci/config.yml
alvaroaleman: We could do a docker build but then we would have to push it on every build

alvaroaleman:
IMHO it won't get any simpler if you put the glue code in a bash script instead of a Makefile. Also if you do it like K8s you always create the container, then decide if you actually need it.

mrIncompetent:
As we use Alpine, creating a Container does not hurt.
The performance improvement we get, does not justify the overall script complexity IMO.

@guusvw @metalmatze @p0lyn0mial Some opinions here?

p0lyn0mial:
I'd prefer to treat CI as a tool that makes calls to our code. By doing so source code is outside of a tool that can be potentially replaced in the future. We can also share the common code. If one wants to change something it doesn't have to dive in CI specific code.

alvaroaleman:
> As we use Alpine, creating a Container does not hurt.

That is probably true

>The performance improvement we get, does not justify the overall script complexity IMO.

The complexity has nothing to do with using or not using `make` - If we moved that to a bashscipt, we would have the same complexity but in a different place. The advantage of make is that it offers a nicer interface for end users.

mrIncompetent:
The complexity was targetet at the `docker cp`

mrIncompetent:
Sorry for the delay!

p0lyn0mial:
yay! we have e2e tests :)

PR #116, opened 2018-02-26
no need to use additional group for synchronisation.

p0lyn0mial:
**What this PR does / why we need it**: removes additional group for synchronisation.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -335,16 +335,8 @@ func startControllerViaLeaderElection(runOptions controllerRunOptions) error {
 	}
 	go le.Run()
 
-	var g run.Group
-	{
-		g.Add(func() error {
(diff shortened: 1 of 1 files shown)


cmd/controller/main.go
```diff
@@ -335,16 +335,8 @@ func startControllerViaLeaderElection(runOptions controllerRunOptions) error {
...
-	{
-		g.Add(func() error {
-			<-runOptions.parentCtx.Done()
-			return errors.New("closing the app because leadership position was lost")
-		}, func(err error) {
-			runOptions.parentCtxDone()
-		})
-	}
-	return g.Run()
+	<-runOptions.parentCtx.Done()
```
p0lyn0mial: I've just realised that there is no need to use additional group. At some point in the past we had a code that was using a derived context what seemed to be valid approach.

PR #124, opened 2018-03-02
removes ssh-key generation code

p0lyn0mial:
**What this PR does / why we need it**: this pr removes ssh-key generation logic from all providers except digitalocean.


pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -28,13 +28,14 @@ import (
 )
 
 type provider struct {
+	// TODO(p0lyn0mial): remove privateKey
```
mrIncompetent: As mentioned in the todo, please move it to a local variable within the Create function.

For simplicity we can also always create a ssh key. Regardless if coreos or something else

p0lyn0mial: yes, I'll issue a separate pull (step 2) for that, is that okay ?

PR #125, opened 2018-03-02
creates temporal ssh-key for DO provider as it is mandatory

### LINKED ISSUE #119: Create temporary ssh key during instance creation when required. Delete afterwards
Current state:
On initial start, we check if a secret with a private ssh key exists.
If no secret is found, we generate a secret with a private key.

This ssh key will be later used when creating instances at cloud-providers.
This was made so the user does not have to specify a ssh public key in the machine-manifest, as some cloud providers require to specify a public key when creating a instance (digitalocean).

All public keys from the machine manifest are getting deployed via cloud-init.

Desired state:
The whole ssh key logic should be removed.
If a cloud provider requires a ssh key during instance creation:
- Create a temporary key before the instance get created
- Use the temporary key for instance creation
- Delete the temporary key after the key has been created

p0lyn0mial:
**What this PR does / why we need it**: creates temporal ssh-key for DO provider as it is mandatory

Fixes #119


pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -268,19 +266,20 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
...
 
-	// TODO(p0lyn0mial): an ssh-key should be created only if the CoreOS was requested
-	// TODO(p0lyn0mial): an ssh-key must be temporal - deleted after an instance has been created.
-	if p.privateKey == nil {
-		var err error
-		p.privateKey, err = machinessh.NewPrivateKey("machine-controller")
-		if err != nil {
-			return nil, err
-		}
+	tmpSSHKeyPair, err := machinessh.NewPrivateKey("machine-controller-do-tmp-key")
```
p0lyn0mial: we could also drag the controller name all the way from main.go and use it here

mrIncompetent: I think the whole `machinessh` isn't necessary anymore.

pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -268,19 +266,20 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
...
 	}
-	fingerprint, err := ensureSSHKeysExist(ctx, client.Keys, p.privateKey)
+	fingerprint, err := ensureSSHKeysExist(ctx, client.Keys, tmpSSHKeyPair)
 	if err != nil {
-		return nil, fmt.Errorf("failed ensure that the ssh key '%s' exists: %v", p.privateKey.Name(), err)
+		return nil, fmt.Errorf("failed ensure that the ssh key '%s' exists: %v", tmpSSHKeyPair.Name(), err)
 	}
+	defer func() {
+		_, err := client.Keys.DeleteByFingerprint(ctx, fingerprint)
+		if err != nil {
```
p0lyn0mial: not sure what else we can do in case of an error. I was thinking add retry logic but I am pretty sure it is already implemented by `client.Keys` library

pkg/cloudprovider/provider/digitalocean/provider.go
p0lyn0mial: movie it to `pkg/cloudprovider/provider/digitalocean/provider.go` ?

pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -268,19 +266,20 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
...
-		p.privateKey, err = machinessh.NewPrivateKey("machine-controller")
-		if err != nil {
-			return nil, err
-		}
+	tmpSSHKeyPair, err := machinessh.NewPrivateKey("machine-controller-do-tmp-key")
+	if err != nil {
+		return nil, err
 	}
-	fingerprint, err := ensureSSHKeysExist(ctx, client.Keys, p.privateKey)
+	fingerprint, err := ensureSSHKeysExist(ctx, client.Keys, tmpSSHKeyPair)
```
mrIncompetent: `ensureSSHKeysExist` can be deleted. It contains more code than actual necessary anymore.

I would put the `client.Keys.Create` in here and generate a random name.

pkg/cloudprovider/provider/digitalocean/provider.go
p0lyn0mial: can we set TTL on an ssh-key :D ?

pkg/cloudprovider/provider/digitalocean/provider.go
mrIncompetent: Yeah, probably. Though it contains a wrapper around the actual ssh key, which we dont need anymore.

The only code we need is 
```go
	priv, err := rsa.GenerateKey(rand.Reader, privateKeyBitSize)
	if err != nil {
		return nil, fmt.Errorf("failed to create private key: %v", err)
	}
	publicKey := priv.PublicKey
```

mrIncompetent: Better generate a new random name. Otherwise we run into concurrency issues here

p0lyn0mial: ok.

pkg/cloudprovider/provider/digitalocean/provider.go
mrIncompetent: Keep it simple, simply logging the error is fine.
In the rare case this really fails, it does not hurt.

pkg/cloudprovider/provider/digitalocean/provider.go
p0lyn0mial: good point, before we had only once instance of that key. I think `UUID` is fine.

pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -230,33 +223,30 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
...
+	publicKey := key.publicKey()
 	pk, err := ssh.NewPublicKey(&publicKey)
 	if err != nil {
 		return "", fmt.Errorf("failed to parse publickey: %v", err)
 	}
 
 	fingerprint := ssh.FingerprintLegacyMD5(pk)
-	dokey, res, err := service.GetByFingerprint(ctx, fingerprint)
+	existingkey, res, err := service.GetByFingerprint(ctx, fingerprint)
+	if err == nil && existingkey != nil && res.StatusCode >= http.StatusOK && res.StatusCode <= http.StatusAccepted {
```
p0lyn0mial: maybe this is a little bit defensive since `UUID` gives a a good distribution.

pkg/cloudprovider/provider/digitalocean/rsaKey.go
```diff
@@ -0,0 +1,36 @@
+package digitalocean
+
+import (
+	"crypto/rand"
+	"crypto/rsa"
+	"fmt"
+)
+
+const privateKeyBitSize = 4096
```
p0lyn0mial: why not :) ?

PR #126, opened 2018-03-08
trigger deployment when a tag with "release-" prefix has been added.

p0lyn0mial:
**What this PR does / why we need it**: this PR adds functionality that will trigger a deployment when a tag with "release-" prefix has been added.

Fixes #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,16 +84,28 @@ workflows:
   version: 2
   build:
     jobs:
-      - checkout_code
+      - checkout_code:
+          filters:
+            tags:
+              only: /v.*/
       - install-dependencies:
           requires:
             - checkout_code
+          filters:
+            tags:
+              only: /v.*/
       - test:
           requires:
             - install-dependencies
+          filters:
+            tags:
+              only: /v.*/
       - build:
           requires:
             - install-dependencies
+          filters:
+            tags:
+              only: /v.*/
       - publish:
           requires:
             - build
@@ -101,3 +113,5 @@ workflows:
             branches:
               only:
                 - master
+            tags:
+              only: /v.*/


.circleci/config.yml
```diff
@@ -84,20 +84,34 @@ workflows:
...
           requires:
             - install-dependencies
+          filters:
+            tags:
+              only: /release.*/
       - build:
           requires:
             - install-dependencies
+          filters:
+            tags:
```
p0lyn0mial: unfortunately we have to add `tags` to all jobs, as stated in the documentation:
> a job must have a filters tags section to run as a part of a tag push and all its transitively dependent jobs must also have a filters tags section

see also:
https://circleci.com/docs/2.0/workflows/#git-tag-job-execution

mrIncompetent: Not pretty, but i'm fine with doing so

.circleci/config.yml
```diff
@@ -84,20 +84,34 @@ workflows:
...
+              only: /release.*/
       - publish:
           requires:
             - build
           filters:
             branches:
               only:
                 - master
+            tags:
+              only: /release.*/
```
p0lyn0mial: unfortunately CircleCI treats `tags` and `branches` separately. I'm afraid we cannot say IF branch=master AND tag=release.* then act. Which means that if we have a match on a tag then we will fire off `publish` job.

an example:
https://circleci.com/workflow-run/a3bd9483-36f8-42a2-8fd8-c97a51b92cf7

p0lyn0mial: Which means that `publish` can be triggered on any branch not just master.

mrIncompetent: So this code means, we only publish when the tag begins with `release.` ?

p0lyn0mial: yes, but also on branch=master push (as before)

mrIncompetent: The prefix should only be `v` as we'll only tag with the version.
For example:
- `v0.2.1`
- `v0.3.0`
etc.

p0lyn0mial: we could however read a branch name directly in `publish` job and fail if there is no match.

p0lyn0mial:
FYI: tag will be derived from `CIRCLE_TAG` environment.

PR #134, opened 2018-03-17
Deploy K8s CACert with kubelet and use it to verify requests

alvaroaleman:

Allow the API server to authenticate itself as valid client for the Kubelets REST api

Fixes #

I manually verified logs and exec through the APIserver work again but do not work when just curling the Kubelet.

### DIFF
--- a/pkg/clusterinfo/configmap.go
+++ b/pkg/clusterinfo/configmap.go
@@ -45,6 +45,22 @@ func (p *KubeconfigProvider) GetKubeconfig() (*clientcmdapi.Config, error) {
 	return cm, nil
 }
 
+func (p *KubeconfigProvider) GetCACert() (string, error) {
+	//Need the kubeconfig to extract the CACert for the userdata rendering
+	cm, err := p.GetKubeconfig()
+	if err != nil {
+		return "", fmt.Errorf("failed to get kubeconfig: '%v'", err)
+	}
+	if len(cm.Clusters) != 1 {
+		return "", fmt.Errorf("kubeconfig does not contain exactly one cluster, can not extract cacert...")
+	}
+	for _, clusterConfig := range cm.Clusters {
+		return string(clusterConfig.CertificateAuthorityData), nil
+	}
+
+	return "", fmt.Errorf("No CACert found!")
+}
+
 func (p *KubeconfigProvider) getKubeconfigFromConfigMap() (*clientcmdapi.Config, error) {
 	cm, err := p.configMapLister.ConfigMaps(metav1.NamespacePublic).Get(configMapName)
 	if err != nil {
--- a/pkg/clusterinfo/configmap.go
+++ b/pkg/clusterinfo/configmap.go
@@ -58,7 +58,7 @@ func (p *KubeconfigProvider) GetCACert() (string, error) {
 		return string(clusterConfig.CertificateAuthorityData), nil
 	}
 
-	return "", fmt.Errorf("No CACert found!")
+	return "", fmt.Errorf("no CACert found!")
 }
 
 func (p *KubeconfigProvider) getKubeconfigFromConfigMap() (*clientcmdapi.Config, error) {
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -86,6 +86,7 @@ type Controller struct {
 
 type KubeconfigProvider interface {
 	GetKubeconfig() (*clientcmdapi.Config, error)
+	GetCACert() (string, error)
 }
 
 // MetricsCollection is a struct of all metrics used in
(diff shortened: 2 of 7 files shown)


pkg/controller/machine.go
```diff
@@ -404,13 +404,24 @@ func (c *Controller) syncHandler(key string) error {
...
+				}
+				if len(cm.Clusters) != 1 {
+					return fmt.Errorf("kubeconfig does not contain exactly one cluster, can not extract cacert...")
+				}
+				var userData string
+				for _, clusterConfig := range cm.Clusters {
+					userData, err = userdataProvider.UserData(machine.Spec, kubeconfig, prov, c.clusterDNSIPs, string(clusterConfig.CertificateAuthorityData))
+					if err != nil {
+						return fmt.Errorf("failed get userdata: %v", err)
+					}
```
mrIncompetent: I would rather implement a function in the KubeconfigProvider.

pkg/controller/machine.go
```diff
@@ -404,13 +404,24 @@ func (c *Controller) syncHandler(key string) error {
...
+				cm, err := c.kubeconfigProvider.GetKubeconfig()
 				if err != nil {
-					return fmt.Errorf("failed get userdata: %v", err)
+					return fmt.Errorf("failed to get kubeconfig: '%v'", err)
+				}
+				if len(cm.Clusters) != 1 {
+					return fmt.Errorf("kubeconfig does not contain exactly one cluster, can not extract cacert...")
+				}
+				var userData string
+				for _, clusterConfig := range cm.Clusters {
```
mrIncompetent: Why iterating over all clusters?
If there's more than 1 cluster, we should throw an error. As we cannot determine which cluster to use

pkg/controller/machine.go
```diff
@@ -404,13 +404,24 @@ func (c *Controller) syncHandler(key string) error {
...
+					userData, err = userdataProvider.UserData(machine.Spec, kubeconfig, prov, c.clusterDNSIPs, string(clusterConfig.CertificateAuthorityData))
```
mrIncompetent: Maybe passing the kubeconfigProvider into the `UserData` function would make more sense.
As this is very specific logic in the main sync loop

pkg/controller/machine.go
alvaroaleman: Because its a map and we cant know the right key in advance

pkg/controller/machine.go
alvaroaleman: Not sure I understand that, a function in the KubeConfigProvider to get the userdata..?

pkg/controller/machine.go
alvaroaleman: Because of performance reasons or because of readability? This only gets called when a new machine without associated instance at the cloud provider exists.

I dislike the idea of putting the whole kubeconfigProvider into `UserData`, because the latter only needs the CA from it.

I could move the whole `getCAAndUseItToRenderUserdata` into a seperate function within the `controller` package thought, that would probably improve readability.

pkg/controller/machine.go
mrIncompetent: Fetching of the CA data should be implemented as a function in the KubeconfigProvider

pkg/controller/machine.go
mrIncompetent: The UserData function needs the kubeconfig & ca data (depending if you agree with my suggestion above) from it.

If you don't agree, it's not needed

alvaroaleman:
@mrIncompetent  PTAL

pkg/clusterinfo/configmap.go
```diff
@@ -45,6 +45,22 @@ func (p *KubeconfigProvider) GetKubeconfig() (*clientcmdapi.Config, error) {
...
+		return "", fmt.Errorf("failed to get kubeconfig: '%v'", err)
+	}
+	if len(cm.Clusters) != 1 {
+		return "", fmt.Errorf("kubeconfig does not contain exactly one cluster, can not extract cacert...")
+	}
+	for _, clusterConfig := range cm.Clusters {
+		return string(clusterConfig.CertificateAuthorityData), nil
+	}
+
+	return "", fmt.Errorf("No CACert found!")
```
mrIncompetent: errors are always start lowercase

pkg/controller/machine.go
```diff
@@ -404,13 +405,17 @@ func (c *Controller) syncHandler(key string) error {
 					return fmt.Errorf("failed to create bootstrap kubeconfig: %v", err)
 				}
 
-				data, err := userdataProvider.UserData(machine.Spec, kubeconfig, prov, c.clusterDNSIPs)
+				clusterCACert, err := c.kubeconfigProvider.GetCACert()
+				if err != nil {
+					return fmt.Errorf("Error getting CACert: '%v'", err)
```
mrIncompetent: errors are always start lowercase

alvaroaleman:
@mrIncompetent Addressed your nits, PTAL

PR #135, opened 2018-03-19
return terminal errors when working with cloud providers

p0lyn0mial:
**What this PR does / why we need it**:  this pull extends cloud providers in a way that they can return terminal errors that indicate if a user intervention is required to recover from faulty situation.
for example a user may provide incorrect credentials which would prevent the system from instantiating or deleting any machines in the cloud. those errors are part of `Machine.Status` and can make their way to the end user.

Fixes #

### DIFF
--- a/pkg/cloudprovider/provider/digitalocean/provider.go
+++ b/pkg/cloudprovider/provider/digitalocean/provider.go
@@ -241,11 +241,14 @@ func uploadSSHPublicKey(ctx context.Context, service godo.KeysService, key *rsa.
 		return "", fmt.Errorf("failed to create ssh public key, the key already exists")
 	}
 
-	newDoKey, _, err := service.Create(ctx, &godo.KeyCreateRequest{
+	newDoKey, rsp, err := service.Create(ctx, &godo.KeyCreateRequest{
 		PublicKey: string(ssh.MarshalAuthorizedKey(pk)),
 		Name:      string(uuid.NewUUID()),
 	})
 	if err != nil {
+		if terminalError := httpStatusToTerminalError(rsp.StatusCode); terminalError != nil {
+			return "", terminalError
+		}
 		return "", fmt.Errorf("failed to create ssh public key on digitalocean: %v", err)
 	}
 
@@ -255,7 +258,10 @@ func uploadSSHPublicKey(ctx context.Context, service godo.KeysService, key *rsa.
 func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.Instance, error) {
 	c, pc, err := p.getConfig(machine.Spec.ProviderConfig)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse config: %v", err)
+		return nil, cloudprovidererrors.TerminalError{
+			Reason:  v1alpha1.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Failed to parse MachineSpec, due to %v", err),
+		}
 	}
 
 	ctx := context.TODO()
@@ -271,7 +277,7 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 	}
 	fingerprint, err := uploadSSHPublicKey(ctx, client.Keys, &tmpRSAKeyPair.PublicKey)
 	if err != nil {
-		return nil, fmt.Errorf("failed upload the ssh key, error = %v", err)
+		return nil, err
 	}
 	defer func() {
 		_, err := client.Keys.DeleteByFingerprint(ctx, fingerprint)
@@ -282,7 +288,10 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 
 	slug, err := getSlugForOS(pc.OperatingSystem)
 	if err != nil {
-		return nil, fmt.Errorf("invalid operating system specified %q: %v", pc.OperatingSystem, err)
+		return nil, cloudprovidererrors.TerminalError{
+			Reason:  v1alpha1.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Failed to parse MachineSpec, invalid operating system specified %q: %v", pc.OperatingSystem, err),
+		}
 	}
 	createRequest := &godo.DropletCreateRequest{
 		Image:             godo.DropletCreateImage{Slug: slug},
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -354,7 +354,11 @@ func (p *provider) Get(machine *v1alpha1.Machine) (instance.Instance, error) {
 
 	ctx := context.TODO()
 	client := getClient(c.Token)
-	droplets, _, err := client.Droplets.List(ctx, &godo.ListOptions{PerPage: 1000})
+	droplets, rsp, err := client.Droplets.List(ctx, &godo.ListOptions{PerPage: 1000})
+
+	if terminalError := httpStatusToTerminalError(rsp.StatusCode); terminalError != nil {
+		return nil, terminalError
+	}
 	if err != nil {
```
mrIncompetent: Shouldn't the check be within the `if err != nil` block?

p0lyn0mial:
sharing some screen shot that show how the terminal errors make their way to the dashboard.

screen shot 2018-03-20 at 10 15 12


screen shot 2018-03-20 at 10 14 00

mrIncompetent:
That looks awesome!

pkg/controller/machine.go
```diff
@@ -223,22 +223,18 @@ func (c *Controller) updateMachine(machine *machinev1alpha1.Machine) (*machinev1
...
-	return machine, nil
+func (c *Controller) clearMachineErrorIfSet(machine *machinev1alpha1.Machine) (*machinev1alpha1.Machine, error) {
+	machine.Status.ErrorMessage = nil
+	machine.Status.ErrorReason = nil
+	return c.updateMachine(machine)
 }
 
+// updateMachine updates machine's ErrorMessage and ErrorReason regardless if they were set or not.
+// This essentially overwrites previous values.
 func (c *Controller) updateMachineError(machine *machinev1alpha1.Machine, reason machinev1alpha1.MachineStatusError, message string) (*machinev1alpha1.Machine, error) {
```
p0lyn0mial: those errors are not permanent e.g. a user can provide incorrect credentials that would result in setting a `terminal` error during creation but later on a user may wish to delete machine that would result in a `terminal` error during deletion.

pkg/controller/machine.go
```diff
@@ -307,43 +303,7 @@ func (c *Controller) syncHandler(key string) error {
...
-		}
-		glog.V(4).Infof("Removed delete finalizer from machine %s", machine.Name)
-
-		// Remove error message in case it was set
-		if machine, err = c.clearMachineErrorIfSet(machine, machinev1alpha1.DeleteMachineError); err != nil {
-			return fmt.Errorf("failed to update machine after removing the delete error: %v", err)
-		}
-
-		return nil
+		return c.deleteMachineAndProviderInstance(prov, providerInstance, machine)
```
p0lyn0mial: I decided to move deletion logic to a separate function to reduce cyclomatic complexity

pkg/controller/machine.go
```diff
@@ -373,95 +333,176 @@ func (c *Controller) syncHandler(key string) error {
...
+	var err error
+
+	// step 1: delete provider instance
+	if err = c.deleteProviderInstance(prov, machine); err != nil {
+		if ok, _, _ := cloudprovidererrors.IsTerminalError(err); ok {
+			message := fmt.Sprintf("%v. Please manually delete finalizers from the machine object.", err)
+			if _, errNested := c.updateMachineError(machine, machinev1alpha1.DeleteMachineError, message); errNested != nil {
+				return fmt.Errorf("failed to update machine error after failed delete: %v", errNested)
+			}
+			return err
```
p0lyn0mial: we could also return `nil` which is quite reasonable since a `terminal` error shouldn't be requeued.

p0lyn0mial:
> That looks awesome!

yeah, hopefully those messages will help us to streamline troubleshooting. We just have to be careful about what we put into the error messages. As they can reveal too much data which could be used against the system, or they can scare away the users.

pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -342,20 +360,34 @@ func (p *provider) Delete(machine *v1alpha1.Machine) error {
...
 		return fmt.Errorf("failed to convert instance id %s to int: %v", i.ID(), err)
 	}
-	_, err = client.Droplets.Delete(ctx, doID)
-	return err
+
+	rsp, err := client.Droplets.Delete(ctx, doID)
+	if err != nil {
+		if terminalError := httpStatusToTerminalError(rsp.StatusCode); terminalError != nil {
```
p0lyn0mial: I slightly improved the way we return terminal errors in this pull
https://github.com/kubermatic/machine-controller/pull/136

pkg/controller/machine.go
```diff
@@ -374,99 +334,180 @@ func (c *Controller) syncHandler(key string) error {
...
+	glog.V(4).Infof("Removed delete finalizer from machine %s", machine.Name)
+
+	// step 4: remove error message in case it was set
+	if machine, err = c.clearMachineErrorIfSet(machine); err != nil {
+		return fmt.Errorf("failed to update machine after removing the delete error: %v", err)
+	}
+	return nil
+}
+
+func (c *Controller) createNodeForMachine(prov cloud.Provider, providerInstance instance.Instance, machine *machinev1alpha1.Machine, userdataProvider userdata.Provider, providerConfig *providerconfig.Config) error {
```
mrIncompetent: Maybe rename to something like `createCloudProviderInstanceForMachine` or `createInstanceForMachine` ?
`Node` might be a bit confusing as we have the kubernetes-node resource we handle as well

pkg/controller/machine.go
```diff
@@ -374,99 +334,180 @@ func (c *Controller) syncHandler(key string) error {
...
 		}
+
+		// case 1.3: transient error was returned requeue the request and try again in the future
+		return fmt.Errorf("failed to get instance from provider: %v", err)
 	}
 
-	if node != nil {
-		return c.ensureNodeLabelsAnnotationsAndTaints(node, machine)
+	// case 2: retrieving instance from cloudprovider was successfull
+	node, exists, err := c.getNode(providerInstance, string(providerConfig.CloudProvider))
```
mrIncompetent: i would move the code which handles the kubernetes node object into a separate function

pkg/controller/machine.go
p0lyn0mial: I like `createInstanceForMachine`

PR #138, opened 2018-03-23
Add CentOS support

### LINKED ISSUE #137: Add support for Enterprise Linux based distros (RHEL, CentOS)
Aside from RedHat ContainerLinux and Ubuntu, we should also support Enterprise Linux based distros, e.G. CentOS

alvaroaleman:

Adds support for CentOS 7. We will be able to reuse this for RHEL as well, except for image ids and potentially a configuration for subscription.

Fixes #137


pkg/userdata/centos/userdata.go
```diff
@@ -0,0 +1,246 @@
...
+	tmpl, err := template.New("user-data").Funcs(machinetemplate.TxtFuncMap()).Parse(ctTemplate)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse user-data template: %v", err)
+	}
+
+	semverKubeletVersion, err := semver.NewVersion(spec.Versions.Kubelet)
+	if err != nil {
+		return "", fmt.Errorf("invalid kubelet version: '%v'", err)
+	}
+	kubeletVersion := fmt.Sprintf("-%s", semverKubeletVersion.String())
```
mrIncompetent: Move the `-`  to the template. It's not obvious why it should be added here already.
Otherwise it will create confusion when this variable will be used in some other context

pkg/userdata/centos/userdata.go
```diff
@@ -0,0 +1,246 @@
...
+    # This file controls the state of SELinux on the system.
+    # SELINUX= can take one of these three values:
+    #     enforcing - SELinux security policy is enforced.
+    #     permissive - SELinux prints warnings instead of enforcing.
+    #     disabled - No SELinux policy is loaded.
+    SELINUX=permissive
+    # SELINUXTYPE= can take one of three two values:
+    #     targeted - Targeted processes are protected,
+    #     minimum - Modification of targeted policy. Only selected processes are protected.
+    #     mls - Multi Level Security protection.
```
mrIncompetent: I assume those are comments?
Are those necessary?

pkg/userdata/centos/userdata.go
```diff
@@ -0,0 +1,246 @@
...
+- chage -d $(date +%s) root
+- systemctl enable --now kubelet
+
+packages:
+- {{ .DockerPackageName }}
+- kubelet{{ .KubeletVersion }}
+- ebtables
+- ethtool
+- nfs-utils
+- bash-completion # Have mercy for the poor operators
```
mrIncompetent: You sir, deserve this comment!

pkg/userdata/centos/userdata.go
```diff
@@ -0,0 +1,246 @@
...
+      --resolv-conf=/etc/resolv.conf \
+      --rotate-certificates=true \
+      --kubeconfig=/etc/kubernetes/kubeconfig \
+      --bootstrap-kubeconfig=/etc/kubernetes/bootstrap.kubeconfig \
+      --lock-file=/var/run/lock/kubelet.lock \
+      --exit-on-lock-contention \
+      --read-only-port 0 \
+      --authorization-mode=Webhook \
+      --anonymous-auth=false \
+      --client-ca-file=/etc/kubernetes/ca.crt
```
mrIncompetent: We should move to `kubeadm join` in the future to get rid of this duplication.
We have the same set of flags in every user-data :/

pkg/userdata/centos/userdata.go
alvaroaleman: Yes those are comments. They are not necessary for functionality but they are part how the file looks by default

pkg/userdata/centos/userdata.go
alvaroaleman: Opened https://github.com/kubermatic/machine-controller/issues/139 for that

PR #142, opened 2018-03-29
Clear noderef on machine before creating a new instance at the cloud provider

### LINKED ISSUE #140: Correctly set a nodes `ownerRef` after it gets recreated
Right now if a node gets deleted for whatever reason (e.G. manually), the machine-controller will recreate it. This is fine, but results in the node not having an `ownerRef`.

alvaroaleman:

Fixes #140 


pkg/controller/machine.go
```diff
@@ -819,3 +825,15 @@ func (c *Controller) updateMetrics() {
 	c.updateMachinesMetric()
 	c.updateNodesMetric()
 }
+
+func (c *Controller) clearNodeRef(machine *machinev1alpha1.Machine) (*machinev1alpha1.Machine, error) {
+	var err error
```
p0lyn0mial: nit: could be defined inside the `if` statement.

PR #144, opened 2018-03-30
 validates machine's specification before doing anything else with the machine

### LINKED ISSUE #131: deadlock when trying to delete a machine
Steps to reproduce :
1. Create an invalid machine - you can use the following manifest that doesn't specify required credentials https://github.com/kubermatic/machine-controller/blob/master/examples/machine-digitalocean.yaml
2. Delete the previously created machine.
3. List machine resources

Result:
The machine was not deleted and the server keeps saying `machine1 failed with: failed to get instance for machine machine1 after the delete got triggered`
The only way of getting out of this situation is to manually edit the machine's spec and remove finalizers.

In general the described state exists because we add finalizers to a machine before creating a node because we want to prevent deletion of a machine resource. 

As I can imagine that the call that requests a node can fail for many reasons, I think that this issue could help us track discussion on possible solutions to this issue.

p0lyn0mial:
**What this PR does / why we need it**: moves code that validates the machine spec before getting the instance from cloud provider even though this is a little bit premature and inefficient, it helps us detect invalid specification as soon as possible.

Fixes #131


alvaroaleman:
Oh this is actually really cool and makes the creation workflow more intuitive

pkg/controller/machine.go
```diff
@@ -410,37 +410,41 @@ func (c *Controller) deleteMachineAndProviderInstance(prov cloud.Provider, provi
 }
 
 func (c *Controller) createInstanceForMachine(prov cloud.Provider, providerInstance instance.Instance, machine *machinev1alpha1.Machine, userdataProvider userdata.Provider, providerConfig *providerconfig.Config) error {
+	// case 1: validate the machine spec before getting the instance from cloud provider.
+	// even though this is a little bit premature and inefficient, it helps us detect invalid specification
```
alvaroaleman: I don't even think its a big difference in efficiency , because `createInstanceForMachine` is only called when `!nodeExistsAndIsReady`

pkg/controller/machine.go
```diff
@@ -410,37 +410,41 @@ func (c *Controller) deleteMachineAndProviderInstance(prov cloud.Provider, provi
...
+	{
```
alvaroaleman: What's the advantage of putting this into a dedicated block?

p0lyn0mial: some variables are invisible outside of this block and their lifetime is limited also to this block.

alvaroaleman: Ah okay, I personally am not too fond of that, because my monitor has a limited width and having a block that is not strictly required takes away some of that width :P

p0lyn0mial: @alvaroaleman  glad to hear you have a monitor :) 
@alvaroaleman  no problem I can remove it.

pkg/controller/machine.go
```diff
@@ -354,7 +353,7 @@ func (c *Controller) syncHandler(key string) error {
 	}
 	if !nodeExistsAndIsReady {
 		glog.V(6).Infof("Requesting instance for machine '%s' from cloudprovider because no associated node with status ready found...", machine.Name)
-		err = c.createInstanceForMachine(prov, providerInstance, machine, userdataProvider, providerConfig)
```
p0lyn0mial: i just realised that there is no need to pass `providerInstance` as an arg.

pkg/controller/machine.go
```diff
@@ -394,48 +394,51 @@ func (c *Controller) deleteMachineAndProviderInstance(prov cloud.Provider, provi
 	}
 	glog.V(4).Infof("Deleted machine %s at cloud provider", machine.Name)
 
-	// step 3: remove finalizers this essentially will remove the machine object from the system
+	// step 3: remove error message in case it was set
```
p0lyn0mial: I moved `step 3` before `step 4` because it caused errors like this one:

` failed to update machine after removing the delete error: Operation cannot be fulfilled on machines.machine.k8s.io "machine-kubermatic-z5psw2ws48-m9hh5": StorageError: invalid object, Code: 4, Key: /registry/machine.k8s.io/machines/machine-kubermatic-z5psw2ws48-m9hh5, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 389cc5b7-3732-11e8-bdd4-0a580a2c021a, UID in object meta: `

removing finalizers can cause the object to be garbage collected.

p0lyn0mial:
@alvaroaleman, @mrIncompetent I have made some changes, PTAL

pkg/controller/machine.go
alvaroaleman: I noticed that error as well while implementing vsphere, the following patch fixed it:

```
commit 7f95e3bb920db0aea2c4ae5b7c5316435c0b121d
Author: Alvaro Aleman <mailaddr>
Date:   Sat Mar 31 17:18:55 2018 +0200

    Only remove machine error if set

diff --git a/pkg/controller/machine.go b/pkg/controller/machine.go
index d9e039e..8348079 100644
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -227,9 +227,12 @@ func (c *Controller) updateMachine(machine *machinev1alpha1.Machine) (*machinev1
 }
 
 func (c *Controller) clearMachineErrorIfSet(machine *machinev1alpha1.Machine) (*machinev1alpha1.Machine, error) {
-       machine.Status.ErrorMessage = nil
-       machine.Status.ErrorReason = nil
-       return c.updateMachine(machine)
+       if machine.Status.ErrorMessage != nil || machine.Status.ErrorReason != nil {
+               machine.Status.ErrorMessage = nil
+               machine.Status.ErrorReason = nil
+               return c.updateMachine(machine)
+       }
+       return machine, nil
```

pkg/controller/machine.go
```diff
@@ -369,8 +368,9 @@ func (c *Controller) syncHandler(key string) error {
 }
 
 // deleteMachineAndProviderInstance makes sure that an instance has gone in a series of steps.
-func (c *Controller) deleteMachineAndProviderInstance(prov cloud.Provider, providerInstance instance.Instance, machine *machinev1alpha1.Machine) error {
+func (c *Controller) deleteMachineAndProviderInstance(prov cloud.Provider, machine *machinev1alpha1.Machine) error {
 	var err error
+	var providerInstance instance.Instance
```
p0lyn0mial: this cost me `nil pointer dereference` - golang shadowing :)

https://circleci.com/gh/kubermatic/machine-controller/2231?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

pkg/controller/machine.go
```diff
@@ -279,7 +279,6 @@ func (c *Controller) validateMachine(prov cloud.Provider, machine *machinev1alph
 }
 
 func (c *Controller) syncHandler(key string) error {
-	var providerInstance instance.Instance
```
alvaroaleman: Nice, that caused some NPEs which is why I added a `== nil` check for it in `ensuereNodeOwnerRefAndConfigSource`

alvaroaleman: Oh it just got moved :S

pkg/controller/machine.go
p0lyn0mial: in this case it might be a little bit different, since removing finalizers from the object can actually remove the object from the system.

mrIncompetent:
@alvaroaleman your points have been addressed.
Are you able to lgtm?

p0lyn0mial:
do you know why `ci/circleci: install-dependencies` step is pending ?

mrIncompetent:
Probably because this PR was opened before we removed the install-dependencies step from the circle-ci config.
This should not happen with future PR's - as long as the branch got created from a up-to-date master branch.

@p0lyn0mial i'll force merge if you are ok?

p0lyn0mial:
@mrIncompetent oh, I see. Yes, please merge it. Thanks !

PR #148, opened 2018-03-31
Feature/vmware

### LINKED ISSUE #50: integrate vmware vsphere

alvaroaleman:

Fixes #50 


alvaroaleman:
@mrIncompetent Ready for review

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -0,0 +1,363 @@
...
+	localUserdataIsoFilePath, err := generateLocalUserdataIso(userdata, machine.Spec.Name)
+	if err != nil {
+		return nil, err
+	}
+	err = uploadAndAttachISO(finder, virtualMachine, localUserdataIsoFilePath, config.Datastore, client)
+	if err != nil {
+		return nil, fmt.Errorf("failed to upload and attach userdata iso: %v", err)
+	}
+
+	err = os.Remove(localUserdataIsoFilePath)
```
mrIncompetent: Maybe put this into a `defer` function so we always cleanup the created ISO ?

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -0,0 +1,289 @@
...
+
+	command := "genisoimage"
+	args := []string{"-o", isoFilePath, "-volid", "cidata", "-joliet", "-rock", userdataDir}
+	cmd := exec.Command(command, args...)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("error executing command `%s %s`: output: `%s`, error: `%v`", command, args, string(output), err)
+	}
+
+	err = os.RemoveAll(userdataDir)
```
mrIncompetent: Maybe put this into a `defer` function so we always cleanup in the end?

docs/vsphere.md
```diff
@@ -0,0 +1,13 @@
+# VMWware vsphere
+
+To use the machine-controller to create machines on VMWare vsphere, you must first
+create a template:
+
+1. Go into the VSphere WebUI, select your datacenter, right click onto it and choose "Deploy OVF Template"
+2. Enter `https://cloud-images.ubuntu.com/releases/16.04/release/ubuntu-16.04-server-cloudimg-amd64.ova` into the
```
mrIncompetent: Please add a line for centos and coreos

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -0,0 +1,363 @@
...
+type RawConfig struct {
+	TemplateVMName providerconfig.ConfigVarString `json:"templateVMName"`
+	Username       providerconfig.ConfigVarString `json:"username"`
+	Password       providerconfig.ConfigVarString `json:"password"`
+	VSphereURL     providerconfig.ConfigVarString `json:"vsphereURL"`
+	Datacenter     providerconfig.ConfigVarString `json:"datacenter"`
+	Cluster        providerconfig.ConfigVarString `json:"cluster"`
+	Datastore      providerconfig.ConfigVarString `json:"datastore"`
+	CPUs           int32                          `json:"cpus"`
+	MemoryMB       int64                          `json:"memoryMB"`
```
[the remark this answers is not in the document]
alvaroaleman: Well that is obviously a matter of opinion, but I prefer a self-explanatory API over one that looks nice

docs/vsphere.md
alvaroaleman: Those are currently not tested on vsphere, I will add docs once that's done (probably not within this PR)

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -236,6 +236,12 @@ func generateLocalUserdataIso(userdata, name string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("failed to create local temp directory for userdata at %s: %v", userdataDir, err)
 	}
+	defer func(dir string) {
+		err := os.RemoveAll(userdataDir)
+		if err != nil {
+			glog.Errorf("error cleaning up local userdata tempdir %s: %v", userdataDir, err)
```
mrIncompetent: You have a parameter for your defer func but still use the `userdataDir` variable from the parent scope.
Just relying on the parent scope is fine here. No need to specify a parameter

examples/machine-vsphere.yaml
```diff
@@ -0,0 +1,51 @@
...
+    name: node1
+    labels:
+      foo: "bar"
+  providerConfig:
+    sshPublicKeys:
+      - "<< YOUR_PUBLIC_KEY >>"
+    cloudProvider: "vsphere"
+    cloudProviderSpec:
+      templateVMName: ubuntu-template
+      username: '<VSPHERE_USERNAME>'
```
p0lyn0mial: please change it to `<< VSPHERE_USERNAME >>` as this seems to be our "naming convention"
this also applies to `<< VSPHERE_ADDRESS >>`

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -0,0 +1,366 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.Instance, error) {
+	config, _, err := p.getConfig(machine.Spec.ProviderConfig)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse config: %v", err)
```
p0lyn0mial: some time ago we introduced the concept of `terminal` errors. Please change the following line to return a terminal error. This also applies to other places in the code.

examples:
https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/aws/provider.go#L497
https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/aws/provider.go#L608

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -0,0 +1,366 @@
...
+
+	localUserdataIsoFilePath, err := generateLocalUserdataIso(userdata, machine.Spec.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		err := os.Remove(localUserdataIsoFilePath)
+		if err != nil {
+			glog.Errorf("failed to clean up local userdata iso file at %s: %v", localUserdataIsoFilePath, err)
```
p0lyn0mial: consider using `utilruntime.HandleError`

https://github.com/kubermatic/machine-controller/blob/master/pkg/controller/machine.go#L200

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -0,0 +1,366 @@
...
+	err = removeFloppyDevice(virtualMachine)
+	if err != nil {
+		return nil, err
+	}
+
+	powerOnTask, err := virtualMachine.PowerOn(context.TODO())
+	if err != nil {
+		return nil, fmt.Errorf("failed to power on machine: %v", err)
+	}
+	powerOnTask.Wait(context.TODO())
```
p0lyn0mial: should we set some sane timeout for this operation ?

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -0,0 +1,366 @@
...
+
+	glog.V(2).Infof("Successfully created a vm with name '%s' and id '%s'", machine.Spec.Name, vmID)
```
p0lyn0mial: not sure if we need to log this message, it will be put into the log anyway by the controller
https://github.com/kubermatic/machine-controller/blob/master/pkg/controller/machine.go#L470

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	powerState, err := virtualMachine.PowerState(context.TODO())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get powerstate: %v", err)
+	}
+
+	var status instance.Status
+	switch powerState {
+	case "poweredOn":
```
p0lyn0mial: already defined - use `VirtualMachinePowerStatePoweredOn` const

test/tools/integration/create-and-destroy-machine.sh
```diff
@@ -7,14 +7,17 @@ cd $(dirname $0)
...
 
 ssh_exec() { ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@$ADDR $@; }
 cat <<EOF |ssh_exec
 set -e
 echo "Testing create of a node via machine-controller...."
 ./verify \
-  -input examples/machine-hetzner.yaml \
+  -input examples/machine-$TYPE.yaml \
   -parameters "<< HETZNER_TOKEN >>=$HZ_TOKEN,cri-o=docker,node1=testnode-${CIRCLE_BUILD_NUM:-local},ubuntu=${OS}" \
+  -parameters "<< VSPHERE_PASSWORD >>=${VSPHERE_PASSWORD:-undef},<VSPHERE_USERNAME>=${VSPHERE_USERNAME:-undef}" \
```
p0lyn0mial: I think we also need to fix `-parameters`, at the moment the one defined in line 18 is overwritten

pkg/cloudprovider/provider/vsphere/provider.go
alvaroaleman: I guess we could, but it would require some effort since that is not supported natively, so we would have to write some wrapper that calls `powerOnTask.Wait` in a dedicated goroutine.

Do you think that's worth it? Or is there an easier approach?

p0lyn0mial: we could use a context that has timeout set.
https://golang.org/pkg/context/#WithTimeout

test/tools/integration/create-and-destroy-machine.sh
alvaroaleman: That should work just fine, check out the changes I made to the verify tool

p0lyn0mial: you are right, I missed that one. Nice !

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -0,0 +1,402 @@
...
+	defer cancel()
+	err = powerOnTask.Wait(powerOnTaskContext)
+	if err != nil {
+		return nil, fmt.Errorf("timed out waiting to power on vm %s: %v", virtualMachine.Name(), err)
+	}
+
+	return VSphereServer{name: virtualMachine.Name(), status: instance.StatusRunning, id: virtualMachine.Reference().Value}, nil
+}
+
+func (p *provider) Delete(machine *v1alpha1.Machine) error {
```
alvaroaleman: I've not used TerminalErrors here, because if something fails here its most likely due to the fact that something was deleted in vsphere manually. The only exceptions to that are `getConfig` and `getClient` which return TerminalErrors by themselves

alvaroaleman:
@p0lyn0mial PTAL, I think I've addressed all your remarks :)

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -0,0 +1,402 @@
...
+		return nil, nil, machineSpecParsingTerminalError(err)
+	}
+
+	c.CPUs = rawConfig.CPUs
+	c.MemoryMB = rawConfig.MemoryMB
+
+	return &c, &pconfig, nil
+}
+
+func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
```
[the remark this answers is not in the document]
alvaroaleman: Ah okay, this is pretty confusing tbh because it means there are some places in a provider where we must return TerminalErrors and some where we must not. Also it sounds to me as if `Create` should not return a TerminalError, because `Validate` contains all possible validations already and if create fails, the controller will retry which means it will first execute `Validate`.

What do you think about keeping it as simple as possible and use `Validate` as the only func that returns TerminalErrors? We could even reflect that in its signature.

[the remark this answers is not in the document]
p0lyn0mial: I really like this idea. I think we could follow this path only if we are sure that things are constant and don't change in the system. That is, there are not updates to manifest, user credentials and things inside cloud provider after a machine has been created (we called `Validate` method) and before we call `Delete` method. I am not sure if we can guarantee that ?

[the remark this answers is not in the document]
alvaroaleman: Well, we could run `Validate` under some more conditions, e.G.:

* `Get` fails and the error is not `instanceNotFound`
* `Delete` fails (Not sure if that makes sense)

But if we do that it should IMHO be a separate PR.

For this PR, is it okay if I:

* Remove the terminalErrors from Validate
* Thror them in `Create` only if appropriate

?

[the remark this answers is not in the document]
p0lyn0mial: > For this PR, is it okay if I:
> Remove the terminalErrors from Validate
> Throw them in only Create if appropriate

yes.

alvaroaleman:
@mrIncompetent @p0lyn0mial I think I updated everything requested and e2e passes again, so PTAL  :)

test/tools/integration/create-and-destroy-machine.sh
```diff
@@ -7,14 +7,17 @@ cd $(dirname $0)
...
 echo "Testing create of a node via machine-controller...."
 ./verify \
-  -input examples/machine-hetzner.yaml \
+  -input examples/machine-$TYPE.yaml \
   -parameters "<< HETZNER_TOKEN >>=$HZ_TOKEN,cri-o=docker,node1=testnode-${CIRCLE_BUILD_NUM:-local},ubuntu=${OS}" \
+  -parameters "<< VSPHERE_PASSWORD >>=${VSPHERE_PASSWORD:-undef},<< VSPHERE_USERNAME >>=${VSPHERE_USERNAME:-undef}" \
+  -parameters "<< VSPHERE_ADDRESS >>=${VSPHERE_ADDRESS:-undef}" \
   -logtostderr true || (kubectl logs -n kube-system \$(kubectl get pods \
       -n kube-system|egrep '^machine-con'|awk '{ print \$1 }'); exit 1)
 #./verify \
```
p0lyn0mial: sorry, I have just noticed this commented code :)

p0lyn0mial:
@alvaroaleman  great, pull. Brings huge value to the system. 
@alvaroaleman thanks for addressing my comment, I have just added one more and this is lgtm.

mrIncompetent:
after addressing @p0lyn0mial last comment, lgtm

Issue #129, opened 2018-03-12
Extend e2e testing


mrIncompetent:
We should add the following test cases:
- Hetzner
  - Ubuntu + Docker 1.13
  - Ubuntu + Docker 17.03
  - Ubuntu + CRI-O 1.9
- Digitalocean
  - Ubuntu + Docker 1.13
  - Ubuntu + Docker 17.03
  - Ubuntu + CRI-O 1.9
  - CoreOS + Docker 1.13
  - CoreOS + Docker 17.03
- AWS
  - Ubuntu + Docker 1.13
  - Ubuntu + Docker 17.03
  - Ubuntu + CRI-O 1.9
  - CoreOS + Docker 1.13
  - CoreOS + Docker 17.03
- Openstack (We need a sponsor here)
  - Ubuntu + Docker 1.13
  - Ubuntu + Docker 17.03
  - Ubuntu + CRI-O 1.9
  - CoreOS + Docker 1.13
  - CoreOS + Docker 17.03

alvaroaleman:
To keep tests still efficient and also not repeat ourselves too much I suggest putting this in the verify tool by:

* Adding a config file that allows to specify a matrix just like the one above
* Add a `--create-only` flag that only creates the machine objects so they will be created in parallel
* Add a `--wait-for=<provider>-<distro>-<container-runtime>` flag that blocks until the given machine was successfully created and then deletes it to see at a glance in CircleCI what failed if something failed

mrIncompetent:
Why not implementing decent concurrency into the command?
Might be cleaner than just juggling with different command line calls

alvaroaleman:
Because then you can't have a dedicated step per scenario in CircleCI which helps a lot if tests failed and you want to find out which one

mrIncompetent:
We could also do the setup (terraform & cluster creation) in one step, and then run all tests in parallel.
Then we don't have to modify the command at all.

p0lyn0mial:
@alvaroaleman @mrIncompetent Could we slightly change approach to writing test cases ?

I'd like to start writing test cases as standard golang tests. 
But at the same I'd like to utilise what we have so far, that is: cluster creation (kubeadm) and the verify tool. The whole process would do the following steps:
 1. create cluster (provision_master.sh)
 2. run machine-controller (provision_master.sh)
 3. run e2e test (using standard `go test` command)

[...]

PR #150, opened 2018-04-04
verify tool checks ownerReferences

p0lyn0mial:
**What this PR does / why we need it**: verify tool checks ownerReferences to determine if a node for a machine has been created. This allows for running the verify tool multiple times.

Fixes #

**Special notes for your reviewer**: see also #129

### DIFF
--- a/test/tools/verify/cmd/main.go
+++ b/test/tools/verify/cmd/main.go
@@ -24,8 +24,8 @@ import (
 )
 
 const (
-	machineReadyCheckPeriod  = 2 * time.Second
-	machineReadyCheckTimeout = 5 * time.Minute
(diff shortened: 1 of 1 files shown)


test/tools/verify/cmd/main.go
```diff
@@ -24,8 +24,8 @@ import (
 )
 
 const (
-	machineReadyCheckPeriod  = 2 * time.Second
-	machineReadyCheckTimeout = 5 * time.Minute
+	machineReadyCheckPeriod  = 5 * time.Second
+	machineReadyCheckTimeout = 10 * time.Minute
```
p0lyn0mial: I have increased timeous, as the verify tool failed when more that one machine was being created at the same time

PR #153, opened 2018-04-10
reads some fields of cloudProviderSpec from env variables.

### LINKED ISSUE #123: Add support for accepting cloud-provider credentials as EnvVar's
The Machine object accepts multiple sources for cloudProviderSpec fields:
- Direct value 
```yaml
...
spec:
...
  providerConfig:
    cloudProvider: "aws"
    cloudProviderSpec:
      accessKeyId: "foo"
```
- Secret ref 
```yaml
...
spec:
...
  providerConfig:
    cloudProvider: "aws"
    cloudProviderSpec:
      accessKeyId:
        secretKeyRef:
          namespace: kube-system
          name: machine-controller-aws
          key: accessKeyId
```
- ConfigMap ref 
```yaml
...
spec:
...
  providerConfig:
    cloudProvider: "aws"
    cloudProviderSpec:
      accessKeyId:
        configMapKeyRef:
          namespace: kube-system
          name: machine-controller-aws
          key: accessKeyId
```

It should also be possible to pass in the secret values implicitly as environment variable.
The secret values differ from cloud provider.
- AWS
  - Access Key ID
  - Secret Access Key
- Hetzner
  - Token
- Digitalocean
  - Token
- OpenStack
  - Username
  - Password

Each secret field needs to have one specific environment key. Like `AWS_ACCESS_KEY_ID`.
During the processing of the `cloudProviderSpec` we would need to check if the environment variable is set, and if so we need to use this value.

Reason: In scenarios where the master components is managed by an external entity (Loodse kubermatic/ SAP Gardener) it might not be possible to expose the cloud provider specific secrets to the users.

p0lyn0mial:
**What this PR does / why we need it**: reads some fields of cloudProviderSpec from env variables.
the names of variables are hardcoded in the source code and cannot be changed.
there are other mechanisms through which the fields can be set, we only fallback
to reading from env variables when all fails.

Fixes #123 

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -95,25 +95,25 @@ func (p *provider) getConfig(s runtime.RawExtension) (*Config, *providerconfig.C
 		return nil, nil, nil, err
 	}
 	c := Config{}
-	c.IdentityEndpoint, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.IdentityEndpoint)
+	c.IdentityEndpoint, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.IdentityEndpoint, "OS_AUTH_URL")
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"identityEndpoint\" field, error = %v", err)
 	}
-	c.Username, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Username)
+	c.Username, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Username, "OS_USER_NAME")
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"username\" field, error = %v", err)
 	}
-	c.Password, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Password)
+	c.Password, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Password, "OS_PASSWORD")
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"password\" field, error = %v", err)
 	}
-	c.DomainName, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.DomainName)
+	c.DomainName, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.DomainName, "OS_DOMAIN_NAME")
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"domainName\" field, error = %v", err)
 	}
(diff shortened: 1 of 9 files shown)


p0lyn0mial:
TODO: @p0lyn0mial update the documentation

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -99,13 +99,13 @@ func (p *provider) getConfig(s runtime.RawExtension) (*Config, *providerconfig.C
...
-		return nil, nil, nil, err
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"username\" field, error = %v", err)
 	}
-	c.Password, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Password)
+	c.Password, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Password, "OS_PASSWORD")
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"password\" field, error = %v", err)
 	}
 	c.DomainName, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.DomainName)
```
alvaroaleman: Could you also add `OS_AUTH_URL` for `identityEndpoint`, `OS_TENANT_NAME` for `tenantName` and `OS_DOMAIN_NAME` for `domainName` so we have the full set of auth-related env vars?

pkg/providerconfig/types.go
```diff
@@ -159,6 +161,21 @@ func (configVarResolver *ConfigVarResolver) GetConfigVarStringValue(configVar Co
...
+// getting the value from an environment variable specified by envVarName parameter
+func (configVarResolver *ConfigVarResolver) GetConfigVarStringValueOrEnv(configVar ConfigVarString, envVarName string) (string, error) {
+	cfgVar, err := configVarResolver.GetConfigVarStringValue(configVar)
+	if err == nil && len(cfgVar) > 0 {
+		return cfgVar, err
+	}
+
+	envVal, envValFound := os.LookupEnv(envVarName)
+	if !envValFound {
+		return "", fmt.Errorf("all machanisms(secret, configMap) of getting the value failed, including reading from environment varialbe = %s which was not set", envVarName)
```
alvaroaleman: `s/varialbe/variable/g`

alvaroaleman:
I must say I really like this change, finally we can define a global auth info :)

As for the docs, maybe it would be easiest to just add a comment to the respective fields in the example yamls?

pkg/cloudprovider/provider/openstack/provider.go
p0lyn0mial: Many thanks for pointing this out. I missed them.

pkg/providerconfig/types.go
p0lyn0mial: not sure if I understand that :)

p0lyn0mial:
@alvaroaleman thanks for the review, comments addressed. PTAL

mrIncompetent:
That is super awesome! Thank you very much for the very fast implementation!

alvaroaleman:
@p0lyn0mial `s/a/b` is a short notion of "Please replace `a` with `b`, in the case above there is a spelling mitaske, its not `varialbe` but `variable`

PR #159, opened 2018-04-12
implements E2E tests for machine controller

p0lyn0mial:
at the moment we have test scenarios for DO and AWS.
tests can be run in two modes -short and normal.
when run in -short mode only a subset of tests will be tested.
by default e2e tests are disabled, we require special build tag to be provided.

Fixes #

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,8 +94,8 @@ jobs:
           name: Test create and destroy ubuntu on vsphere
           command: ./test/tools/integration/create-and-destroy-machine.sh ubuntu vsphere
       - run:
-          name: Test create and destroy ubuntu machine
-          command: ./test/tools/integration/create-and-destroy-machine.sh
+          name:  Run E2E test
+          command: ./test/tools/integration/run_e2e_tests.sh
       - run:
           name: Test create and destroy centos machine
           command: ./test/tools/integration/create-and-destroy-machine.sh centos
--- a/test/e2e/provisioning/aws_e2e_test.go
+++ b/test/e2e/provisioning/aws_e2e_test.go
@@ -1,0 +1,102 @@
+// +build e2e
+
+package provisioning
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+)
+
+const (
+	aws_manifest = "./testdata/machine-aws.yaml"
+)
+
+var awsScenarios = []struct {
+	// name holds short description of the test scenario, it is also used to create machines and nodes names
+	// so please don't put "strange" characters there
+	name                    string
+	osName                  string
+	containerRuntime        string
+	containerRuntimeVersion string
+	// if the -short flag was provided and this variable is set
+	// the test scenario will be run otherwise it will be skipped.
+	short bool
+}{
+	{
+		name:                    "scenario 1 Ubuntu Docker 1.13",
+		osName:                  "ubuntu",
+		containerRuntime:        "docker",
+		containerRuntimeVersion: "1.13",
+		short: true,
+	},
+
+	{
+		name:                    "scenario 2 Ubuntu Docker 17.03",
+		osName:                  "ubuntu",
+		containerRuntime:        "docker",
+		containerRuntimeVersion: "17.03",
+	},
+
+	{
+		name:                    "scenario 3 Ubuntu CRI-O 1.9",
+		osName:                  "ubuntu",
+		containerRuntime:        "cri-o",
+		containerRuntimeVersion: "1.9",
+	},
+
+	{
+		name:                    "scenario 4 CoreOS Docker 1.13",
+		osName:                  "coreos",
+		containerRuntime:        "docker",
+		containerRuntimeVersion: "1.13",
+	},
+
+	{
+		name:                    "scenario 5 CoreOS Docker 17.03",
+		osName:                  "coreos",
+		containerRuntime:        "docker",
+		containerRuntimeVersion: "17.03",
+	},
+}
+
+// TestAWSProvisioning - a test suite that exercises AWS provider
+// by requesting nodes with different combination of container runtime type, container runtime version and the OS flavour.
+func TestAWSProvisioningE2E(t *testing.T) {
+	if testing.Short() {
+		t.Parallel()
+	}
+
+	// test data
+	awsKeyID := os.Getenv("AWS_E2E_TESTS_KEY_ID")
+	awsSecret := os.Getenv("AWS_E2E_TESTS_SECRET")
+	if len(awsKeyID) == 0 || len(awsSecret) == 0 {
+		t.Fatal("unable to run the test suite, AWS_E2E_TESTS_KEY_ID or AWS_E2E_TESTS_SECRET environment variables cannot be empty")
+	}
+
+	// act
+	for _, scenario := range awsScenarios {
+		scenario := scenario // capture range variable
+		t.Run(scenario.name, func(t *testing.T) {
+			if testing.Short() == true && scenario.short == false {
+				t.SkipNow()
+			}
+			t.Parallel()
+
+			nameSufix := strings.Replace(scenario.name, " ", "-", -1)
+			nameSufix = strings.ToLower(nameSufix)
+			nameSufix = fmt.Sprintf("aws-%s", nameSufix)
+			machineName := fmt.Sprintf("machine-%s", nameSufix)
+			nodeName := fmt.Sprintf("node-%s", nameSufix)
+			params := fmt.Sprintf("<< AWS_ACCESS_KEY_ID >>=%s,<< AWS_SECRET_ACCESS_KEY >>=%s", awsKeyID, awsSecret)
+			params = fmt.Sprintf("%s,<< MACHINE_NAME >>=%s,<< NODE_NAME >>=%s", params, machineName, nodeName)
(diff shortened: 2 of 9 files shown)


test/e2e/provisioning/aws_e2e_test.go
```diff
@@ -0,0 +1,102 @@
+// +build e2e
```
alvaroaleman: What is the advantage of putting this into a dedicated tool?

test/e2e/provisioning/aws_e2e_test.go
```diff
@@ -0,0 +1,102 @@
...
+	// so please don't put "strange" characters there
+	name                    string
+	osName                  string
+	containerRuntime        string
+	containerRuntimeVersion string
+	// if the -short flag was provided and this variable is set
+	// the test scenario will be run otherwise it will be skipped.
+	short bool
+}{
+	{
```
alvaroaleman: Wouldn'it it make more sense to have a `Scenerios` struct that accepts a `providerConfig` and can be re-used across different providers?

alvaroaleman: Escpecially since we most likely want to run the same set of tests for each provider

test/e2e/provisioning/aws_e2e_test.go
```diff
@@ -0,0 +1,102 @@
...
+				t.SkipNow()
+			}
+			t.Parallel()
+
+			nameSufix := strings.Replace(scenario.name, " ", "-", -1)
+			nameSufix = strings.ToLower(nameSufix)
+			nameSufix = fmt.Sprintf("aws-%s", nameSufix)
+			machineName := fmt.Sprintf("machine-%s", nameSufix)
+			nodeName := fmt.Sprintf("node-%s", nameSufix)
+			params := fmt.Sprintf("<< AWS_ACCESS_KEY_ID >>=%s,<< AWS_SECRET_ACCESS_KEY >>=%s", awsKeyID, awsSecret)
```
alvaroaleman: I must say that I do not like very much this string replacing and would prefer it if you could just do something along the lines of  `-overwrite=spec.providerconfig.accessKeyID.Value=MYKEY`. Maybe now is a good time?

.circleci/config.yml
```diff
@@ -91,8 +91,8 @@ jobs:
           name: Create kubeadm cluster and install machine-controller onto it
           command: ./test/tools/integration/provision_master.sh
       - run:
-          name: Test create and destroy ubuntu machine
-          command: ./test/tools/integration/create-and-destroy-machine.sh
+          name:  Run E2E test
+          command: ./test/tools/integration/run_e2e_tests.sh
       - run:
```
alvaroaleman: The following step needs to be removed ("Test create and destroy centos machine") and moved into the tool

test/tools/integration/provision_master.sh
```diff
@@ -53,6 +53,11 @@ if ! ls kube-flannel.yml; then
   curl -LO https://raw.githubusercontent.com/coreos/flannel/v0.9.1/Documentation/kube-flannel.yml
   kubectl apply -f kube-flannel.yml
 fi
+if ! which go; then
```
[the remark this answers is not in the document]
alvaroaleman: Just out of curiosity, why does it take so long, shouldn't it be just as parallelized as possible and t(e2eTest) == t(SlowestCreateAndDestroyMachine)?

.circleci/config.yml
p0lyn0mial: `create-and-destroy-machine.sh` exercises `HETZNER` provider, whereas `run_e2e_tests.sh` exercises `DO` and `AWS`

keeping `create-and-destroy-machine.sh` is temporary. I'll create an issue for that. Just wanted to keep this pull small.

test/tools/integration/provision_master.sh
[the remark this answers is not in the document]
p0lyn0mial: To be honest I don't know :) Haven't got much time to debug it. We could definitely follow that. I suspect it is a combination of too much concurrency and  the arbitrary time it takes to create/delete machine.

for example the following snippet shows that tests are paused and resumed, preemption ?
```
=== RUN   TestAWSProvisioningE2E/scenario_5_CoreOS_Docker_17.03
=== PAUSE TestAWSProvisioningE2E/scenario_5_CoreOS_Docker_17.03
=== CONT  TestAWSProvisioningE2E/scenario_1_Ubuntu_Docker_1.13
```

test/e2e/provisioning/aws_e2e_test.go
p0lyn0mial: I agree it doesn't look pretty :( Those are the parameters that are passed to the `verify` tool. The `verify` tool can replace arbitrary fields in the manifest based on the input. 

We can refactor this behaviour but I'd rather do this in a separate pull request. How about that ?

test/e2e/provisioning/aws_e2e_test.go
p0lyn0mial: I think that table driven test are quite elegant. Provider specific tests only prepare input for the `verify` tool, the rest of the logic is shared across providers.

test/e2e/provisioning/aws_e2e_test.go
mrIncompetent: I would like to get rid of the verify tool and move it completely into the tests suite.
It feels weird having a go-native test suite which does a exec to another tool, which does the actual testing

mrIncompetent: But a separate PR seem a good place

alvaroaleman: I was about to suggest to get rid of the verify tool as well, so I agree there :) But its totally fine to do that in a dedicated PR

test/e2e/provisioning/aws_e2e_test.go
alvaroaleman: I am not sure I understand your comment. We have the same table of tests for all providers, which is also reflected in the code here, as there is `awsScenarios ` and `digitalOceanScenarios`, both with the same content - I think it makes more sense to have a global struct `scenarios`  (or `e2eScenenarios` maybe) which is imported and used by all provider

test/tools/integration/provision_master.sh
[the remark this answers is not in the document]
p0lyn0mial: do you know, how much `CPU, Memory` the machine we are running test on have ?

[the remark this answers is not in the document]
alvaroaleman: Ah okay, I think we need to investigate that at some point, because IMHO running the tests scheduled at night instead of Pre/Post-merge is a bad idea, because there is no clear responsibility anymore in case they fail and 1h is definitely too long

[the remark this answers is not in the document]
alvaroaleman: Or we start creating release branches and run the long test suite only when PRing into a release branch..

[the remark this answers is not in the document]
p0lyn0mial: Yeah, sounds good to me. I'd prefer to schedule a night test suite - this assures that the tests are executed frequently, at least at the beginning. The code base of the machine-controller doesn't grow very fast - a few commits during a day.

test/e2e/provisioning/aws_e2e_test.go
p0lyn0mial: ok, I'll create `scenarios` struct and place it it `helpers.go` file.

test/e2e/provisioning/aws_e2e_test.go
p0lyn0mial: sgtm :)

test/e2e/provisioning/aws_e2e_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"os"
+	"strings"
+	"testing"
+)
+
+const (
+	aws_manifest = "./testdata/machine-aws.yaml"
+)
+
+var awsScenarios = []scenarios{
```
alvaroaleman: What I actually meant was not only moving the type but the testdefinitions as well, since they are equal for all cloud providers. We should then add the `[]scenearios` as arg to the `TestCloudProviderProvisioninE2"` funcs

p0lyn0mial:
@alvaroaleman @mrIncompetent comments addressed. I have also created a bunch of tasks to follow up. PTAL

@alvaroaleman do you know how much CPU, Memory the machine we are running the tests
on have ?

test/tools/integration/provision_master.sh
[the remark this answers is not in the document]
alvaroaleman: @mrIncompetent just discussed about this and we would both like to have the full suite run on each PR and not scheduled. Also I think the runtime of this can and should definitely be improved.

If I understand this correctly, we currently only parralelize the machine scenarios, but not the cloudproviders?

E.G. we only start with DO scenarios once AWS finished? I think that should be changed.

p0lyn0mial:
> If I understand this correctly, we currently only parralelize the machine scenarios, but not the cloudproviders?

right, I think I have tested this and I didn't help it was quite the opposite. Let me unblock it one more time.

alvaroaleman:
Can we still address my comment about using a common set of scenarios for all providers??

test/e2e/provisioning/helper.go
```diff
@@ -0,0 +1,74 @@
...
+type scenario struct {
+	// name holds short description of the test scenario, it is also used to create machines and nodes names
+	// so please don't put "strange" characters there
+	name                    string
+	osName                  string
+	containerRuntime        string
+	containerRuntimeVersion string
+}
+
+func runScenarios(t *testing.T, testCases []scenario, testParams string, manifestPath string, cloudProvider string) {
```
p0lyn0mial: @alvaroaleman I hope this addresses your last comment

PR #168, opened 2018-04-15
Adds marshallJSON to configVar types

### LINKED ISSUE #130: Add MarshalJSON function to all configVar* types
Add MarshalJSON function to all configVar* types.
Also check if we can refactor the logic in the Unmarshall functions.

alvaroaleman:

Adds marshallJSON to configVar types, this avoids having a huge cluttered machine definition when the spec gets written back after defaulting.

Fixes #130 

### DIFF
--- a/pkg/providerconfig/types.go
+++ b/pkg/providerconfig/types.go
@@ -67,6 +67,56 @@ type ConfigVarString struct {
 // causing a recursion
 type configVarStringWithoutUnmarshaller ConfigVarString
 
+// This is done to not have the json object cluttered with empty strings
+// This will eventually hopefully be resolved within golang itself
+// https://github.com/golang/go/issues/11939
+func (configVarString ConfigVarString) MarshalJSON() ([]byte, error) {
+	var secretKeyRefEmpty, configMapKeyRefEmpty bool
+	if configVarString.SecretKeyRef.ObjectReference.Namespace == "" &&
+		configVarString.SecretKeyRef.ObjectReference.Name == "" &&
+		configVarString.SecretKeyRef.Key == "" {
+		secretKeyRefEmpty = true
+	}
+
+	if configVarString.ConfigMapKeyRef.ObjectReference.Namespace == "" &&
+		configVarString.ConfigMapKeyRef.ObjectReference.Name == "" &&
+		configVarString.ConfigMapKeyRef.Key == "" {
+		configMapKeyRefEmpty = true
+	}
+
+	buffer := bytes.NewBufferString("{")
+	if !secretKeyRefEmpty {
+		jsonVal, err := json.Marshal(configVarString.SecretKeyRef)
+		if err != nil {
+			return nil, err
+		}
+		buffer.WriteString(fmt.Sprintf("\"secretKeyRef\":%s", string(jsonVal)))
+	}
+
+	if !configMapKeyRefEmpty {
+		var leadingComma string
+		if !secretKeyRefEmpty {
+			leadingComma = ","
+		}
+		jsonVal, err := json.Marshal(configVarString.ConfigMapKeyRef)
+		if err != nil {
+			return nil, err
+		}
+		buffer.WriteString(fmt.Sprintf("%s\"configMapKeyRef\":%s", leadingComma, jsonVal))
+	}
+
+	if configVarString.Value != "" {
+		if !secretKeyRefEmpty || !configMapKeyRefEmpty {
+			buffer.WriteString(fmt.Sprintf(",\"value\":\"%s\"", configVarString.Value))
+		} else {
+			return []byte(fmt.Sprintf(`"%s"`, configVarString.Value)), nil
+		}
+	}
+
+	buffer.WriteString("}")
+	return buffer.Bytes(), nil
+}
+
 func (configVarString *ConfigVarString) UnmarshalJSON(b []byte) error {
 	if !bytes.HasPrefix(b, []byte("{")) {
 		b = bytes.TrimPrefix(b, []byte(`"`))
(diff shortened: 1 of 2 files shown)


pkg/providerconfig/types_test.go
```diff
@@ -28,32 +30,148 @@ func TestConfigVarStringUnmarshalling(t *testing.T) {
 
 func TestConfigVarBoolUnmarshalling(t *testing.T) {
 	jsonBool := []byte("true")
-	jsonString := []byte(`"true"`)
```
alvaroaleman: I've removed the possibility to define a `configVarBool` as boolean string because IMHO there is little reason to have it and it clutters the `UnmarshalJSON` code a lot

pkg/providerconfig/types_test.go
```diff
@@ -28,32 +30,148 @@ func TestConfigVarStringUnmarshalling(t *testing.T) {
...
 	if err != nil || !reflect.DeepEqual(expectedResult, jsonMapBoolTarget) {
 		t.Fatalf("Decoding map bool into configVarBool failed! Error: '%v'", err)
 	}
-	err = json.Unmarshal(jsonMapString, &jsonMapStringTarget)
-	if err != nil || !reflect.DeepEqual(expectedResult, jsonMapStringTarget) {
-		t.Fatalf("Decoding map string bool into configVarBool failed! Error: '%v'", err)
+}
+
+func TestConfigVarStringMarshalling(t *testing.T) {
+	t.Parallel()
```
p0lyn0mial: this is a little bit unusual, because those test take fraction of second to complete :)

pkg/providerconfig/types.go
```diff
@@ -67,6 +67,56 @@ type ConfigVarString struct {
...
+		configVarString.SecretKeyRef.ObjectReference.Name == "" &&
+		configVarString.SecretKeyRef.Key == "" {
+		secretKeyRefEmpty = true
+	}
+
+	if configVarString.ConfigMapKeyRef.ObjectReference.Namespace == "" &&
+		configVarString.ConfigMapKeyRef.ObjectReference.Name == "" &&
+		configVarString.ConfigMapKeyRef.Key == "" {
+		configMapKeyRefEmpty = true
+	}
```
mrIncompetent: Maybe rather write 
```go
configMapKeyRefEmpty := configVarString.ConfigMapKeyRef.ObjectReference.Namespace == "" &&
		configVarString.ConfigMapKeyRef.ObjectReference.Name == "" &&
		configVarString.ConfigMapKeyRef.Key == ""
```

pkg/providerconfig/types.go
```diff
@@ -67,6 +67,56 @@ type ConfigVarString struct {
...
+// This is done to not have the json object cluttered with empty strings
+// This will eventually hopefully be resolved within golang itself
+// https://github.com/golang/go/issues/11939
+func (configVarString ConfigVarString) MarshalJSON() ([]byte, error) {
+	var secretKeyRefEmpty, configMapKeyRefEmpty bool
+	if configVarString.SecretKeyRef.ObjectReference.Namespace == "" &&
+		configVarString.SecretKeyRef.ObjectReference.Name == "" &&
+		configVarString.SecretKeyRef.Key == "" {
+		secretKeyRefEmpty = true
+	}
```
mrIncompetent: Maybe rather write
```go
secretKeyRefEmpty := configVarString.SecretKeyRef.ObjectReference.Namespace == "" &&
		configVarString.SecretKeyRef.ObjectReference.Name == "" &&
		configVarString.SecretKeyRef.Key == ""
```

pkg/providerconfig/types.go
```diff
@@ -67,6 +67,56 @@ type ConfigVarString struct {
...
+
+	if configVarString.ConfigMapKeyRef.ObjectReference.Namespace == "" &&
+		configVarString.ConfigMapKeyRef.ObjectReference.Name == "" &&
+		configVarString.ConfigMapKeyRef.Key == "" {
+		configMapKeyRefEmpty = true
+	}
+
+	if secretKeyRefEmpty && configMapKeyRefEmpty {
```
mrIncompetent: Maybe rename the variables to something like `isSecretKeyRef` ?
I find it a bit more intuitive having true-focused variables instead of variables which are `true` when a condition is false.
Also it might make the following code a bit more readable.

```go
if !secretKeyRefEmpty {
    // handle secret ref 
}
```
vs.
```go
if isSecretKeyRef {
    // handle secret ref 
}
```


Though is it 100% personal preference. Not a request for change

pkg/providerconfig/types.go
mrIncompetent: Applies to `ConfigVarBool` as well

pkg/providerconfig/types.go
mrIncompetent: Applies to `ConfigVarBool` as well

pkg/providerconfig/types.go
mrIncompetent: Applies to `ConfigVarBool` as well

mrIncompetent:
Just realized all my comments are personal preference. I'll remove my request for change.

p0lyn0mial:
@alvaroaleman I'd remove `t.Parallel()` and this also lgtm to me.

pkg/providerconfig/types_test.go
alvaroaleman: Well its probably not strictly required but is there any drawback in using it? If I understand the docs correctly, golang will run everything sequentually that does not have that call above, since this test can be parallelized it doesn't harm or am I wrong there?

pkg/providerconfig/types.go
alvaroaleman: Good suggestions thought, I'll change that later

pkg/providerconfig/types_test.go
p0lyn0mial: I think the time needed to create and schedule a separate `goroutine` is longer than just let it run :)

All tests in the same package are run sequentially.
Tests across packages are run concurrently.

PR #171, opened 2018-04-16
adds e2e test for hetzner

### LINKED ISSUE #160: Hetzner E2E tests
Add the following test cases to the existing E2E test suite.

```
Ubuntu + Docker 1.13
Ubuntu + Docker 17.03
Ubuntu + CRI-O 1.9
```

Clean up `.circle/config.yaml`, so that it doesn't run the test using `create-and-destroy-machine.sh` script

p0lyn0mial:

Fixes #160


test/e2e/provisioning/all_e2e_test.go
```diff
@@ -0,0 +1,86 @@
...
+var hzScenarios = []scenario{
+	scenarios[0],
+	scenarios[1],
+	scenarios[2],
+
+	{
+		name:                    "scenario 4 CentOS Docker 1.12",
+		osName:                  "centos",
+		containerRuntime:        "docker",
+		containerRuntimeVersion: "1.12",
```
p0lyn0mial: version `17.03` not supported.

.circleci/config.yml
```diff
@@ -96,9 +96,7 @@ jobs:
       - run:
           name:  Run E2E test
           command: ./test/tools/integration/run_e2e_tests.sh
-      - run:
-          name: Test create and destroy centos machine
-          command: ./test/tools/integration/create-and-destroy-machine.sh centos
+          no_output_timeout: 15m
```
p0lyn0mial: I have increased this timeout to be greater than the running time of the tests so that we can grab the logs.

test/tools/integration/run_e2e_tests.sh
```diff
@@ -12,8 +12,9 @@ set -e
 export AWS_E2E_TESTS_KEY_ID=$AWS_E2E_TESTS_KEY_ID
 export AWS_E2E_TESTS_SECRET=$AWS_E2E_TESTS_SECRET
 export DO_E2E_TESTS_TOKEN=$DO_E2E_TESTS_TOKEN
+export HZ_TOKEN=$HZ_TOKEN
```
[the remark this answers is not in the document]
p0lyn0mial: yes, good point. I like `HZ_E2E_TESTS_TOKEN`

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -0,0 +1,86 @@
...
+package provisioning
+
+import (
+	"fmt"
+	"os"
+	"testing"
+)
+
+const (
+	do_manifest  = "./testdata/machine-digitalocean.yaml"
```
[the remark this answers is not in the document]
p0lyn0mial: I'd prefer to keep them separate. 

The ones under `examples/` are for the operators that wish to use the machine-controller. Thus, most of the fields are filled out for them. They would have to only provide credentials.

The ones under `testdata/` are used by `e2e` test and are prepared in a way that suits theirs needs.

test/e2e/provisioning/all_e2e_test.go
p0lyn0mial: That particular test case tests `centos`. I'm not sure if the others (`DO`, `AWS` etc.) support this flavour. Just by looking at the source code, I'm not sure if we support `docker 1.12` for CoreOS or Ubuntu.

p0lyn0mial: here is the reference: https://github.com/kubermatic/machine-controller/issues/129

alvaroaleman: Yeah `centos` works for everything except vsphre (allthought its just untested there, so it could potentially work).

Refs:
* https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/aws/provider.go#L110
* https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/digitalocean/provider.go#L87
* https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/hetzner/provider.go#L50

test/e2e/provisioning/all_e2e_test.go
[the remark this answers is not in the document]
alvaroaleman: Well what bothers me a little about keeping them seperate is that it means our examples are completely untested and they are the first thing a new user will see, so that potentially results in a poor experience ;)

Do you think it would be feasible to use the example, considering we currently replace all fields that need to be configured test-specific anyways?

PR #172, opened 2018-04-16
Use `kubeadm join` to join nodes

### LINKED ISSUE #139: Use `kubeadm join` instead of manually maintaining kubelet config
Right now we maintain the kubelet config as part of the distro-specific templates. This has some drawbacks:

* We may miss important configuration parameters
* Whenever we change something, we have to change it at multiple places
* There is no way to have different configs based on Kubelet version

Instead it would be easier if we just used `kubeadm join` to configure the Kubelet

alvaroaleman:

Fixes #139


pkg/userdata/ubuntu/testdata/cri-o-1.9-digitalocean.golden
```diff
@@ -0,0 +1,221 @@
...
+    [Install]
+    WantedBy=multi-user.target
+- path: "/etc/sysconfig/crio-network"
+  content: |
+    CRIO_NETWORK_OPTIONS="--registry=docker.io"
+
+runcmd:
+# Required for Hetzner, because they set some arbitrary password
+# if the sshkey wasnt set via their API and require as to change
+# that password on first login, which we cant do since we dont know
```
mrIncompetent: Then we should maybe - like with digitalocean - generate a random key for each machine we use for the API call?

alvaroaleman: That is actually a really smart idea, will do it in a follow-up

pkg/userdata/ubuntu/testdata/cri-o-1.9-digitalocean.golden
```diff
@@ -0,0 +1,221 @@
...
+      curl -L --fail https://storage.googleapis.com/cni-plugins/cni-plugins-amd64-v0.6.0.tgz|tar -xvz
+    fi
+
+- path: "/usr/local/bin/download-kubelet-kubeadm-unitfile"
+  permissions: "0777"
+  content: |
+    #!/bin/bash
+    set -xeuo pipefail
+    if ! [[ -f /etc/systemd/system/kubelet.service.d/10-kubeadm.conf ]]; then
+      curl -L --fail https://raw.githubusercontent.com/kubernetes/kubernetes/v1.9.2/build/debs/10-kubeadm.conf \
```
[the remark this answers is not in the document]
alvaroaleman: There never was an apt repo, it was hardcoded so far ;)

PR #174, opened 2018-04-17
Vsphere e2e tests

p0lyn0mial:

Fixes #


test/tools/integration/run_e2e_tests.sh
```diff
@@ -13,6 +13,9 @@ export AWS_E2E_TESTS_KEY_ID=$AWS_E2E_TESTS_KEY_ID
 export AWS_E2E_TESTS_SECRET=$AWS_E2E_TESTS_SECRET
 export DO_E2E_TESTS_TOKEN=$DO_E2E_TESTS_TOKEN
 export HZ_E2E_TOKEN=$HZ_E2E_TOKEN
+export VSPHERE_E2E_PASSWORD=$VSPHERE_PASSWORD
+export VSPHERE_E2E_USERNAME=$VSPHERE_USERNAME
+export VSPHERE_E2E_ADDRESS=$VSPHERE_ADDRESS
```
p0lyn0mial: Do I have to generate the new username and password ? Or Can I use the ones you gave me ?

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -84,3 +85,27 @@ func TestHetznerProvisioningE2E(t *testing.T) {
 	params := fmt.Sprintf("<< HETZNER_TOKEN >>=%s", hzToken)
 	runScenarios(t, hzScenarios, params, hz_manifest, "hz")
 }
+
+var vsphereScenarios = []scenario{
+	scenarios[0],
+	scenarios[1],
+	scenarios[2],
+}
```
alvaroaleman: Rather than having to specify individual scenarios at the cloud provider level as a index reference, do you think it would be possible to alter `runScenarios` in a way that it allows to specify a distribution list as argument and will skip all tests that have an `osName` that is in that list?

That way if we ever adjust our global test list, e.G. because of a new docker version, we do not have to go throught the individual providers.

Also I think that way it is more understandable as to why specific tests are not run withing a provider (it does not support that distribution). Right now that is really hard to see by just looking at the tests

p0lyn0mial: What if you could access individual scenarios by name ? something like this"
```
...
scenarios["scenario 1 Ubuntu Docker 1.13"],
scenarios["scenario 2 Ubuntu Docker 17.03"]
...
```

alvaroaleman: That makes it more readable, but it still means the provider has to opt in for tests rather than to opt out, also it does not tell one why this scenarios but not others. Why need some kind of selector for that that makes it obvious, e.G. the above mentioned `skip if distro == "ubuntu"`

The assumption and goal is that all providers implement all distros. We have right now one case where thats not true (vsphere) but that is something that has to be fixed very soon.

p0lyn0mial: If all test can use the global list, that's fine. If not they can pick and choose or define their own list, what's wrong with that ? I think that having duplicated list is much more readable than having a separate list that will tell you which tests you shoudn't run ?

PR #177, opened 2018-04-18
Run e2e locally

alvaroaleman:

This makes the e2e tests work from localhost without having to copy binaries over

Fixes #


test/e2e/provisioning/helper.go
```diff
@@ -127,28 +126,27 @@ func runScenarios(t *testing.T, excludeSelector *scenarioSelector, testParams st
 }
 
 func runVerifyTool(manifestPath string, params string) error {
-	homeDir := os.Getenv("HOME")
-	kubeConfigDir := filepath.Join(homeDir, ".kube/config")
+	gopath := os.Getenv("GOPATH")
```
p0lyn0mial: nit: usually home dir is the default path for `kubeconfig`.

test/e2e/provisioning/helper.go
```diff
@@ -127,28 +126,27 @@ func runScenarios(t *testing.T, excludeSelector *scenarioSelector, testParams st
...
 		"true",
 	}
 
-	cmd := exec.Command("../../tools/verify/verify", args...)
-	var out bytes.Buffer
-	cmd.Stderr = &out
-
-	err := cmd.Run()
+	cmd := exec.Command(filepath.Join(projectDir, "test/tools/verify/verify"), args...)
+	out, err := cmd.CombinedOutput()
```
p0lyn0mial: this is elegant :)

test/tools/integration/provision_master.sh
```diff
@@ -17,7 +17,7 @@ done
 
 
 rsync -av  -e "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" \
-    ../../../{examples,machine-controller,Dockerfile,test} ../verify/verify \
+    ../../../{examples,machine-controller,Dockerfile,test/tools/verify/verify} \
```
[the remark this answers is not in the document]
p0lyn0mial: also do we have to copy `examples` directory ?

test/e2e/provisioning/helper.go
alvaroaleman: Yeah that is true, but I wanted to allow to run it locally and do not want it to overwrite my kubeconfig

p0lyn0mial: how can I run the tests from my local machine ?

alvaroaleman: If you set all env variables, run the `provision_master.sh` once manually you can just execute `go test -tags=e2e -parallel 24 -v -timeout 10m  ./test/e2e/...`

I could probably polish it a little bit more and write a maketarget that does both

PR #185, opened 2018-04-24
add os dependend root device path for aws

mrIncompetent:
Every image comes with a default root device.
As we dynamically set a device, depending on the user settings, we need to replace the default root device.
Otherwise we end up with a 8gb default device mounted at `/`  and a 50gb SSD unused.

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -3,7 +3,6 @@ package aws
 import (
 	"encoding/base64"
 	"encoding/json"
-	"errors"
 	"fmt"
 
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider/cloud"
@@ -190,7 +189,7 @@ func getDefaultRootDevicePath(os providerconfig.OperatingSystem) (string, error)
 		return "/dev/xvda", nil
 	}
 
-	return "", errors.New("no default root path found")
+	return "", fmt.Errorf("no default root path found for %s operating system", os)
 }
 
 func (p *provider) getConfig(s runtime.RawExtension) (*Config, *providerconfig.Config, error) {
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -3,6 +3,7 @@ package aws
 import (
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider/cloud"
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -530,8 +544,11 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 			return nil, err
 		}
 		securityGroupIDs = append(securityGroupIDs, sgID)
-	} else {
+	}
 
+	rootDevicePath, err := getDefaultRootDevicePath(pc.OperatingSystem)
```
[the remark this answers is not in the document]
p0lyn0mial: oh, I see. It aligns with the list of supported operating systems.

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -179,6 +180,19 @@ func getDefaultAMIID(os providerconfig.OperatingSystem, region string) (string,
...
+	switch os {
+	case providerconfig.OperatingSystemUbuntu:
+		return "/dev/sda1", nil
+	case providerconfig.OperatingSystemCentOS:
+		return "/dev/sda1", nil
+	case providerconfig.OperatingSystemCoreos:
+		return "/dev/xvda", nil
+	}
+
+	return "", errors.New("no default root path found")
```
p0lyn0mial: nit: "no default root path found for %s operating system."

test/e2e/provisioning/testdata/machine-aws.yaml
```diff
@@ -13,9 +13,9 @@ spec:
       region: "eu-west-2"
       availabilityZone: "eu-west-2a"
       vpcId: "vpc-40827629"
-      instanceType: "t2.micro"
+      instanceType: "t2.medium"
```
p0lyn0mial: may i know why we want to use bigger machines ?

mrIncompetent: Too small machines are one of the reasons we run into timeouts during the e2e tests

p0lyn0mial: that's interesting, bigger machines are provisioned faster :)

p0lyn0mial: do you want to change the value for the others ? (`vsphere` etc.)

mrIncompetent: I'm not 100% convinced yet if it helps but definitly worth a try.

alvaroaleman: Well I don't think small machines were the reason why we ran into timeouts because most of the time the e2e tests passed, just sometimes they would fail because of outliers.

But bigger machines definitely improve the speed, allthought I assume there is some threshold at which it doesn't make any difference anymore (e.G. at vsphere 2 cpu was much faster than one, but I doubt if 4cpus would make it faster)

PR #186, opened 2018-04-24
treats InstanceLimitExceeded as terminal error

### LINKED ISSUE #183: Quota error from AWS should cause terminal error
Replicate:
Create 21 t2.small instances. (20 is the default quota)

p0lyn0mial:
**What this PR does / why we need it**: this pull treats `InstanceLimitExceeded` as terminal error. We are limited to running up to 20 on demand. See for more details https://aws.amazon.com/ec2/faqs/#How_many_instances_can_I_run_in_Amazon_EC2

Fixes #183 


p0lyn0mial:
@alvaroaleman @mrIncompetent here is the list of the errors that might be returned by `AWS`

https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html

I have just thought we could be proactive and classify some of the errors on the list as terminal but perhaps it will be easier to treat them as they will appear.

p0lyn0mial:
does it mean that we cannot run more than 20 instances of certain type ?

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -752,6 +752,11 @@ func awsErrorToTerminalError(err error, msg string) error {
 			return prepareAndReturnError()
 		}
 		switch aerr.Code() {
+		case "InstanceLimitExceeded":
```
alvaroaleman: I am not sure if that makes sense, because this error doesn't get fixed by altering the current's machine spec but instead by deleting instances that may or may not be managed by the machine-controller.

IMHO from the POV of the machine-controller this is a transient error, which we probably should make available  as an event but I wouldn't consider it a terminal error

p0lyn0mial: the error is an indicator to the end user that we cannot create more machines unless (a manual interaction):

a) some machines will be removed
b) the limit will be increased  (i'm not sure if this is possible)

I'm calling out to @mrIncompetent, as he created the linked issue.

mrIncompetent: Terminal errors are defined as errors which can only be solved by user interaction.
IMO that is the case here.

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -752,6 +752,11 @@ func awsErrorToTerminalError(err error, msg string) error {
...
+			return cloudprovidererrors.TerminalError{
+				Reason:  v1alpha1.InsufficientResourcesMachineError,
+				Message: "You've reached the limit on the number of instances you can run concurrently. The limit depends on the instance type",
```
mrIncompetent: Rather
`You've reached the AWS quota for number of instances of this type`

We should make it explicit that its about the AWS instance quote, as the termin quota occurs in k8s as well. Just to avoid confusion

pkg/cloudprovider/provider/aws/provider.go
alvaroaleman: But this error is a condition of your fleet, not of an individual instance. The machine-controller stops acting on machines that have a terminal errors set, right? This does not make sense in this case, as the error can be resolved without changing anything on the machines spec.

mrIncompetent: I don't think we stop processing when a terminal error is set.
We definitly should not.

PR #193, opened 2018-04-25
Generate temporary ssh key for Hetzner

### LINKED ISSUE #191: Create temporary SSH key for Hetzner
If a server at Hetzner is created without a SSH key configured, Hetzner will set a random password, send an email to the project owner and set the password to be expired on next login.

We could avoid this hassle by creating a temporary ssh key just like we do for digital ocean.

alvaroaleman:

Fixes #191


pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -176,6 +182,34 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
...
+	if err := tmpRSAKeyPair.Validate(); err != nil {
+		return nil, fmt.Errorf("failed to validate private RSA key: %v", err)
+	}
+	pubKey, err := ssh.NewPublicKey(&tmpRSAKeyPair.PublicKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate ssh public key: %v", err)
+	}
+
+	hkey, res, err := client.SSHKey.Create(ctx, hcloud.SSHKeyCreateOpts{
+		Name:      fmt.Sprintf("machine-controller-temp-%s", machine.Spec.Name),
```
p0lyn0mial: not sure if this is enough, I can give the same name to a machine in my cluster.

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -176,6 +182,34 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
...
+		return nil, fmt.Errorf("failed to create private RSA key: %v", err)
+	}
+
+	if err := tmpRSAKeyPair.Validate(); err != nil {
+		return nil, fmt.Errorf("failed to validate private RSA key: %v", err)
+	}
+	pubKey, err := ssh.NewPublicKey(&tmpRSAKeyPair.PublicKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate ssh public key: %v", err)
+	}
```
mrIncompetent: We have the same logic now twice, fir Hetzner and Digitalocean.
Better extract it into a separate package, which returns a ssh.PublicKey & a random name.

.circleci/config.yml
```diff
@@ -62,7 +62,11 @@ jobs:
           command: ./test/tools/integration/provision_master.sh
       - run:
           name:  Run E2E test
-          command: go test -tags=e2e -parallel 24 -v -timeout 20m  ./test/e2e/... -identifier=$CIRCLE_BUILD_NUM
+          command: >-
```
p0lyn0mial: do you think we could export the content of this command to a file? So that I can run the test from my local machine the same way we they are run in `CI` ? 

I'm not saying we have to do this in this pull, just an idea...

alvaroaleman: Well, the provisioning script copy the kubeconfig over to `$GOPATH/src/github.com/kubermatic/machine-controller/.kubeconfig` so  you can run the `go test -tags=e2e -parallel 24 -v -timeout 20m  ./test/e2e/...` fine locally.

The `export KUBECONFIG=...` is just to make configure `kubectl` to use the correct kubeconfig for logs in case of failure. I am a bit hesitant to write a script that changes the users environment because you are suddenly talking to a diffrent k8s cluster

p0lyn0mial: I think that you cannot change parent's shell variables from a child process. So `KUBECONFIG` will not be overwritten.

PR #202, opened 2018-05-07
fix npe during openstack machine-creation

mrIncompetent:
When no tags are specified we hit a nil-pointer-exception


pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -406,6 +406,9 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 
 	// we check against reserved tags in Validation method
 	allTags := c.Tags
+	if allTags == nil {
```
p0lyn0mial: do you think we could add this check in line 162 inside `getConfig` method ?
so that whenever you call `getConfig` it will be properly initialized.

PR #213, opened 2018-05-14
create a bootstrap token only if doesn't exist

### LINKED ISSUE #210: Endless creation of bootstrap tokens when terminal error occurs
Bootstrap token should be reused. Therefore we cannot use `generateName`.
But when reusing tokens check the lifetime. Or simply always update the lifetime to now+24hours.

p0lyn0mial:
**What this PR does / why we need it**: creates a bootstrap token for a machine only if it doesn't exist.

Fixes #210 


pkg/controller/kubeconfig.go
```diff
@@ -6,60 +6,111 @@ import (
...
+	}
+
 	tokenID := rand.String(6)
 	tokenSecret := rand.String(16)
 
 	secret := v1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name: fmt.Sprintf("bootstrap-token-%s", tokenID),
+			Name:   fmt.Sprintf("bootstrap-token-%s", tokenID),
+			Labels: map[string]string{machineNameLabelKey: name},
```
mrIncompetent: Why matching via labels when you could use a static name for the secret?

p0lyn0mial: this will work only for new `Secrets`. In the worst case scenario we will create additional `Secret` next to existing.

p0lyn0mial: fair enough, do you think we could use the name of the machine ?

pkg/controller/kubeconfig.go
```diff
@@ -5,61 +5,100 @@ import (
...
+
 	}
+	tokenID := updatedSecret.StringData[tokenIDKey]
+	tokenSecret := updatedSecret.StringData[tokenSecretKey]
+	return fmt.Sprintf("%s.s", tokenID, tokenSecret), nil
+}
 
-	infoKubeconfig, err := c.kubeconfigProvider.GetKubeconfig()
+func (c *Controller) getSecretIfExists(name string) (*v1.Secret, error) {
+	secret, err := c.secretSystemNsLister.Secrets(metav1.NamespaceSystem).Get(fmt.Sprintf(tokenNameFormatter, name))
```
p0lyn0mial: using x can be a tad premature, but why not ?

pkg/controller/kubeconfig.go
```diff
@@ -5,61 +5,101 @@ import (
...
+	return fmt.Sprintf(tokenFormatter, tokenID, tokenSecret), nil
 }
 
-func (c *Controller) createBootstrapKubeconfig(name string) (*clientcmdapi.Config, error) {
-	token, err := c.createBootstrapToken(name)
+func (c *Controller) updateSecretExpirationAndGetToken(secret *v1.Secret) (string, error) {
+	if secret.StringData == nil {
+		secret.StringData = map[string]string{}
+	}
+	secret.StringData["expiration"] = metav1.Now().Add(24 * time.Hour).Format(time.RFC3339)
```
mrIncompetent: Maybe we should check if the token will expire within the next minutes and only then update it?
WDYT @p0lyn0mial 
It could help to reduce the load

p0lyn0mial:
the e2e test fail due to `User "system:serviceaccount:kube-system:machine-controller" cannot get secrets in the namespace "kube-system"
` 

I was convinced that `machine-controller` use admin account and can do everything.

pkg/controller/kubeconfig.go
p0lyn0mial: happy path doesn't have burden on performance it is one-off operation. 

I would leave it as it is, mainly because I don't want to deal with `Data["expiration"]` which is defined as `[]bytes`. I'd prefer to treat it as opaque value.

Unless we think this is important I can try to implement that.

mrIncompetent: I was thinking about the PUT request a few lines down.

pkg/controller/kubeconfig.go
```diff
@@ -6,60 +6,129 @@ import (
...
+	tokenSecret := secret.StringData[tokenSecretKey]
+	token := fmt.Sprintf(tokenFormatter, tokenID, tokenSecret)
+
+	expBytes, ok := secret.Data["expiration"]
+	if !ok {
+		return "", fmt.Errorf("haven't found %s key in the secret's Data field", expirationKey)
+	}
+	expString := string(expBytes)
+	expVal := metav1.Now()
+	err := expVal.UnmarshalQueryParameter(expString)
```
p0lyn0mial: not particularly elegant but it works :)

examples/machine-controller.yaml
```diff
@@ -83,10 +83,17 @@ rules:
...
-  - secrets
   - events
   verbs:
   - create
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
```
alvaroaleman: I think we need update here as well

pkg/controller/kubeconfig.go
alvaroaleman: Is there a reason why we can't use the same token for all machines?

p0lyn0mial:
> Is there a reason why we can't use the same token for all machines?

what would be lifetime of such a token ?

pkg/controller/kubeconfig.go
```diff
@@ -6,60 +6,129 @@ import (
...
+		return "", err
+	}
+	now := metav1.Now()
+	// expVal has to point to a time in the future otherwise we need to update expiration time
+	if now.Before(&expVal) {
+		return token, nil
 	}
 
-	infoKubeconfig, err := c.kubeconfigProvider.GetKubeconfig()
+	_, err = c.kubeClient.CoreV1().Secrets(metav1.NamespaceSystem).Update(secret)
```
p0lyn0mial: haven't we made this token immortal :( ?

pkg/controller/kubeconfig.go
```diff
@@ -6,60 +6,129 @@ import (
...
+	return fmt.Sprintf(tokenFormatter, tokenID, tokenSecret), nil
 }
 
-func (c *Controller) createBootstrapKubeconfig(name string) (*clientcmdapi.Config, error) {
-	token, err := c.createBootstrapToken(name)
+func (c *Controller) updateSecretExpirationAndGetToken(secret *v1.Secret) (string, error) {
+	if secret.StringData == nil {
+		secret.StringData = map[string]string{}
+	}
+	secret.StringData[expirationKey] = metav1.Now().Add(24 * time.Hour).Format(time.RFC3339)
```
p0lyn0mial: Can we lower the expiration time to `1 hour ?`

p0lyn0mial:
Are the expired tokens removed (`TokenCleaner) ?

pkg/controller/kubeconfig.go
p0lyn0mial: what would be lifetime of such a token ?

alvaroaleman:
> what would be lifetime of such a token ?

The life time? Probably infinity, we could add an `ownerRef` pointing to the machine-controllers service account, but that may cause issues if ppl don't use rbac.

The reason why I even came up with this idea is that I realized we currently do not have any logic to delete tokens, thus they will stay forever

The validity time would be just like now, Increase it by some time everytime we want to use .t

pkg/controller/kubeconfig.go
alvaroaleman: Yep, makes sense. I think the 24h originate from `kubeadm`, but there the use case is a user using the token to manually join one or many nodes which is obviously different from what we do

p0lyn0mial:
> The reason why I even came up with this idea is that I realized we currently do not have any logic to delete tokens, thus they will stay forever

Tokens are sensitive in a sense that they can be use to authorise, obtain client cert and other things.

Wouldn't be better to decrease the expiration time and make sure that `TokenCleaner` plugin is enabled, so that expired tokens are deleted ?

mrIncompetent:
Bootstrap-tokens are for a one-time use. They are only being used to create a CSR. After a certificate has been fetched, the token does not get used anymore.
That's why they get called `bootstrap-tokens`. Therefore they should expire and each node should get a individual one.

The lifetime should probably reduced to 1 hour.

alvaroaleman:
Ah okay, makes perfect sense. I didn't realize there is a `TokenCleaner` controller and thought we'd have to clean them up ourselves

mrIncompetent:
> Are the expired tokens removed (`TokenCleaner) ?

When using kubeadm or kubermatic yes. Kubeadm enables all relevant token controllers.
Which is something we should document.

pkg/controller/kubeconfig.go
p0lyn0mial: I would remove "extend token expiration" logic. If the token is no longer valid it will be `GCed` and we will crate a new valid token for a machine. WDYT ?

p0lyn0mial: makes sense, then we should add some duration to `now`, so that there is enough time for a machine to join the cluster (`now.Before(&expVal`)

examples/machine-controller.yaml
p0lyn0mial: ups :)

pkg/controller/kubeconfig.go
mrIncompetent: Yes. We need some definition. 30min? Wdyt?

p0lyn0mial: I set it to 15 min :)

p0lyn0mial:
i'm putting this PR onto hold

1. `e2e` test fail due to `CentOS and `docker 1.12` - this issue has been fixed by @kdomanski (awaiting merge)
2. vsphere is unreachable - waiting for @scheeles to fix the port forwarding

p0lyn0mial:
whooha tests are green :) - finally!

@mrIncompetent thank you :)

PR #214, opened 2018-05-14
fix cloud-config and provider setting on CoreOS

### LINKED ISSUE #203: Set cloud-provider flag on vsphere + write cloud-config

kdomanski:
Fixes #203


mrIncompetent:
Do the flags on the kubelet already get set?

kdomanski:
Yes, the `kubelet.service` template contains the `--cloud-provider` and `--cloud-config` flags. It's just their values that never made it to the template.

PR #215, opened 2018-05-14
userdata/centos: drop docker 1.12 which is no longer available for install

kdomanski:


kdomanski:
It has actually been pulled from 7.4 as well.

p0lyn0mial:
found it, but wasn't easy.

running `more /var/log/cloud-init.log  | grep yum` gave me

`Running command ['yum', '-t', '-y', 'install', 'docker-1.12.6', 'kubelet-1.9.6', 'kubeadm-1.9.6', 'ebtables', 'ethtool', 'nfs-utils', 'bash-completion', 'sudo'] with allowed return codes [0] (shell=False, capture=False)`

then running `more /var/log/yum.log` gave me
```
May 14 12:34:49 Installed: ebtables-2.0.10-16.el7.x86_64
May 14 12:34:49 Installed: socat-1.7.3.2-2.el7.x86_64
May 14 12:34:50 Installed: kubectl-1.10.2-0.x86_64
May 14 12:34:50 Updated: 2:ethtool-4.8-7.el7.x86_64
May 14 12:34:52 Installed: kubernetes-cni-0.6.0-0.x86_64
May 14 12:34:55 Installed: kubelet-1.9.6-0.x86_64
May 14 12:35:02 Installed: kubeadm-1.9.6-0.x86_64
May 14 12:35:03 Updated: 1:nfs-utils-1.3.0-0.54.el7.x86_64
May 14 12:35:04 Updated: sudo-1.8.19p2-13.el7.x86_64
May 14 12:35:04 Installed: 1:bash-completion-2.1-6.el7.noarch
```

 How about removing the `-t` flag from the pkg manager, so that the errors are be more visible ?

p0lyn0mial:
I was hoping that there is a way to tell the `cloud-init` how to be a man :)

alvaroaleman:
I'd argue not talking about whats going on in the inside is a pretty manly move lol

PR #216, opened 2018-05-17
image builder

### LINKED ISSUE #212: [Documentation] Create offline os images
The documentation needs to cover the following topics
- Installation of required tools
- Creation of RHEL/Centos images
- Creation of Ubuntu images
- Creation of CoreOS images
- Usage of those images in vSphere

kdomanski:
Fixes #212


alvaroaleman:
Well first off, we want to be able to do this for all clouds, so why start off with something we are going to change anyways later on?

image-builder/build.sh
```diff
@@ -0,0 +1,187 @@
...
+  EXPECTED_SHA256="$(grep "CentOS-7-x86_64-GenericCloud-$CENTOS7_BUILD.qcow2.xz$" < "$TEMPDIR/centos7-sha256sum.txt" | cut -f1 -d ' ')"
+  CALCULATED_SHA256="$(sha256sum "$TEMPDIR/CentOS-7-x86_64-GenericCloud-$CENTOS7_BUILD.qcow2.xz" | cut -f1 -d ' ')"
+  if [[ "$CALCULATED_SHA256" != "$EXPECTED_SHA256" ]]; then
+    echo " * SHA256 digest verification failed. '$CALCULATED_SHA256' != '$EXPECTED_SHA256'"
+    exit 1
+  fi
+
+  echo " * Decompressing"
+  unxz --keep "$TEMPDIR/CentOS-7-x86_64-GenericCloud-$CENTOS7_BUILD.qcow2.xz"
+  mv "$TEMPDIR/CentOS-7-x86_64-GenericCloud-$CENTOS7_BUILD.qcow2" "$SCRIPT_DIR/downloads/CentOS-7-x86_64-GenericCloud.qcow2"
```
alvaroaleman: This will not work for vsphere because it can not handle qcow2 images

image-builder/build.sh
```diff
@@ -0,0 +1,187 @@
...
+./download_kubernetes.sh --release "$K8S_RELEASE"
+
+echo " * Mouting the image"
+cp "$CLEAN_IMAGE" "$TEMPDIR/work-in-progress-image"
+sudo guestmount -a "$TEMPDIR/work-in-progress-image" -m "$ROOTFS_PARTITION" "$TARGETFS"
+
+echo " * Copying kubernetes binaries"
+sudo mkdir -p "$TARGETFS/opt/bin/"
+sudo cp "$SCRIPT_DIR/downloads/kubeadm-$K8S_RELEASE" "$TARGETFS/opt/bin/kubeadm"
+sudo cp "$SCRIPT_DIR/downloads/kubectl-$K8S_RELEASE" "$TARGETFS/opt/bin/kubectl"
```
alvaroaleman: For CoreOS we use the hyperkube image instead of kubelet/kubeadm:

https://github.com/kubermatic/machine-controller/blob/master/pkg/userdata/coreos/userdata.go#L170-L222

image-builder/build.sh
alvaroaleman: Oh indeed, my bad

mrIncompetent:
My take on it:
The feature to run in an offline environment is most important for vSphere setups.
Therefore i think the approach is fine how it is at the moment. 

Though i really would like to see a comparison to packer here. 
Packer seems to be the standard solution for pre-baking images.

PR #217, opened 2018-05-18
add missing logout in vsphere client

mrIncompetent:
add missing logout in vSphere client to to prevent killing the vSphere.


pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -176,6 +176,7 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 	if err != nil {
 		return fmt.Errorf("failed to get vsphere client: '%v'", err)
 	}
+	defer client.Logout(context.TODO())
```
p0lyn0mial: how about returning additional value from `getClient` method e.g.
```
client, disposeClient, err := getClient...
defer disposeClient
```

PR #218, opened 2018-05-23
Add alerts.yaml to examples

metalmatze:
These are the alerts we're using for Kubermatic.


p0lyn0mial:
`e2e` test will fail due to issues with `vsphere`
as the changes don't change the controller logic feel free to merge.

PR #222, opened 2018-05-24
Make vsphere cluster configurable

mrIncompetent:
Fixes our e2e tests

### DIFF
--- a/examples/machine-vsphere.yaml
+++ b/examples/machine-vsphere.yaml
@@ -25,6 +25,7 @@ spec:
     cloudProviderSpec:
       templateVMName: ubuntu-template
       username: '<< VSPHERE_USERNAME >>'
+      # example: 'https://your-vcenter:8443'. '/sdk' gets appended automatically
       vsphereURL: '<< VSPHERE_ADDRESS >>'
       datacenter: 'Datacenter'
       password:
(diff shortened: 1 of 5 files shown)


examples/machine-vsphere.yaml
```diff
@@ -25,7 +25,7 @@ spec:
     cloudProviderSpec:
       templateVMName: ubuntu-template
       username: '<< VSPHERE_USERNAME >>'
-      vsphereURL: 'https://<< VSPHERE_ADDRESS >>'
+      vsphereURL: '<< VSPHERE_ADDRESS >>'
```
p0lyn0mial: we could add a comment that would show an example value.

mrIncompetent: I'll add an example

test/tools/integration/README.md
```diff
@@ -14,7 +14,7 @@ creates a single-node-cluster (servertype: `cx11`) via `kubeadm` at
 
 ## Usage
 
-* `export HZ_TOKEN=<my_hetzner_cloud_token>`
+* `export HZ_E2E_TOKEN=<my_hetzner_cloud_token>`
```
p0lyn0mial: nice :)

p0lyn0mial:
one nit and this is lgtm
@mrIncompetent thank you for fixing our e2e tests on vsphere !

PR #223, opened 2018-05-24
fix setting of coreos userdata on vSphere

mrIncompetent:
Finally fixed the e2e tests.
Was picked from https://github.com/kubermatic/machine-controller/pull/214


pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -106,21 +106,48 @@ func CreateLinkClonedVm(vmName, vmImage, datacenter, clusterName, folder string,
...
-			{
-				Info:            &types.VAppPropertyInfo{Key: 7, Id: "guestinfo.coreos.config.data.encoding", Value: "base64"},
-				ArrayUpdateSpec: types.ArrayUpdateSpec{Operation: types.ArrayUpdateOperationEdit},
-			},
-			{
-				Info:            &types.VAppPropertyInfo{Key: 4, Id: "guestinfo.coreos.config.data", Value: userdataBase64},
-				ArrayUpdateSpec: types.ArrayUpdateSpec{Operation: types.ArrayUpdateOperationEdit},
-			},
-		}}
+		// The properties describing userdata will already exist in the CoreOS VM template.
```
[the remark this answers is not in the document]
mrIncompetent: kudos to @kdomanski

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -106,21 +106,48 @@ func CreateLinkClonedVm(vmName, vmImage, datacenter, clusterName, folder string,
...
+		// In order to overwrite them, we need to specify their numeric Key values,
+		// which we'll extract from that template.
+		var mvm mo.VirtualMachine
+		err = templateVm.Properties(ctx, templateVm.Reference(), []string{"config", "config.vAppConfig", "config.vAppConfig.property"}, &mvm)
+		if err != nil {
+			return err
+		}
+
+		var property []types.VAppPropertySpec
+		info := mvm.Config.VAppConfig.GetVmConfigInfo()
```
p0lyn0mial: doesn't work in golang ?

`properties := mvm.Config.VAppConfig.GetVmConfigInfo().Property`

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -106,21 +106,48 @@ func CreateLinkClonedVm(vmName, vmImage, datacenter, clusterName, folder string,
...
+		properties := info.Property
+		for _, item := range properties {
+			switch item.Id {
+			case "guestinfo.coreos.config.data":
+				property = append(property, types.VAppPropertySpec{
+					ArrayUpdateSpec: types.ArrayUpdateSpec{
```
p0lyn0mial: just wanted to see if this doesn't have to be a pointer. the source code looks really ugly.

```
type ArrayUpdateSpec struct {

   DynamicData
    
   Operation ArrayUpdateOperation `xml:"operation"`
   RemoveKey AnyType              `xml:"removeKey,omitempty,typeattr"`
   }
    
   func init() {
   t["ArrayUpdateSpec"] = reflect.TypeOf((*ArrayUpdateSpec)(nil)).Elem()
   }
```

pkg/cloudprovider/provider/vsphere/helper.go
mrIncompetent: It does but i encountered at some point a nil-pointer exception. I was not able to reproduce it but splitting the one line makes it easier to debug later on

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -106,21 +106,48 @@ func CreateLinkClonedVm(vmName, vmImage, datacenter, clusterName, folder string,
...
-			{
-				Info:            &types.VAppPropertyInfo{Key: 4, Id: "guestinfo.coreos.config.data", Value: userdataBase64},
-				ArrayUpdateSpec: types.ArrayUpdateSpec{Operation: types.ArrayUpdateOperationEdit},
-			},
-		}}
+		// The properties describing userdata will already exist in the CoreOS VM template.
+		// In order to overwrite them, we need to specify their numeric Key values,
+		// which we'll extract from that template.
+		var mvm mo.VirtualMachine
+		err = templateVm.Properties(ctx, templateVm.Reference(), []string{"config", "config.vAppConfig", "config.vAppConfig.property"}, &mvm)
```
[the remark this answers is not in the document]
mrIncompetent: I'm not entirely sure if they'll always exist. They definitely do on the official coreos-template.

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -106,21 +106,48 @@ func CreateLinkClonedVm(vmName, vmImage, datacenter, clusterName, folder string,
...
+						Key:   item.Key,
+						Id:    item.Id,
+						Value: "base64",
+					},
+				})
+				break
+			}
+		}
+
+		vAppAconfig = &types.VmConfigSpec{Property: property}
```
p0lyn0mial: oh, we are taking only the first encountered value.

p0lyn0mial:
Do we have to export `CreateLinkClonedVm` method ?

pkg/cloudprovider/provider/vsphere/helper.go
p0lyn0mial: oh, should we then add the following check ?
`if info != nil {...}`

p0lyn0mial:
it seems like the API returns only pointers, checking every returned value against nil seems counter productive. It seems like the authors of the `govoc` don't evaluate the pointers 

https://github.com/vmware/govmomi/blob/master/govc/vapp/destroy.go

kdomanski:
I could have just rebased the old PR against master.

kdomanski:
Nah, just pointing out for the future.

pkg/cloudprovider/provider/vsphere/helper.go
mrIncompetent: Was actually just one thing i needed to check.
Now we return an meaningful error

pkg/cloudprovider/provider/vsphere/helper.go
mrIncompetent: Yeah, the vmware sdk is quite special...

pkg/cloudprovider/provider/vsphere/helper.go
mrIncompetent: `property` was actually a slice. But as the naming of the vmware type is fairly bad, it's super confusing.
I renamed the var to make it more clear

mrIncompetent:
> Do we have to export CreateLinkClonedVm method ?

Nope, changed

mrIncompetent:
PTAL @p0lyn0mial

PR #228, opened 2018-05-28
leader election: increase the log level of successfully renewer lease

kdomanski:
The current log level (4) floods over things that are actually of note during debugging.

Setting log level to 4 uncovers some important information that gets drowned in renewed lease messages coming every second. Although it is something worth debug-logging, it isn't even a significant state change - thus I'd like to move it to the higher log level for improved clarity.

### DIFF
--- a/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go
+++ b/vendor/k8s.io/client-go/tools/leaderelection/leaderelection.go
@@ -196,7 +196,7 @@ func (le *LeaderElector) renew() {
 		le.maybeReportTransition()
 		desc := le.config.Lock.Describe()
 		if err == nil {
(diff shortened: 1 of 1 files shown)


p0lyn0mial:
We use an external library `client-go/leaderelection` provided by k8s for leader election.  

This pull changes vendored code so it will not survive updates :). If you want to change the log level, there is no other way than opening a pull request upstream :)

mrIncompetent:
I would love to get rid of those messages but changing code in vendor is not an option.
You would need to create a PR upstream:https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go

kdomanski:
Riiiiight, missed that.

Issue #211, opened 2018-05-11
Add gometaliner to CI pipeline


alvaroaleman:
To avoid oversights like e.G. unhandled `err` returns, we should add gometalinter to the CircleCI pipeline.

kdomanski:
done by #1669

PR #230, opened 2018-05-30
add Azure support

### LINKED ISSUE #225: Implement Azure cloud provider

### LINKED ISSUE #226: Add e2e tests for Azure
Blocked by #225

kdomanski:
Fixes #225
Fixes #226


pkg/cloudprovider/provider/azure/create_delete_resources.go
```diff
@@ -0,0 +1,282 @@
...
+	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-04-01/network"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/golang/glog"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// deleteInterfacesByMachineUID will remove all network interfaces tagged with the specific machine's UID.
+// The machine has to be deleted or disassociated with the interfaces beforehand, since Azure won't allow
+// us to remove interfaces connected to a VM.
+func deleteInterfacesByMachineUID(ctx context.Context, c *config, machineUID types.UID) error {
```
alvaroaleman: Can you quickly explain why we need this? The terraform azure provider does not do it: https://github.com/terraform-providers/terraform-provider-azurerm/blob/aab882f476253e05bae288062c42421d7bf77746/azurerm/resource_arm_virtual_machine.go#L747

pkg/cloudprovider/provider/azure/create_delete_resources.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		return err
+	}
+
+	list, err := vmClient.ListAll(ctx)
+	if err != nil {
+		return err
+	}
+
+	allServers := make([]compute.VirtualMachine, 0)
```
alvaroaleman: I think its more idiomatic to write `var allServers []compute.VirtualMachine` instead of instantiating a zero-lenght array

pkg/cloudprovider/provider/azure/create_delete_resources.go
```diff
@@ -0,0 +1,282 @@
...
+	if err != nil {
+		return err
+	}
+
+	list, err := disksClient.List(ctx)
+	if err != nil {
+		return err
+	}
+
+	allDisks := make([]compute.Disk, 0)
```
alvaroaleman: I think its more idiomatic to write var allServers []compute.Disk instead of instantiating a zero-lenght array

pkg/cloudprovider/provider/azure/create_delete_resources.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	subnet, err := getSubnet(ctx, config)
+	if err != nil {
+		return network.Interface{}, fmt.Errorf("failed to fetch subnet: %v", err)
+	}
+
+	glog.Infof("Creating public IP for interface %q", ifName)
+	ipName := ifName + "-pubip"
+	_, err = createPublicIPAddress(ctx, ipName, machineUID, config)
```
alvaroaleman: Do instances need a publicIP to be able to communicate with the Internet? I am wondering, because in e.G. openstack they do not, thus its optional to assign a public IP to the nodes. If you e.G. have your whole cluster inside Azure, there is most likely no reason for the instances to have a publicIP

pkg/cloudprovider/provider/azure/get_client.go
```diff
@@ -0,0 +1,64 @@
...
+
+	return ipClient, nil
+}
+
+func getSubnetsClient(c *config) (network.SubnetsClient, error) {
+	var err error
+	ipClient := network.NewSubnetsClient(c.SubscriptionID)
+	ipClient.Authorizer, err = auth.NewClientCredentialsConfig(c.ClientID, c.ClientSecret, c.TenantID).Authorizer()
+	if err != nil {
+		return network.SubnetsClient{}, fmt.Errorf("failed to create authorizer: %s", err.Error())
```
alvaroaleman: I think its easier if you let the func return a pointer, because then you can do `return nil, fmt.Errorf("my error")`

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,581 @@
...
+	providerconfig.OperatingSystemCoreos: compute.ImageReference{
+		Publisher: to.StringPtr("CoreOS"),
+		Offer:     to.StringPtr("CoreOS"),
+		Sku:       to.StringPtr("Stable"),
+		Version:   to.StringPtr("latest"),
+	},
+	providerconfig.OperatingSystemCentOS: compute.ImageReference{
+		Publisher: to.StringPtr("OpenLogic"),
+		Offer:     to.StringPtr("CentOS"),
+		Sku:       to.StringPtr("7.4"),
```
alvaroaleman: The current CentOS 7.5 is not available on Azure?

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,581 @@
...
+	c.SubnetName, err = p.configVarResolver.GetConfigVarStringValue(rawCfg.SubnetName)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get the value of \"subnetName\" field, error = %v", err)
+	}
+
+	return &c, &pconfig, nil
+}
+
+func getVMIPAddresses(ctx context.Context, c *config, vm *compute.VirtualMachine) ([]string, error) {
+	ipAddresses := make([]string, 0)
```
alvaroaleman: I think its more idiomatic to write `var ipAddresses []string` instead of instantiating a zero-lenght array

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,581 @@
...
+
+	if vm.VirtualMachineProperties == nil {
+		return nil, fmt.Errorf("machine is missing properties")
```
alvaroaleman: I don't think we should do any such checks here, these should all happen inside `Validate()`

alvaroaleman: Oh, this is for the returned object. Maybe just put all the checks together and return an empty slice and no error if any of the properties is missing?

Because after all this is not a real error that has to be acted upon, it just means the instance is not ready yet

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,581 @@
...
+		return nil, err
+	}
+
+	// We genete a random SSH key, since Azure won't let us create a VM without an SSH key or a password
+	key, err := ssh.NewSSHKey()
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate ssh key: %v", err)
+	}
+
+	ifaceName := machine.Spec.Name + "-netiface"
```
alvaroaleman: Ugh, Azure doesnt allow to implicitly create the interface when creating the instance?

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,581 @@
...
+			},
+			OsProfile: &compute.OSProfile{
+				AdminUsername: to.StringPtr(adminUserName),
+				ComputerName:  &machine.Spec.Name,
+				LinuxConfiguration: &compute.LinuxConfiguration{
+					DisablePasswordAuthentication: to.BoolPtr(true),
+					SSH: &compute.SSHConfiguration{
+						PublicKeys: &[]compute.SSHPublicKey{
+							{
+								Path:    to.StringPtr(fmt.Sprintf("/home/%s/.ssh/authorized_keys", adminUserName)),
```
alvaroaleman: wtf.

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,581 @@
...
+	default:
+		glog.Warningf("unknown Azure power status %q", *powerStatus.Code)
+		return instance.StatusUnknown, nil
+	}
+}
+
+func (p *provider) Get(machine *v1alpha1.Machine) (instance.Instance, error) {
+	config, _, err := p.getConfig(machine.Spec.ProviderConfig)
+	if err != nil {
+		return nil, cloudprovidererrors.TerminalError{
```
alvaroaleman: I think we settled on throwing `TerminalError` only in `Create(): https://github.com/kubermatic/machine-controller/pull/148#pullrequestreview-110013150

CC @p0lyn0mial

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,581 @@
...
+		}
+	}
+
+	vm, err := getVMByUID(context.TODO(), config, machine.UID)
+	if err != nil {
+		if err == cloudprovidererrors.ErrInstanceNotFound {
+			return nil, cloudprovidererrors.ErrInstanceNotFound
+		}
+
+		return nil, fmt.Errorf("Failed to find machine %q by its UID: %s", machine.UID, err)
```
alvaroaleman: Returned errors must start lowercase, also you probably want to use `%v` to display the string val of the error

examples/machine-azure.yaml
```diff
@@ -0,0 +1,62 @@
...
+        secretKeyRef:
+          namespace: kube-system
+          name: machine-controller-azure
+          key: subscriptionID
+      location: "westeurope"
+      resourceGroup: "<< YOUR_RESOURCE_GROUP >>"
+      vmSize: "Standard_B1ms"
+      vnetName: "<< VNET_NAME >>"
+      subnetName: "<< SUBNET_NAME >>"
+      bootDiagnosticsEnabled: true
```
alvaroaleman: Is this used anywhere?

pkg/cloudprovider/provider/azure/create_delete_resources.go
mrIncompetent: If it's not necessary - we should add a flag to optionally add a public ip

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: https://github.com/terraform-providers/terraform-provider-azurerm/blob/aab882f476253e05bae288062c42421d7bf77746/azurerm/resource_arm_network_interface.go#L394

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: For now I need it for debugging, but yes, I will create a flag.

pkg/cloudprovider/provider/azure/provider.go
kdomanski: Oh yeah. A value is required but currently only one value is supported...

pkg/cloudprovider/provider/azure/provider.go
kdomanski: Not that I've seen.

pkg/cloudprovider/provider/azure/provider.go
kdomanski: Cannot find it. Also, there doesn't seem to be an "official" image.

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/provider.go
kdomanski: fixed

examples/machine-azure.yaml
kdomanski: Not anymore. Removed.

pkg/cloudprovider/provider/azure/create_delete_resources.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	list, err := ifClient.List(ctx, c.ResourceGroup)
+	if err != nil {
+		return fmt.Errorf("failed to list interfaces in resource group %q", c.ResourceGroup)
+	}
+
+	var allInterfaces []network.Interface
+
+	for {
```
guusvw: why not `for list.NotDone() {`?

pkg/cloudprovider/provider/azure/create_delete_resources.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	list, err := ipClient.List(ctx, c.ResourceGroup)
+	if err != nil {
+		return fmt.Errorf("failed to list public IP addresses in resource group %q", c.ResourceGroup)
+	}
+
+	var allIPs []network.PublicIPAddress
+
+	for {
```
guusvw: why not for list.NotDone() {?

pkg/cloudprovider/provider/azure/create_delete_resources.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	list, err := vmClient.ListAll(ctx)
+	if err != nil {
+		return err
+	}
+
+	var allServers []compute.VirtualMachine
+
+	for {
```
guusvw: same pattern as previous ;)

pkg/cloudprovider/provider/azure/create_delete_resources.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	list, err := disksClient.List(ctx)
+	if err != nil {
+		return err
+	}
+
+	var allDisks []compute.Disk
+
+	for {
```
guusvw: same pattern as previous ;)

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,615 @@
...
+		return nil, fmt.Errorf("machine has no network interfaces data")
+	}
+
+	for n, iface := range *vm.NetworkProfile.NetworkInterfaces {
+		if iface.ID == nil {
+			return nil, fmt.Errorf("interface %d has no ID", n)
+		}
+
+		splitIfaceID := strings.Split(*iface.ID, "/")
+		ifaceName := splitIfaceID[len(splitIfaceID)-1]
```
guusvw: Could you check the length of `splitIfaceID` before accessing a calculated index ;) will panic if the length is zero

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,615 @@
...
+			var name string
+			if conf.Name != nil {
+				name = *conf.Name
+			} else {
+				glog.Warning("IP configuration of NIC %q was returned with no name, trying to dissect the ID.", ifaceName)
+				if conf.ID == nil {
+					return nil, fmt.Errorf("IP configuration of NIC %q was returned with no ID", ifaceName)
+				}
+				splitConfID := strings.Split(*conf.ID, "/")
+				name = splitConfID[len(splitConfID)-1]
```
guusvw: have a look at line 210

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,615 @@
...
+	list, err := vmClient.ListAll(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	var allServers []compute.VirtualMachine
+
+	for {
+		allServers = append(allServers, list.Values()...)
+		if list.NotDone() {
```
guusvw: move ` list.NotDone()` into the condition for the for loop

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,615 @@
...
+	if err != nil {
+		return instance.StatusUnknown, fmt.Errorf("failed to get instance view for machine %q: %s", vmName, err)
+	}
+
+	if iv.Statuses == nil {
+		return instance.StatusUnknown, nil
+	}
+
+	// it seems that this field should contain two entries: a provisioning status and a power status
+	if len(*iv.Statuses) < 2 {
```
guusvw: So, why not checking for `if len(*iv.Statuses) = 2 {`

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,615 @@
...
+			return instance.StatusDeleting, nil
+		default:
+			glog.Warningf("unknown Azure provisioning status %q", *provisioningStatus.Code)
+			return instance.StatusUnknown, nil
+		}
+	}
+
+	// the second field is supposed to be the power status
+	// https://docs.microsoft.com/en-us/azure/virtual-machines/windows/tutorial-manage-vm#vm-power-states
+	powerStatus := (*iv.Statuses)[1]
```
guusvw: the length of the slice should previously checked ;)

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: Because that would skip the last page.

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: dito

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: dito

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: dito

pkg/cloudprovider/provider/azure/create_delete_resources.go
guusvw: https://sourcegraph.com/github.com/kubermatic/machine-controller@91b500fd63c88454374cdd949b9a1b75ccaa509e/-/blob/vendor/github.com/Azure/azure-sdk-for-go/services/network/mgmt/2018-04-01/network/models.go#L12852:59

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: done

pkg/cloudprovider/provider/azure/provider.go
kdomanski: Because sometimes it doesn't. :P

pkg/cloudprovider/provider/azure/provider.go
kdomanski: Because previous branch will return, when len < 2

pkg/cloudprovider/provider/azure/get_client.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/provider.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/provider.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/create_delete_resources.go
kdomanski: right. fixed

pkg/cloudprovider/provider/azure/provider.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/provider.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,616 @@
...
+		return nil, fmt.Errorf("failed to create IP address client: %v", err)
+	}
+
+	ip, err := ipClient.Get(ctx, c.ResourceGroup, addrName, "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get IP %q: %s", addrName, err)
+	}
+
+	if ip.IPConfiguration == nil {
+		return nil, fmt.Errorf("IP %q has nil IPConfiguration", addrName)
```
alvaroaleman: Please return lowercase errors only

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,581 @@
...
+		return cloudprovidererrors.TerminalError{
+			Reason:  v1alpha1.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Failed to parse MachineSpec, due to %v", err),
+		}
+	}
+
+	// TODO: Disassociate and remove the resources in reverse order, to prevent
+	// orphaned NICs, IPs and disks in case the operation is interrupted in the middle.
+	glog.Infof("deleting VM %q", machine.Name)
+	if err = deleteVMsByMachineUID(context.TODO(), config, machine.UID); err != nil {
```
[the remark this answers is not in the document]
alvaroaleman: This wasn't addresses, was it?

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,616 @@
...
+		return fmt.Errorf("failed to remove public IP addresses of machine %q: %s", machine.Name, err)
+	}
+
+	glog.Infof("deleting disks of VM %q", machine.Name)
+	if err = deleteDisksByMachineUID(context.TODO(), config, machine.UID); err != nil {
+		return fmt.Errorf("failed to remove disks of machine %q: %s", machine.Name, err)
```
alvaroaleman: err should be formatted via `%v`

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,616 @@
...
+	}
+
+	glog.Infof("deleting network interfaces of VM %q", machine.Name)
+	if err = deleteInterfacesByMachineUID(context.TODO(), config, machine.UID); err != nil {
+		return fmt.Errorf("failed to remove network interfaces of machine %q: %s", machine.Name, err)
```
alvaroaleman: err should be formatted via `%v`

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,616 @@
...
+	}
+
+	glog.Infof("deleting public IP addresses of VM %q", machine.Name)
+	if err = deleteIPAddressesByMachineUID(context.TODO(), config, machine.UID); err != nil {
+		return fmt.Errorf("failed to remove public IP addresses of machine %q: %s", machine.Name, err)
```
alvaroaleman: err should be formatted via `%v`

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -0,0 +1,616 @@
...
+
+func getVMStatus(ctx context.Context, c *config, vmName string) (instance.Status, error) {
+	vmClient, err := getVMClient(c)
+	if err != nil {
+		return instance.StatusUnknown, err
+	}
+
+	iv, err := vmClient.InstanceView(ctx, c.ResourceGroup, vmName)
+	if err != nil {
+		return instance.StatusUnknown, fmt.Errorf("failed to get instance view for machine %q: %s", vmName, err)
```
alvaroaleman: err should be formatted via `%v`

test/tools/integration/provision_master.sh
```diff
@@ -42,17 +42,17 @@ if ! which kubelet; then
...
 if ! ls \$HOME/.kube/config; then
   mkdir -p \$HOME/.kube
   cp -i /etc/kubernetes/admin.conf \$HOME/.kube/config
   kubectl taint nodes --all node-role.kubernetes.io/master-
 fi
-if ! ls kube-flannel.yml; then
-  curl -LO https://raw.githubusercontent.com/coreos/flannel/v0.9.1/Documentation/kube-flannel.yml
-  kubectl apply -f kube-flannel.yml
-fi
+
```
alvaroaleman: Please leave this script declaratively and only let it apply stuff that's not already there, makes local testing much faster

alvaroaleman: Also, are you sure flannel does not work? I would prefer to avoid having e2e tests here run with a different networking solution

pkg/cloudprovider/provider/azure/provider.go
guusvw: :+1:

pkg/cloudprovider/provider/azure/provider.go
guusvw: in case of `IP` I would be fine, but the linter will mark it also =/

pkg/cloudprovider/provider/azure/provider.go
guusvw: Oh, you're right

pkg/cloudprovider/provider/azure/provider.go
kdomanski: It doesn't mark it.

pkg/cloudprovider/provider/azure/provider.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/provider.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/provider.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/provider.go
kdomanski: fixed

pkg/cloudprovider/provider/azure/provider.go
alvaroaleman: But it will after #211

pkg/cloudprovider/provider/azure/provider.go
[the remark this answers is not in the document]
kdomanski: I don't see how we can recover from a partial error here. In case of a VM not found, this will simply return nil, since we're iterating through the list of all VMs.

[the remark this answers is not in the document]
mrIncompetent: Right, missed it. This function can be called over and over again, so it indeed recovers correctly

pkg/cloudprovider/provider/azure/provider.go
alvaroaleman: It doesn't allow Errors starting with uppercase letters. We can keep it as-is, but will create more efforts later on.

kdomanski:
Rebased against master

PR #235, opened 2018-06-04
Add migrator to machines.cluster.k8s.io

alvaroaleman:

Adds a migrator tool to migrate existing `machines.machine.k8s.io` to the new `machines.cluster.k8s.io` crd.
This does not change the controller in any way and if this is run, the controller will not work anymore.

The purpose of this PR is to give a heads-up and collect feedback. After this PR is merged, another one will be created to move the `machine-controller` to the new API group and slightly changed type.

As a side-effect, this downgrades `client-go` back to v6, because that is what is used inside the cluster-api repo and if not done, we got unresolvable dependencies.

Fixes #


p0lyn0mial:
thanks for letting us know, does it also mean that we have to adjust our API server at the same time as we are going to change machine controller ?

alvaroaleman:
@p0lyn0mial We import tagged versions of the machine-controller in kubermatic, thus we will have to adjust it sometime afterwards, yes. Should be quickly done thought.

alvaroaleman:
Thinking about this, we probably should only let the migrator run when it has the `machien-controller` leader lease to prevent concurrency issues

cmd/migrate/main.go
```diff
@@ -0,0 +1,161 @@
...
+
+func migrateIfNecesary(kubeClient kubernetes.Interface,
+	apiextClient apiextclient.Interface,
+	clusterv1alpha1Client clusterv1alpha1clientset.Interface,
+	config *restclient.Config) error {
+
+	_, err := apiextClient.ApiextensionsV1beta1().CustomResourceDefinitions().Get(downstreammachines.CRDName, metav1.GetOptions{})
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			return nil
```
p0lyn0mial: would be nice to inform the operator that there is noting to migrate since required `CRD` is not registered.

cmd/migrate/main.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		return fmt.Errorf("failed to get crds: %v", err)
+	}
+
+	downstreamClient, err := downstreammachineclientset.NewForConfig(config)
```
p0lyn0mial: what if there is more than one context/cluster in kubecofnig ?
in other words how do we handle multi-cluster scenario ?

cmd/migrate/main.go
```diff
@@ -0,0 +1,161 @@
...
+	flag.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+	flag.Parse()
+
+	cfg, err := clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
+	if err != nil {
+		glog.Fatalf("Error building kubeconfig: %v", err)
+	}
+
+	apiExtClient := apiextclient.NewForConfigOrDie(cfg)
+	clusterv1alpha1Client := clusterv1alpha1clientset.NewForConfigOrDie(cfg)
```
p0lyn0mial: do we want to ensure that the required `CRD` has been registred by calling `apiextClient.ApiextensionsV1beta1().CustomResourceDefinitions()...` ?

cmd/migrate/main.go
```diff
@@ -0,0 +1,161 @@
...
+	// Get downstreamMachines
+	downstreamMachines, err := downstreamClient.MachineV1alpha1().Machines().List(metav1.ListOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to list downstream machines: %v", err)
+	}
+
+	// Convert the machine, create the new machine, delete the old one, wait for it to be absent
+	// We do this in one loop to avoid ending up having all machines in  both the new and the old format if deletion
+	// failes for whatever reason
+	for _, downstreamMachine := range downstreamMachines.Items {
```
p0lyn0mial: should we discard machines that have  `downstreamMachine.DeletionTimestamp` field set ?

p0lyn0mial: should we discard machines that have `downstreamMachine Status.ErrorReason` field set ?

cmd/migrate/main.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		finalizers := sets.NewString(downstreamMachine.Finalizers...)
+		finalizers.Delete(controller.FinalizerDeleteInstance)
+		downstreamMachine.Finalizers = finalizers.List()
+		if _, err := downstreamClient.MachineV1alpha1().Machines().Update(&downstreamMachine); err != nil {
+			return fmt.Errorf("failed to update downstream machine %s after removing finalizer: %v", convertedClusterv1alpha1Machine.Name, err)
+		}
+		if err := downstreamClient.MachineV1alpha1().Machines().Delete(convertedClusterv1alpha1Machine.Name, nil); err != nil {
+			return fmt.Errorf("failed to delete machine %s: %v", convertedClusterv1alpha1Machine.Name, err)
```
p0lyn0mial: deleting the machine resource will effectively delete the provider instance.

cmd/migrate/main.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		if err = wait.Poll(500*time.Millisecond, 60*time.Second, func() (bool, error) {
+			return isDownstreamMachineDeleted(convertedClusterv1alpha1Machine.Name, downstreamClient)
+		}); err != nil {
+			return fmt.Errorf("failed to wait for machine %s to be deleted: %v", convertedClusterv1alpha1Machine.Name, err)
+		}
+	}
+	return nil
```
p0lyn0mial: would be nice to inform the operator that migration of the machine has been completed successfully.

cmd/migrate/main.go
```diff
@@ -0,0 +1,161 @@
...
+	cfg, err := clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
+	if err != nil {
+		glog.Fatalf("Error building kubeconfig: %v", err)
+	}
+
+	apiExtClient := apiextclient.NewForConfigOrDie(cfg)
+	clusterv1alpha1Client := clusterv1alpha1clientset.NewForConfigOrDie(cfg)
+	kubeClient := kubernetes.NewForConfigOrDie(cfg)
+
+	if err = migrateIfNecesary(kubeClient, apiExtClient, clusterv1alpha1Client, cfg); err != nil {
```
p0lyn0mial: how do we want to design that tool ? can we tolerate partial errors ? for example what if we successfully migrated 9 out of 10 machines ?

cmd/migrate/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+		ErrExpected       bool
+	}{
+		{
+			DownStreamMachine: &machinev1alpha1downstream.Machine{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-machine"},
+				Spec:       machinev1alpha1downstream.MachineSpec{ObjectMeta: metav1.ObjectMeta{Name: "test-node"}},
+			},
+			Node: &corev1.Node{ObjectMeta: metav1.ObjectMeta{Name: "test-node"}},
+		},
+		// We currently fail if no corresponding node exists. Is this desired?
```
p0lyn0mial: Given that the machine resource is healthy no errors and deletion hasn't been scheduled this is valid assumption.

cmd/migrate/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+	fakeclusterv1alpha1clientset "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/fake"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	fakekube "k8s.io/client-go/kubernetes/fake"
+)
+
+func TestMigrateMachines(t *testing.T) {
+	testcases := []struct {
```
p0lyn0mial: would be nice to see / test the machine object after migration.

cmd/migrate/main.go
alvaroaleman: I  do not think this will happen, because the migrator is supposed to run in an `initContainer`

cmd/migrate/main.go
alvaroaleman: I assumed this will panic if the crd is not registered. Am I wrong there?

cmd/migrate/main.go
alvaroaleman: I don't think so because they still need to be reconciled by the controller. Once we update the controller it should act exactly like before, just on a different type (That contains the very same information thought)

cmd/migrate/main.go
alvaroaleman: Only if a machine-controller is running. I will wrap the actual migration with  a leader-election to ensure no controller runs and no other migrator runs.

cmd/migrate/main.go
alvaroaleman: Then it is supposed to fail and try again. If it can not migrate the `machine-controller` will not come up as this runs in an `initContainer`. If that case every occurs an operator has to check what the issue is.

cmd/migrate/main_test.go
alvaroaleman: Well yeah but this is supposed to work even when a machine was just created or is misconfigured. 

I think I should change this to not require a corresponding node to exist because since it runs in an `initContainer` it will block any further reconciliation, meaning if you happen to deploy this while a machine did not have a corresponding node manual intervention is required. WDYT?

cmd/migrate/main_test.go
alvaroaleman: The tests here are supposed to test the `migrateMachiens` func from `cmd/migrate` which itself uses `pkg/api/conversions` to do the actual conversions.

The conversion also has tests, you can check the before/after by looking into `pkg/api/conversions/testdata/downstreammachines`/`pkg/api/conversions/testdata/migrated` respectively. 

It is probably best to check this out in your local IDE where you have a tree view for the diff. Or I could split this into dedicated PRs for:

* Downgrade client-go
* `pkg/api/conversions`
* `cmd/migrate`

Would that make sense?

cmd/migrate/main.go
mrIncompetent: I would be surprised if it would panic.
Creating a client does not trigger any api calls afaik.

If the CRD is not registered every call will return a `404 - not found`

cmd/migrate/main.go
mrIncompetent: You're just deleting one finalizer, but you actually need to migrate & delete all finalizers.

Otherwise you cannot guarantee that the machine gets deleted.
Also you should add a check afterwards to ensure that the machine resource is gone

mrIncompetent: Nevermind check exists

mrIncompetent: Other thing, maybe you should set the `controllerNameAnnotationKey` to ensure that no controller will process it.
You cannot guarantee that no controller runs concurrently -> `RollingUpdate` scenario

cmd/migrate/main.go
```diff
@@ -0,0 +1,161 @@
...
+	// Convert the machine, create the new machine, delete the old one, wait for it to be absent
+	// We do this in one loop to avoid ending up having all machines in  both the new and the old format if deletion
+	// failes for whatever reason
+	for _, downstreamMachine := range downstreamMachines.Items {
+		convertedClusterv1alpha1Machine, err := conversions.ConvertV1alpha1DownStreamMachineToV1alpha1ClusterMachine(downstreamMachine)
+		if err != nil {
+			return fmt.Errorf("failed to convert machine %s: %v", downstreamMachine.Name, err)
+		}
+
+		createdClusterV1alpha1Machine, err := clusterv1alpha1Client.ClusterV1alpha1().Machines(convertedClusterv1alpha1Machine.Namespace).Create(convertedClusterv1alpha1Machine)
```
p0lyn0mial: > Then it is supposed to fail and try again. If it can not migrate the machine-controller will not come up as this runs in an initContainer. If that case ever occurs an operator has to check what the issue is.

since we can try and fail, we should do `Get` before, Am I right ?

cmd/migrate/main_test.go
p0lyn0mial: yes, it does. thx.

cmd/migrate/main_test.go
p0lyn0mial: yeah, I think you are right.

cmd/migrate/main.go
p0lyn0mial: it sgtm.

pkg/api/conversions/conversions.go
```diff
@@ -0,0 +1,103 @@
...
+func ConvertV1alpha1DownStreamMachineToV1alpha1ClusterMachine(in machinev1alpha1downstream.Machine) (*clusterv1alpha1.Machine, error) {
+	out := &clusterv1alpha1.Machine{}
+	out.ObjectMeta = in.ObjectMeta
+	out.Spec.ObjectMeta = in.Spec.ObjectMeta
+	out.SelfLink = ""
+	out.UID = ""
+	out.ResourceVersion = ""
+	out.Generation = 0
+	out.CreationTimestamp = metav1.Time{}
+	out.ObjectMeta.Namespace = "kube-system"
```
p0lyn0mial: we are moving machines to `kube-system` ?

pkg/api/conversions/conversions.go
```diff
@@ -0,0 +1,103 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	clustercommon "sigs.k8s.io/cluster-api/pkg/apis/cluster/common"
+	clusterv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+func ConvertV1alpha1DownStreamMachineToV1alpha1ClusterMachine(in machinev1alpha1downstream.Machine) (*clusterv1alpha1.Machine, error) {
+	out := &clusterv1alpha1.Machine{}
+	out.ObjectMeta = in.ObjectMeta
+	out.Spec.ObjectMeta = in.Spec.ObjectMeta
```
p0lyn0mial: I think we can remove it as we are assigning `ObjectMeta` in line 37.

pkg/api/conversions/conversions.go
```diff
@@ -0,0 +1,103 @@
...
+	out.UID = ""
+	out.ResourceVersion = ""
+	out.Generation = 0
+	out.CreationTimestamp = metav1.Time{}
+	out.ObjectMeta.Namespace = "kube-system"
+
+	// sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1.MachineStatus and
+	// pkg/machines/v1alpha1.MachineStatus are semantically identical, the former
+	// only has one additional field, so we cast by serializing and deserializing
+	inStatusJSON, err := json.Marshal(in.Status)
```
p0lyn0mial: nice :)

pkg/api/conversions/conversions.go
```diff
@@ -0,0 +1,103 @@
...
+	for _, inRole := range in.Spec.Roles {
+		if inRole == machinev1alpha1downstream.MasterRole {
+			out.Spec.Roles = append(out.Spec.Roles, clustercommon.MasterRole)
+		}
+		if inRole == machinev1alpha1downstream.NodeRole {
+			out.Spec.Roles = append(out.Spec.Roles, clustercommon.NodeRole)
+		}
+	}
+
+	out.Finalizers = in.Finalizers
```
p0lyn0mial: nit: can we move this line to the top ?

then this block of code would only deal with the `Spec`.

pkg/api/conversions/conversions.go
```diff
@@ -0,0 +1,103 @@
...
+	// To work around this, we put it into the providerConfig
+	inMachineVersionJSON, err := json.Marshal(in.Spec.Versions)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal downstreammachine version: %v", err)
+	}
+	if err = json.Unmarshal(inMachineVersionJSON, &out.Spec.Versions); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal downstreammachine version: %v", err)
+	}
+
+	providerConfigMap, err := addContainerRuntimeInfoToProviderConfig(*out.Spec.ProviderConfig.Value,
```
p0lyn0mial: is `in.Spec.Version.ContainerRuntime` mandatory ?
does it make sense to conditionally check if it was set ?

pkg/api/conversions/conversions.go
```diff
@@ -0,0 +1,103 @@
...
+		providerConfigMap = map[string]interface{}{}
+	}
+	if val, ok := providerConfigMap["operatingSystemSpec"]; ok {
+		if valMap, ok := val.(map[string]interface{}); ok {
+			valMap["containerRuntimeInfo"] = containerRuntimeInfo
+			providerConfigMap["operatingSystemSpec"] = valMap
+			return providerConfigMap, nil
+		}
+	}
+	providerConfigMap["operatingSystemSpec"] = map[string]machinev1alpha1downstream.ContainerRuntimeInfo{"containerRuntimeInfo": containerRuntimeInfo}
```
p0lyn0mial: not sure if `containerRuntimeInfo` is part of `operatingSystemSpec`, how about making `containerRuntimeInfo` top level ?

pkg/api/conversions/conversions.go
```diff
@@ -0,0 +1,103 @@
...
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	clustercommon "sigs.k8s.io/cluster-api/pkg/apis/cluster/common"
+	clusterv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+func ConvertV1alpha1DownStreamMachineToV1alpha1ClusterMachine(in machinev1alpha1downstream.Machine) (*clusterv1alpha1.Machine, error) {
+	out := &clusterv1alpha1.Machine{}
+	out.ObjectMeta = in.ObjectMeta
+	out.Spec.ObjectMeta = in.Spec.ObjectMeta
+	out.SelfLink = ""
```
p0lyn0mial: will `out.APIVersion` and `Kind` be assigned by the API server ?

cmd/migrate/main.go
p0lyn0mial: okay

p0lyn0mial:
migration tool looks really solid :)
I have added some comments mostly questions.

pkg/api/conversions/conversions.go
alvaroaleman: No, in line 37 we assign `out.Spec.ObjectMeta`

pkg/api/conversions/conversions.go
alvaroaleman: I assumed this will implicitly be there because it is a `clusterv1alpha1.Machine{}`

pkg/api/conversions/conversions.go
alvaroaleman: Yes, they are now namespaced. We could maybe allow to overwrite that via a flag?

alvaroaleman:
I've addressed everything noted here except for the leader-election part and opened a dedicated PR for just the conversion code: #245

Lets get that one in, then continue the discussion on how to plumb it in here.

alvaroaleman:
Closing, this got implemented via #323

PR #236, opened 2018-06-05
cloudprovider/Azure: export RawConfig for use by other packages

kdomanski:
We need this to use the RawConfig struct in the cluster-controller

### DIFF
--- a/pkg/cloudprovider/provider/azure/provider.go
+++ b/pkg/cloudprovider/provider/azure/provider.go
@@ -32,7 +32,8 @@ type provider struct {
 	configVarResolver *providerconfig.ConfigVarResolver
 }
 
-type rawConfig struct {
+// RawConfig is a direct representation of an Azure machine object's configuration
+type RawConfig struct {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -32,7 +32,8 @@ type provider struct {
 	configVarResolver *providerconfig.ConfigVarResolver
 }
 
-type rawConfig struct {
+// RawConfig is a direct representation of an Azure machine object's configuration
+type RawConfig struct {
```
p0lyn0mial: I don't see the usage, only exporting ?
Are there any packages that want to use `RawConfig` ?

kdomanski: https://github.com/kubermatic/kubermatic/blob/master/api/pkg/resources/machine/machine.go

p0lyn0mial: oh, I see.

PR #240, opened 2018-06-15
Feature/239 v sphere network change

### LINKED ISSUE #239: Add support to specify Network on vSphere

pkavajin:
Makes the machine controller able to change the network of a newly created vm.

Fixes #239

There's also a matching PR in kubermatic/api.

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/helper.go
+++ b/pkg/cloudprovider/provider/vsphere/helper.go
@@ -179,6 +179,55 @@ func createLinkClonedVm(vmName, vmImage, datacenter, clusterName, folder string,
 	_, err = clonedVmTask.WaitForResult(ctx, nil)
 	return err
 }
+
+func setNetworkForVM(ctx context.Context, vm *object.VirtualMachine, netName string) error {
+	newNet, err := getNetworkFromVM(ctx, vm, netName)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getNetworkBackingInfoFromVM(ctx context.Context, vm *object.VirtualMachine, netName string) (*types.VirtualEthernetCardNetworkBackingInfo, error) {
+	devices, err := vm.Device(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get devices for vm, see: %s", err)
+	}
+
+	for _, device := range devices {
+		ethDevice, ok := device.(types.BaseVirtualEthernetCard)
+		if !ok {
+			continue
+		}
+
+		ethCard := ethDevice.GetVirtualEthernetCard()
+
+		if ethDevice.DeviceInfo.GetDescription().Label == netName {
+			return ethCard.Backing.GetVirtualDeviceBackingInfo().(types.VirtualEthernetCardNetworkBackingInfo), nil
+		}
+	}
+
+	return nil, fmt.Errorf("no networkbacking with the name %s found", netName)
+}
+
+func getNetworkFromVM(ctx context.Context, vm *object.VirtualMachine, netName string) (*types.ManagedObjectReference, error) {
+	cfg, err := vm.QueryConfigTarget(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get query config for vm, see: %v", err)
+	}
+
+	for _, net := range cfg.Network {
+		summary := net.Network.GetNetworkSummary()
+
+		if summary.Accessible && summary.Name == netName {
+			return summary.Network, nil
+		}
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -120,6 +124,16 @@ func (p *provider) getConfig(s runtime.RawExtension) (*Config, *providerconfig.C
 		return nil, nil, err
 	}
 
+	c.TemplateNetName, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.TemplateNetName)
```
alvaroaleman: If there is only one network on the templateVM, please default to that in `AddDefaults`

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -240,6 +254,12 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 		return nil, fmt.Errorf("failed to get virtual machine object: %v", err)
 	}
 
+	// Map networks
+	err = updateNetworkForVM(context.TODO(), virtualMachine, config.TemplateNetName, config.VMNetName)
```
alvaroaleman: Please only update the network if `VMNetName != ""` to not make this a required config option

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -179,6 +179,75 @@ func createLinkClonedVm(vmName, vmImage, datacenter, clusterName, folder string,
...
+	for _, device := range devices {
+		ethDevice, ok := device.(types.BaseVirtualEthernetCard)
+		if !ok {
+			continue
+		}
+
+		ethCard := ethDevice.GetVirtualEthernetCard()
+		ethBacking := ethCard.Backing.(*types.VirtualEthernetCardNetworkBackingInfo)
+
+		// if no network name is specified the first will be returned.
```
alvaroaleman: I would be more comfortable if this func just returns a list of backings and devices in case `netName == ""`. If I am not mistaken this is also how the builtin funcs in `govmi` do it

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -88,7 +92,73 @@ func (vsphereServer VSphereServer) Status() instance.Status {
...
+	cfg, _, rawCfg, err := p.getConfig(spec.ProviderConfig)
+	if err != nil {
+		return spec, changed, cloudprovidererrors.TerminalError{
+			Reason:  v1alpha1.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Failed to parse MachineSpec, due to %v", err),
+		}
+	}
+
+	// default templatenetname to first network of template if none specific was given.
+	if cfg.TemplateNetName == "" {
```
alvaroaleman: Please change the condition here to `if cfg.TemplateNetName == "" && config.VMNetName != ""` so we only act if is actually required

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -101,73 +171,83 @@ func getClient(username, password, address string, allowInsecure bool) (*govmomi
...
 	c.CPUs = rawConfig.CPUs
 	c.MemoryMB = rawConfig.MemoryMB
 
-	return &c, &pconfig, nil
+	return &c, &pconfig, &rawConfig, nil
 }
 
 func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
-	config, _, err := p.getConfig(spec.ProviderConfig)
+	config, _, _, err := p.getConfig(spec.ProviderConfig)
```
alvaroaleman: Please add a check here:  `if config.VMNetName != "" &&  if cfg.TemplateNetName == ""{ return err }`

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -88,7 +98,75 @@ func (vsphereServer VSphereServer) Status() instance.Status {
...
+		}
+
+		availableData, err := getNetworkDevicesAndBackingsFromVM(ctx, templateVm, "")
+		if err != nil {
+			return spec, changed, err
+		}
+		if len(availableData) == 0 {
+			return spec, changed, errors.New("found no network adapter in template vm")
+		}
+
```
alvaroaleman: please also check if `len(availableData) == 1`, otherwise we should not default. Also, please rename `availableData` to `availableNetworkDevices` so its a little easier to read :) Otherwise were good to go

PR #243, opened 2018-06-15
Add the delete finalizer at the latest possible point

### LINKED ISSUE #224: Machine with invalid cloud provider spec can never be deleted
A `Machine` object added with invalid or missing cloud provider access credentials will still have its finalizer added, which upon deletion will fail forever in a loop.

For example adding a `Machine` with `cloudProvider: "vsphere"` but missing `vsphereURL` will naturally fail to create properly, but attempting to delete the object will result in infinite
`machine1 failed with: failed to get instance for machine machine1 after the delete got triggered, due to failed to get vsphere client: 'Post /sdk: unsupported protocol scheme ""'`

alvaroaleman:
We currently add the delete finalizer to the machine at the very
beginning of the sync loop without even validating it before.

This moves the finalizer creation to happen right before the creation.

This will drastically reduce the occurence of situations where the
finalizer has to be manually deleted.

Related issue: #224

Fixes #224 

### DIFF
--- a/pkg/cloudprovider/provider/fake/provider.go
+++ b/pkg/cloudprovider/provider/fake/provider.go
@@ -1,0 +1,82 @@
+package fake
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider/cloud"
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider/instance"
+	"github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
+	"github.com/kubermatic/machine-controller/pkg/providerconfig"
+
+	"github.com/golang/glog"
+)
+
+type provider struct{}
+
+type FakeCloudProviderConfig struct {
+	PassValidation bool `json:"passValidation"`
+}
+
+type FakeCloudProviderInstance struct{}
+
+func (f FakeCloudProviderInstance) Name() string {
+	return ""
+}
+func (f FakeCloudProviderInstance) ID() string {
+	return ""
+}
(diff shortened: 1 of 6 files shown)


alvaroaleman:
I've now added tests, sorry this PR got that big, did not expect this would result in so much Yak shaving :S

I've also changed the logic a little again to either

* Ensure finalizer is there directly before creation
* Ensure finalizer is there if the cloud instance exists

The latter covers cases where someone manually deletes the finalizer

pkg/cloudprovider/provider/fake/provider.go
```diff
@@ -0,0 +1,82 @@
+package fake
```
mrIncompetent: So much love :heart:

pkg/controller/machine.go
```diff
@@ -102,12 +102,12 @@ type KubeconfigProvider interface {
 // MetricsCollection is a struct of all metrics used in
 // this controller.
 type MetricsCollection struct {
-	Machines            metrics.Gauge
-	Nodes               metrics.Gauge
-	Workers             metrics.Gauge
-	Errors              metrics.Counter
-	ControllerOperation metrics.Histogram
-	NodeJoinDuration    metrics.Histogram
+	Machines            prometheus.Gauge
```
mrIncompetent: Was it intended to move them in this PR as well?

alvaroaleman: This PR is the reason I did that change. The `syncHandler` crashes with an NPE if the metrics are not initialized.

I just put those changes in a dedicated PR to easen review, will rebase this one once the one got merged

PR #244, opened 2018-06-16
Move metric creation to controller

alvaroaleman:
We currently have helper funcs to get a metrics struct inside cmd. This
poses a problem when trying to write tests for the controller, as the
tests will run into a NPE when the controller has no metrics struct but
it is not possible to use the existing helper func as that would result
in a cyclic dependency.

This commit also removes github.com/go-kit/kit/metrics as we have no
plans to implement a metric provider other than Prometheus.

Fixes #

The issues with the current approach noted in the commit message arose when trying to write tests for #243. IMHO if an unused flexibility (We can pass pass in arbitrary metrics provider and metrics configuration) comes with the price of making it harder to write tests it is something we should drop.

If we ever really need some more flexibility, we can extend `NewMachineControllerMetrics()`


metalmatze:
For the tests now you simply instantiate those metrics and don't care about them?

alvaroaleman:
> For the tests now you simply instantiate those metrics and don't care about them?

Yep, exactly

alvaroaleman:
@metalmatze @mrIncompetent I missed the metrics registration :S PTAL

PR #249, opened 2018-06-21
Feature/openstack e2e

### LINKED ISSUE #248: Add openstack e2e tests
Openstack is currently not included in the end to end tests, we should change that.

alvaroaleman:

Fixes #248 

### DIFF
--- a/pkg/cloudprovider/provider/openstack/helper.go
+++ b/pkg/cloudprovider/provider/openstack/helper.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"errors"
 	"fmt"
+	"sync"
 
 	"github.com/gophercloud/gophercloud"
 	goopenstack "github.com/gophercloud/gophercloud/openstack"
@@ -20,7 +21,8 @@ import (
 )
 
 var (
-	errNotFound = errors.New("not found")
+	errNotFound               = errors.New("not found")
+	securityGroupCreationLock = sync.Mutex{}
 )
 
 const openstackFloatingIPErrorStatusName = "ERROR"
@@ -249,6 +251,12 @@ func getSubnet(client *gophercloud.ProviderClient, region, nameOrID string) (*os
 }
 
 func ensureKubernetesSecurityGroupExist(client *gophercloud.ProviderClient, region, name string) error {
+	// We need a mutex here because otherwise if more than one machine gets created at roughly the same time
+	// we will create two security groups and subsequently not be able anymore to identify our security group
+	// by name
+	securityGroupCreationLock.Lock()
+	defer securityGroupCreationLock.Unlock()
+
 	netClient, err := goopenstack.NewNetworkV2(client, gophercloud.EndpointOpts{Region: region})
 	if err != nil {
 		return osErrorToTerminalError(err, "failed to get network client")
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -114,7 +114,8 @@ func (p *provider) getConfig(s runtime.RawExtension) (*Config, *providerconfig.C
 	}
 	c.DomainName, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.DomainName, "OS_DOMAIN_NAME")
 	if err != nil {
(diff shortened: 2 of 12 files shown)


pkg/cloudprovider/provider/openstack/helper.go
```diff
@@ -249,6 +251,12 @@ func getSubnet(client *gophercloud.ProviderClient, region, nameOrID string) (*os
 }
 
 func ensureKubernetesSecurityGroupExist(client *gophercloud.ProviderClient, region, name string) error {
+	// We need a mutex here because otherwise if more than one machine gets created at roughly the same time
+	// we will create two security groups and subsequently not be able anymore to identify our security group
+	// by name
+	securityGroupCreationLock.Lock()
+	defer securityGroupCreationLock.Unlock()
```
mrIncompetent: Maybe we should remove that behavior at some time?
I dislike the fact that the controller manages more infrastructure than absolutely needed.

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -114,7 +114,8 @@ func (p *provider) getConfig(s runtime.RawExtension) (*Config, *providerconfig.C
 	}
 	c.DomainName, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.DomainName, "OS_DOMAIN_NAME")
 	if err != nil {
-		return nil, nil, nil, fmt.Errorf("failed to get the value of \"domainName\" field, error = %v", err)
+		glog.V(0).Infof("[warning] could not get the value of the \"domainName\" field, this means instance %s",
```
alvaroaleman: I am not entirely sure if this is the best way to handle that case but didn't come up with a better one. Ideally we would log that as event on the machine, however that would require passing an `EventRecorder` into the cloudprovides.

WDYT? Maybe other ideas?

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -114,7 +114,8 @@ func (p *provider) getConfig(s runtime.RawExtension) (*Config, *providerconfig.C
...
+			"creation will fail when using identity v3 api...")
```
mrIncompetent: Do we actually support something else than `v3` ?

alvaroaleman: Yes, e2e tests are running against v2 endpoint, works fine

pkg/controller/machine.go
```diff
@@ -507,17 +507,17 @@ func (c *Controller) ensureInstanceExistsForMachine(prov cloud.Provider, machine
...
-				return err
-			}
 			// Create the instance
 			if providerInstance, err = c.createProviderInstance(prov, machine, userdata); err != nil {
 				c.recorder.Eventf(machine, corev1.EventTypeWarning, "CreateInstanceFailed", "Instance creation failed: %v", err)
 				message := fmt.Sprintf("%v. Unable to create a machine.", err)
 				return c.updateMachineErrorIfTerminalError(machine, machinev1alpha1.CreateMachineError, message, err, "failed to create machine at cloudprover")
 			}
+			// We created the instance, so ensure finalizer is there
+			machine, err = c.ensureDeleteFinalizerExists(machine)
```
mrIncompetent: Days since last moving `ensureDeleteFinalizerExists` some other place: ~14~ -> 0
(Not really checked when we changed it last time)

pkg/cloudprovider/provider/openstack/provider.go
alvaroaleman: Not for sure, `v2` endpoints often have a `/v2.0` prefix but I do not think that is strictly required

pkg/cloudprovider/provider/openstack/helper.go
alvaroaleman: I guess we could but having it there is not much of an issue either

alvaroaleman:
@mrIncompetent PTAL

pkg/cloudprovider/provider/openstack/provider.go
mrIncompetent: So why then logging it at all?
It will fail during creation - which is fine. I would rely on the openstack error handling here.

alvaroaleman:
@mrIncompetent I've removed the logging, PTAL

PR #251, opened 2018-06-22
Improves `AddDeleteFinalizer` unit test by removing the use of inform…

p0lyn0mial:
**What this PR does / why we need it**: Improves `AddDeleteFinalizer` unit test by removing the use of informers.

**Special notes for your reviewer**: I was looking at this PR (https://github.com/kubermatic/machine-controller/pull/243/files) and noticed that we could slightly improve the unit test by removing the use of shared informers.

### DIFF
--- a/pkg/controller/machine_test.go
+++ b/pkg/controller/machine_test.go
@@ -235,6 +235,11 @@ func TestController_AddDeleteFinalizerOnlyIfValidationSucceeded(t *testing.T) {
 			if len(test.expectedActions) != len(fakeMachineClient.Actions()) {
 				t.Fatalf("unexpected actions %#v", fakeMachineClient.Actions())
 			}
+			for index, action := range fakeMachineClient.Actions() {
+				if !action.Matches(test.expectedActions[index], "machines") {
+					t.Fatalf("unexpected action %#v", action)
+				}
+			}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/machine_test.go
```diff
@@ -183,55 +184,59 @@ func TestController_AddDeleteFinalizerOnlyIfValidationSucceeded(t *testing.T) {
 		cloudProviderSpec string
 		finalizerExpected bool
 		err               string
+		expectedActions   []string
```
alvaroaleman: Wouldnt it make more sense to define this as int, if we only use its length anyways?

p0lyn0mial: ups, let me just update the code to compare actual action's verbs with the expected ones.

p0lyn0mial: I've just added the following code to verify the verbs of actions.

```
for index, action := range fakeMachineClient.Actions() {
	if !action.Matches(test.expectedActions[index], "machines") {
		t.Fatalf("unexpected action %#v", action)
	}
}
```

alvaroaleman: Nice! When are you going to to the "How to test k8s controllers" workshop? :P

p0lyn0mial: don't know if I could show anything more :P

PR #253, opened 2018-06-25
ensure a instance gets recreated after the instance got deleted at the cloud-provider

mrIncompetent:
Ensure a instance gets recreated after the instance got deleted at the cloud-provider.
Otherwise if a nodeRef exists on the machine it will enter a fail loop.

### DIFF
--- a/pkg/controller/machine.go
+++ b/pkg/controller/machine.go
@@ -228,6 +228,9 @@ func (c *Controller) doesNodeForMachineExistAndIsReady(machine *machinev1alpha1.
 	if machine.Status.NodeRef != nil {
 		listerNode, err := c.nodesLister.Get(machine.Status.NodeRef.Name)
 		if err != nil {
+			if kerrors.IsNotFound(err) {
+				return nil, false, nil
+			}
 			return nil, false, err
 		}
 		node := listerNode.DeepCopy()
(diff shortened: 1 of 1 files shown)


pkg/controller/machine.go
```diff
@@ -358,19 +358,24 @@ func (c *Controller) syncHandler(key string) error {
...
 		glog.V(6).Infof("Requesting instance for machine '%s' from cloudprovider because no associated node with status ready found...", machine.Name)
-		err = c.ensureInstanceExistsForMachine(prov, machine, userdataProvider, providerConfig)
-		if err != nil {
+		return c.ensureInstanceExistsForMachine(prov, machine, userdataProvider, providerConfig)
+	}
+
+	node, err := c.getNodeByNodeRef(machine.Status.NodeRef)
+	if err != nil {
+		//In case we cannot find a node for the NodeRef we must remove the NodeRef & recreate an instance on the next sync
+		if kerrors.IsNotFound(err) {
```
alvaroaleman: Actually I am not sure if this will do it, because if the instance at the cloud provider gets deleted manually, the node will always exist, but its state will not be ready. Kube does afaik not remove nodes just because they stop posting their status

alvaroaleman: I think what we need is `if nodes does not exist OR node status != ready { ensureInstanceExistsForMachine }`

mrIncompetent: Should be done in https://github.com/kubermatic/machine-controller/pull/253/commits/af28abac5176315327bf1fd69ebaedc45e8af08e

PR #256, opened 2018-06-28
Add Kubernetes 1.11 support

### LINKED ISSUE #254: Extend e2e testing to different kubernetes versions
We currently run e2e tests only against one Kubernetes version, 1.9.6. Changing this requires adjustments in the test suite, because CRI-O versions are only compatible with exactly one Kubernetes version and we have a test that has a hardcoded Cri-O 1.9 version in  it.

AC:

* We have at least one test for Kube versions 1.8, 1.9, 1.10, 1.11

alvaroaleman:

Fixes #254

* This also changes the baserepo for the hyperkube image to `k8s.gcr.io`
* This adds quite a lot of e2e tests


mrIncompetent:
Looks fine, but i would now extend the test-cases to cover the whole range of versions:
- 1.8
- 1.9
- 1.10
- 1.11

For Ubuntu that would mean 8 tests (Docker & CRI-O)

alvaroaleman:
For all distributions?

.circleci/config.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           name:  Run E2E test
           command: >-
               export KUBECONFIG=$GOPATH/src/github.com/kubermatic/machine-controller/.kubeconfig &&
-              go test -tags=e2e -parallel 24 -v -timeout 20m  ./test/e2e/... -identifier=$CIRCLE_BUILD_NUM ||
+              go test -tags=e2e -parallel 240 -v -timeout 30m  ./test/e2e/... -identifier=$CIRCLE_BUILD_NUM ||
```
p0lyn0mial: how about adding the `-race` flag :) ?

pkg/controller/machine.go
```diff
@@ -479,7 +488,9 @@ func (c *Controller) ensureInstanceExistsForMachine(prov cloud.Provider, machine
 			return fmt.Errorf("invalid provider config: %v", err)
 		}
 		c.recorder.Event(machine, corev1.EventTypeNormal, "ValidationSucceeded", "Validation succeeded")
+		c.validationCacheMutex.Lock()
```
p0lyn0mial: I think that the `-race` flag should detect concurrent writes to the map.

pkg/containerruntime/crio/supported_versions.go
```diff
@@ -22,6 +22,8 @@ func GetOfficiallySupportedVersions(kubernetesVersion string) ([]string, error)
 		return []string{"1.9"}, nil
 	case "1.10":
 		return []string{"1.10"}, nil
+	case "1.11":
+		return []string{"1.10"}, nil
```
mrIncompetent: Looking at the fedora packages, they have a `1.11` package.
But not in the ubuntu repos :/

pkg/controller/machine.go
```diff
@@ -479,7 +488,9 @@ func (c *Controller) ensureInstanceExistsForMachine(prov cloud.Provider, machine
...
 		c.validationCache[cacheKey] = true
+		c.validationCacheMutex.Unlock()
```
mrIncompetent: Why do we need a mutex here?
As a machine does not get processed concurrently i don't see a any benefit here.
Sure workers access the map concurrently but always with a different key

alvaroaleman: I introduced it after my local test machine controller panicked with a concurrent write to the sync map

pkg/userdata/centos/userdata.go
```diff
@@ -138,6 +144,7 @@ func (p Provider) UserData(spec machinesv1alpha1.MachineSpec, kubeconfig *client
 		ClusterDNSIPs:     clusterDNSIPs,
 		KubeadmCACertHash: kubeadmCACertHash,
 		ServerAddr:        serverAddr,
+		IsKube18:          isKube18,
```
mrIncompetent: Sprig has the function `semverCompare` http://masterminds.github.io/sprig/semver.html

pkg/controller/machine.go
p0lyn0mial: I think that the golang runtime detects concurrent writes to the map and marks them as an error   .

mrIncompetent: Good to know, though always it only does this for concurrent access on the same key

test/e2e/provisioning/helper.go
```diff
@@ -11,55 +11,106 @@ import (
...
 var scenarios = []scenario{
 	{
-		name:                    "scenario 1 Ubuntu Docker 1.13",
-		osName:                  "ubuntu",
-		containerRuntime:        "docker",
-		containerRuntimeVersion: "1.13",
+		name:              "Ubuntu Docker Kubernetes v1.8.13",
+		osName:            "ubuntu",
+		containerRuntime:  "docker",
+		kubernetesVersion: "1.8.13",
```
p0lyn0mial: is this kubernetesVersion or only kubelet version ?

pkg/userdata/ubuntu/testdata/docker-17.03-openstack-kubelet-v-version-prefix.golden
```diff
@@ -140,6 +140,7 @@ write_files:
     RemainAfterExit=true
     Environment="PATH=/sbin:/bin:/usr/sbin:/usr/bin:/opt/bin"
     ExecStartPre=/sbin/modprobe br_netfilter
+    ExecStartPre=/sbin/sysctl --system
```
p0lyn0mial: why do we need this command ?

pkg/userdata/ubuntu/testdata/cri-o-1.9-digitalocean.golden
```diff
@@ -78,29 +78,9 @@ write_files:
...
-  content: |
-    [Unit]
-    Requires=network-online.target
-    After=network-online.target
-
-    [Service]
-    Type=oneshot
-    RemainAfterExit=true
-    ExecStart=/usr/local/bin/download-crictl
+- path: /etc/sysctl.d/10-ipv4-forward.conf
```
p0lyn0mial: oh, that's why we need to fire `sysctl --system` ?

pkg/userdata/coreos/userdata.go
```diff
@@ -173,8 +173,9 @@ systemd:
         After=docker.service
         [Service]
         TimeoutStartSec=5min
-        Environment=KUBELET_IMAGE_TAG={{ .HyperkubeImageTag }}
+        Environment=KUBELET_IMAGE=docker://k8s.gcr.io/hyperkube-amd64:{{ .HyperkubeImageTag }}
         Environment="RKT_RUN_ARGS=--uuid-file-save=/var/cache/kubelet-pod.uuid \
+          --insecure-options=image \
```
p0lyn0mial: why do we need to pass this flag ?

README.md
```diff
@@ -94,9 +93,9 @@ locally, the following steps are required:
 Now you can either
 
 * Run the tests for all providers via
-  `go test -tags=e2e -parallel 24 -v -timeout 20m  ./test/e2e/... -identifier $USER`
+  `go test -race -tags=e2e -parallel 24 -v -timeout 20m  ./test/e2e/... -identifier $USER`
```
p0lyn0mial: should we adjust the values of -paraller and -timeout flags ?

pkg/userdata/coreos/userdata.go
alvaroaleman: Because otherwise `rkt` won't run a docker image: https://coreos.com/rkt/docs/latest/running-docker-images.html

.circleci/config.yml
```diff
@@ -64,7 +64,7 @@ jobs:
           name:  Run E2E test
           command: >-
               export KUBECONFIG=$GOPATH/src/github.com/kubermatic/machine-controller/.kubeconfig &&
-              go test -tags=e2e -parallel 24 -v -timeout 20m  ./test/e2e/... -identifier=$CIRCLE_BUILD_NUM ||
+              go test -race -tags=e2e -parallel 240 -v -timeout 30m  ./test/e2e/... -identifier=$CIRCLE_BUILD_NUM ||
```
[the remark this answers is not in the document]
alvaroaleman: I had a long negotiation with @mrIncompetent and got him down to that from his initial 999

test/e2e/provisioning/helper.go
alvaroaleman: Kubelet and kubeadm

pkg/userdata/ubuntu/testdata/cri-o-1.9-digitalocean.golden
alvaroaleman: Yup

.circleci/config.yml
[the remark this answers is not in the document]
p0lyn0mial: lol

test/e2e/provisioning/helper.go
p0lyn0mial: ok, thx.

pkg/userdata/ubuntu/testdata/cri-o-1.9-digitalocean.golden
p0lyn0mial: why do we want to enable IP forwarding then :) ?

.circleci/config.yml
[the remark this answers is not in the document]
mrIncompetent: To put this correct:
He didn't negotiated. He just ignored my direct command as teamlead. This will have consequences. :wink:

examples/machine-controller.yaml
```diff
@@ -24,6 +24,38 @@ subjects:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: machine-controller:kubelet-config-1.11
+  namespace: kube-system
+rules:
+- apiGroups:
+  - ""
+  resourceNames:
+  - kubelet-config-1.11
```
p0lyn0mial: if you don't mind, could you explain why we need to define this role ?

pkg/userdata/ubuntu/testdata/cri-o-1.9-digitalocean.golden
alvaroaleman: Because it is required by Kubernetes. Docker does enable it by default which is why it is only required for cri-o. The check was added in kubeadm 1.11 but it is required to be enabled on earlier versions as well: https://github.com/kubernetes/kubernetes/pull/63872

examples/machine-controller.yaml
alvaroaleman: Becuase kubeadm 1.11 uses dynamic kubelet config which means the kubelet gets its config from a configmap. This behaviour is not configurable: https://github.com/kubernetes/kubernetes/blob/3efd75d061f250108ca5cb5a9cc088518fb03676/cmd/kubeadm/app/cmd/join.go#L286-L289

alvaroaleman:
I think I've addressed everything, PTAL @mrIncompetent @p0lyn0mial

.circleci/config.yml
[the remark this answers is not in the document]
xrstf: Thanks for this laugh. It's now 6 years later and I was wondering where this magic 240 came from. It has survived all these years, the switch from Circle CI to Prow and even from Hetzner to kind. :grin: 

At least now I know it just stands for "some super large number".

PR #261, opened 2018-07-05
Avoid overwrite of cloud provider args in centos

alvaroaleman:
This is only required on CentOS as we do not install via package in
Ubuntu.

Fixes #


alvaroaleman:
CC @pkavajin

PR #268, opened 2018-07-12
Make provisioning more resilient

### LINKED ISSUE #269: missing finalizer prevents machines from being destroyed
The condition at https://github.com/kubermatic/machine-controller/blob/8787ff6157f760f4ecb0170fdcdac7b81a1e72f4/pkg/controller/machine.go#L349 means that a machine absolutely has to have the `machine-delete-finalizer` in order to be properly deleted.

I've encountered some rare state, where the machine already had a deletion timestamp and didn't have the `machine-delete-finalizer` , but had a `foregroundDeletion` finalizer. The machine controller promptly proceeded to try to initialize a node, which failed because an object cannot have more initializers added when it is marked for deletion - therefore the machine object could never be deleted.

mrIncompetent:
- Use kubernetes repos
- Statically define pgp keys for repos
- Add retry logic to scripts

Fixes #269


pkg/userdata/ubuntu/userdata.go
```diff
@@ -178,37 +178,188 @@ write_files:
   content: |
 {{ if ne .CloudConfig "" }}{{ .CloudConfig | indent 4 }}{{ end }}
 
+- path: "/etc/apt/sources.list.d/docker.list"
+  permissions: "0777"
```
alvaroaleman: Please do not make the apt lists executable, a 0644 should do just fine :)

pkg/userdata/ubuntu/userdata.go
```diff
@@ -178,37 +178,188 @@ write_files:
...
+    #!/bin/bash
+    set -xeuo pipefail
+    apt-key add /opt/docker.asc
+    apt-key add /opt/crio.asc
+    apt update
+
+    {{- if .OSConfig.DistUpgradeOnBoot }}
+    apt dist-upgrade -y
+    {{- end }}
+    if [[ -x /var/run/reboot-required ]]; then
```
alvaroaleman: That file is executable? I would expect it just exists or does not exist, so a [[ -e /var/run/reboot-required ]] would be appropriate, otherwise this will never match

test/tools/integration/provision_master.sh
```diff
@@ -59,6 +59,7 @@ if [[ "${1:-deploy_machine_controller}"  == "do-not-deploy-machine-controller" ]
 fi
 if ! ls machine-controller-deployed; then
   docker build -t kubermatic/machine-controller:latest .
+  sed -i -e 's/-worker-count=5/-worker-count=20/g' machine-controller.yaml
```
alvaroaleman: This is a noop because the `examples/machine-controller.yaml` file does not contain a `-worker-count=5` string but instead `-worker-count=20`  - Is it intended to specify 20 workers in the example file?

mrIncompetent: Not intended. For the tests though very helpful

pkg/userdata/ubuntu/userdata.go
```diff
@@ -178,37 +178,189 @@ write_files:
...
+    #!/bin/bash
+    set -xeuo pipefail
+    apt-key add /opt/docker.asc
+    apt-key add /opt/crio.asc
+    apt update
+
+    {{- if .OSConfig.DistUpgradeOnBoot }}
+    apt dist-upgrade -y
+    {{- end }}
+    if [[ -e /var/run/reboot-required ]]; then
```
alvaroaleman: Actually I think this is still a problem, because:

* The `install-depencies` will ultimatively be kicked of by the `kubeadm-join.service`
* `kubeadm-join.service` gets kicked off by cloud init which will only happen on the first boot, not after the reboot


We can however not just activate `kubeadm-join.service` because it will fail on subsequent attempts.

Not sure what the best approach here is. Make `kubeadm-join.service` idempotent by trying to "somehow" find out if the kubelet is already joined and not execute the join in that case and just quit?

pkg/controller/machine.go
```diff
@@ -346,8 +344,13 @@ func (c *Controller) syncHandler(key string) error {
 	}
 
 	// step 2: check if a user requested to delete the machine
-	if machine.DeletionTimestamp != nil && sets.NewString(machine.Finalizers...).Has(finalizerDeleteInstance) {
-		return c.deleteMachineAndProviderInstance(prov, machine)
+	if machine.DeletionTimestamp != nil {
+		if err := c.deleteMachineAndProviderInstance(prov, machine); err != nil {
+			return err
+		}
+		// Always requeue!
```
alvaroaleman: Those comments tend to be ignored if you don't provide a reason ;)

pkg/userdata/centos/testdata/docker-1.13-kubelet-v1.9-aws.golden
```diff
@@ -44,42 +35,75 @@ write_files:
...
+      kubelet-1.9.6 \
+      kubeadm-1.9.6 \
+      ebtables \
+      ethtool \
+      nfs-utils \
+      bash-completion \
+      sudo \
+      ${CNI_PKG}
+
+    cp /etc/sysconfig/kubelet-overwrite /etc/sysconfig/kubelet
```
alvaroaleman: Probably indeed the better approach as that will survive package updates

pkg/userdata/ubuntu/testdata/docker-1.13-dist-upgrade-on-boot-aws.golden
```diff
@@ -124,159 +197,7 @@ write_files:
...
-    Restart=always
-    StartLimitInterval=0
-    RestartSec=10
-    Restart=always
-
-    [Install]
-    WantedBy=multi-user.target
+    ExecStart=/usr/local/bin/supervise.sh /usr/local/bin/setup
 
 runcmd:
```
alvaroaleman: Please add a install section to the install service and use systemctl enable - - now l install.service here, otherwise this won't get started after a reboot e. G. due to updates. Other than that, really cool approach!

alvaroaleman: Also please call systemctl enable for both kubelet and docker in the setup script

pkg/userdata/ubuntu/userdata.go
```diff
@@ -178,306 +161,201 @@ write_files:
...
+  {{ else }}
+    export CR_PKG='{{ .CRAptPackage }}'
+  {{ end }}
+{{ end }}
+
+    # There is a dependency issue in the rpm repo for 1.8, if the cni package is not explicitly
+    # specified, installation of the kube packages fails
+    export CNI_PKG=''
+    {{- if semverCompare "=1.8.X" .KubernetesVersion }}
+    export CNI_PKG='kubernetes-cni=0.5.1-00'
```
alvaroaleman: Funny how this issue exists for both centos and Ubuntu

alvaroaleman:
Not force pushing has btw the nice advantage that you can pin CI failures onto specific changes ;)

PR #272, opened 2018-07-16
static IP management

kdomanski:
For now I only implemented this for CoreOS. After 2 days of trials, I still cannot make it work properly on Ubuntu.


alvaroaleman:
Pull looks good from my side, just one minor point: Could you add an e2e test for this? Its perfectly fine if it runs just on vsphere but I think that will require some adjustments in how the e2e tests work.

kdomanski:
The best way I can think of to do that would be picking an IP at random from a large pool. There will still be occasional collisions and failures. Got any better idea?

alvaroaleman:
Is there anything that prevents us from just hardcoding the IP inside the testcase?

kdomanski:
Let's see... ability to run multiple builds at the same time

alvaroaleman:
> Let's see... ability to run multiple builds at the same time

Arg, indeed. Define an IP pool of size 10 and use the last digit of the CircleCI build number?

mrIncompetent:
I would rather risk collisions instead of some vodoo.
During the tests, the machine only lives until it got ready - then it directly gets deleted.
So we have a very short timeframe in which those collisions could happen.

WDYT?

kdomanski:
Actually, I think @alvaroaleman's suggestion is quite a nice solution, if we go with it all the way, i.e. reserve an entire /24 block and set the last octet to `BUILD_NUMBER%256`.

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -152,3 +154,39 @@ func TestVsphereProvisioningE2E(t *testing.T) {
 	}
 	runScenarios(t, excludeSelector, params, vs_manifest, fmt.Sprintf("vs-%s", *testRunIdentifier))
 }
+
+// TestVsphereProvisioning - a test suite that exercises vsphere provider
+// by requesting nodes with different combination of container runtime type, container runtime version and the OS flavour.
+func TestVsphereStaticIPProvisioningE2E(t *testing.T) {
+	// test data
```
alvaroaleman: Please add `t.Parallel()` here

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -152,3 +154,39 @@ func TestVsphereProvisioningE2E(t *testing.T) {
...
+	vsAddress := os.Getenv("VSPHERE_E2E_ADDRESS")
+	if len(vsPassword) == 0 || len(vsUsername) == 0 || len(vsAddress) == 0 || len(vsCluster) == 0 {
+		t.Fatal("unable to run the test suite, VSPHERE_E2E_PASSWORD, VSPHERE_E2E_USERNAME, VSPHERE_E2E_CLUSTER or VSPHERE_E2E_ADDRESS environment variables cannot be empty")
+	}
+
+	buildNum, err := strconv.Atoi(os.Getenv("CIRCLE_BUILD_NUM"))
+	if err != nil {
+		t.Fatalf("failed to parse CIRCLE_BUILD_NUM: %s", err)
+	}
+	ipOctet := buildNum % 256
```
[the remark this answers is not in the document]
kdomanski: It's a `/20`. https://github.com/kubermatic/infra/blob/master/ansible/playbook.yaml#L5-L19

`192.168.32.1` is the gateway, `192.168.47.254` is broadcast.

mrIncompetent:
We should set an error message on the machine object when Ubuntu/Centos + Static ip gets set

test/e2e/provisioning/all_e2e_test.go
kdomanski: `testScenario` already adds `t.Parallel()`, and there are no subtest here, so we would be running `t.Parallel()` twice, causing a panic. That being said, I just noticed that the function's comment is wrong - I'll fix it promptly.

alvaroaleman:
> We should set an error message on the machine object when Ubuntu/Centos + Static ip gets set

Actually no, we should have validation fail.

mrIncompetent:
> > We should set an error message on the machine object when Ubuntu/Centos + Static ip gets set

> Actually no, we should have validation fail.

Thats what i meant! Validation! Leading to an error message being set on the machine :)

pkg/userdata/centos/userdata.go
```diff
@@ -94,6 +95,10 @@ func (p Provider) UserData(spec machinesv1alpha1.MachineSpec, kubeconfig *client
 		return "", fmt.Errorf("failed to get provider config: %v", err)
 	}
 
+	if pconfig.Network != nil {
```
alvaroaleman: Ideally we would have a `Validate` for the user data providers, I will create an issue for that

PR #273, opened 2018-07-17
fix a panic in Azure provider

kdomanski:


mrIncompetent:
Cherrypicked in https://github.com/kubermatic/machine-controller/pull/268

PR #278, opened 2018-07-19
use an instantiated prometheus registry instead of the global instance

kdomanski:
This allows creating multiple instance of machine controller, e.g. for testing.


metalmatze:
Thanks for taking care. Didn't know we want to run multiple instances. :innocent:

kdomanski:
We need that for better unit tests with multiple test cases. There's more singleton-like behavior there, e.g. https://github.com/kubermatic/machine-controller/blob/fb97b80461354c14660e60c7b1c2abf33ac24851/pkg/controller/machine.go#L124 but we can deal with it as we run into more testing issues.

PR #279, opened 2018-07-19
 pass a machine object updater to provider Create/Delete functions

kdomanski:
This will allow cloud provider packages to add/remove finalizers on machine objects.

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -503,7 +503,7 @@ func ensureDefaultInstanceProfileExists(client *iam.IAM) error {
 	return nil
 }
 
-func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.Instance, error) {
+func (p *provider) Create(machine *v1alpha1.Machine, update cloud.MachineUpdater, userdata string) (instance.Instance, error) {
 	config, pc, err := p.getConfig(machine.Spec.ProviderConfig)
 	if err != nil {
 		return nil, cloudprovidererrors.TerminalError{
@@ -630,7 +630,7 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 		Groups:     aws.StringSlice(securityGroupIDs),
 	})
 	if err != nil {
-		delErr := p.Delete(machine, awsInstance)
+		delErr := p.Delete(machine, update, awsInstance)
 		if delErr != nil {
 			return nil, awsErrorToTerminalError(err, fmt.Sprintf("failed to attach instance %s to security group %s & delete the created instance", aws.StringValue(runOut.Instances[0].InstanceId), defaultSecurityGroupName))
 		}
@@ -640,7 +640,7 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 	return awsInstance, nil
 }
 
-func (p *provider) Delete(machine *v1alpha1.Machine, instance instance.Instance) error {
+func (p *provider) Delete(machine *v1alpha1.Machine, update cloud.MachineUpdater, instance instance.Instance) error {
 	config, _, err := p.getConfig(machine.Spec.ProviderConfig)
 	if err != nil {
 		return cloudprovidererrors.TerminalError{
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -640,7 +640,7 @@ func (p *provider) Create(machine *v1alpha1.Machine, update cloud.MachineUpdater
 	return awsInstance, nil
 }
 
-func (p *provider) Delete(machine *v1alpha1.Machine, update cloud.MachineUpdater, instance instance.Instance) error {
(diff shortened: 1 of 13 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -640,7 +640,7 @@ func (p *provider) Create(machine *v1alpha1.Machine, userdata string) (instance.
 	return awsInstance, nil
 }
 
-func (p *provider) Delete(machine *v1alpha1.Machine, instance instance.Instance) error {
+func (p *provider) Delete(machine *v1alpha1.Machine, update cloud.MachineUpdater, instance instance.Instance) error {
```
alvaroaleman: Could we as convention assign unused function parameters to `_` like we already did in the `fake` cloud provider?

Same for all other providers `Create` / `Delete`

pkg/controller/machine.go
```diff
@@ -237,26 +238,47 @@ func (c *Controller) getNodeByNodeRef(nodeRef *corev1.ObjectReference) (*corev1.
...
 
-func (c *Controller) updateMachine(machine *machinev1alpha1.Machine) (*machinev1alpha1.Machine, error) {
-	machine.Status.LastUpdated = metav1.Now()
-	return c.machineClient.MachineV1alpha1().Machines().Update(machine)
+func (c *Controller) updateMachine(name string, modify func(*machinev1alpha1.Machine)) (*machinev1alpha1.Machine, error) {
+	var updatedMachine *machinev1alpha1.Machine
+	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		var retryErr error
+
+		//Get latest version from cache
```
mrIncompetent: comment does not match

pkg/controller/machine.go
```diff
@@ -237,26 +238,47 @@ func (c *Controller) getNodeByNodeRef(nodeRef *corev1.ObjectReference) (*corev1.
...
+		cacheMachine, err := c.machineClient.Machine().Machines().Get(name, metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+
+		currentMachine := cacheMachine.DeepCopy()
```
mrIncompetent: Not needed when we directly load it from the API

pkg/controller/machine.go
```diff
@@ -644,39 +669,38 @@ func (c *Controller) updateMachineStatus(machine *machinev1alpha1.Machine, node
...
-	}
 
-	if node.Status.NodeInfo.ContainerRuntimeVersion != "" {
-		runtimeName, runtimeVersion, err = parseContainerRuntime(node.Status.NodeInfo.ContainerRuntimeVersion)
-		if err != nil {
-			glog.V(2).Infof("failed to parse container runtime from node %s: %v", node.Name, err)
-			runtimeName = "unknown"
-			runtimeVersion = "unknown"
+	machine, err = c.updateMachine(machine.Name, func(m *machinev1alpha1.Machine) {
+		if !equality.Semantic.DeepEqual(m.Status.NodeRef, ref) {
```
mrIncompetent: Only put the logic to modify the machine object into the update function.
Otherwise we'll risk triggering an update without doing any changes - which could lead to triggering an update call during each sync.

pkg/controller/machine_test.go
```diff
@@ -2,28 +2,28 @@ package controller
...
-	machinelistersv1alpha1 "github.com/kubermatic/machine-controller/pkg/client/listers/machines/v1alpha1"
+	machineinformers "github.com/kubermatic/machine-controller/pkg/client/informers/externalversions"
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider/instance"
+	"github.com/kubermatic/machine-controller/pkg/clusterinfo"
 	"github.com/kubermatic/machine-controller/pkg/containerruntime"
 	"github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
 	"github.com/kubermatic/machine-controller/pkg/providerconfig"
 	"github.com/kubermatic/machine-controller/pkg/userdata"
 
+	"github.com/go-test/deep"
```
mrIncompetent: does not belong into the k8s.io group of imports

pkg/controller/machine_test.go
```diff
@@ -322,9 +312,48 @@ func TestController_defaultContainerRuntime(t *testing.T) {
...
+	kubeClient := kubefake.NewSimpleClientset()
+	machineClient := machinefake.NewSimpleClientset(objects...)
+
+	kubeInformersFactory := kubeinformers.NewSharedInformerFactory(kubeClient, time.Millisecond*50)
+	kubeSystemInformerFactory := kubeinformers.NewSharedInformerFactory(kubeClient, time.Millisecond*50)
+	machineInformerFactory := machineinformers.NewSharedInformerFactory(machineClient, time.Millisecond*50)
+
+	machineIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
+	for _, o := range objects {
+		if err := machineIndexer.Add(o); err != nil {
```
mrIncompetent: this should not be needed as the listers get filled with the objects from the client automatically when starting the informers

PR #286, opened 2018-07-25
Fix some CIS benchmark reported problems

### LINKED ISSUE #258: kube-bench kubelet issues
* [ ] --allow-privileged argument is set to false
* [ ] --anonymous-auth argument is set to false
* [ ] --read-only-port argument is set to 0
* [ ] --streaming-connection-idle-timeout argument is not set to 0
* [ ] --protect-kernel-defaults argument is set to true
* [ ] --make-iptables-util-chains argument is set to true
* [ ] --keep-terminated-pod-volumes argument is set to false
* [ ] --hostname-override argument is not set
* [ ] --event-qps argument is set to 0
* [ ] --tls-cert-file and --tls-private-key-file arguments are set as appropriate
* [ ] RotateKubeletClientCertificate argument is set to true
* [ ] RotateKubeletServerCertificate argument is set to true

kron4eg:
additional kubelet flags:

* --keep-terminated-pod-volumes=false
* --protect-kernel-defaults=true

Fixes some CIS Benchmark reported issues

Fixes #258


pkg/userdata/centos/testdata/docker-1.13-kubelet-v1.11-aws.golden
```diff
@@ -35,14 +33,18 @@ write_files:
...
-      --authentication-token-webhook=true --hostname-override=node1 --read-only-port 0 \
-      --cluster-dns= --cluster-domain=cluster.local
-
-
+    KUBELET_EXTRA_ARGS=--authentication-token-webhook=true \
+      --cloud-provider=aws \
+      --cloud-config=/etc/kubernetes/cloud-config \
+      --hostname-override=node1 \
+      --read-only-port=0 \
+      --keep-terminated-pod-volumes=false \
```
mrIncompetent: Got marked as deprecated https://github.com/kubernetes/kubernetes/pull/47539

mrIncompetent: Also it seems its default by default

alvaroaleman:
I am not sure if `--event-qps=0` is a good idea, now nodes can basically DDOS the api

kron4eg:
flag `--event-qps=0` was removed due to fears for api-server health (single node would be able to flood it).

cc @alvaroaleman

pkg/userdata/centos/testdata/docker-1.13-kubelet-v1.11-aws.golden
kron4eg: yep, to avoid problems like described here: https://github.com/kubernetes/kubernetes/issues/45191 it set to `false`

mrIncompetent: Ah sorry, its `false` by default. So we actually don't even need to set this flag

kron4eg: yeah, but CIS requires it to be like that :)

pkg/userdata/centos/testdata/docker-1.13-kubelet-v1.8-aws.golden
```diff
@@ -3,13 +3,14 @@ hostname: node1
...
-
-
 write_files:
 - path: "/etc/sysctl.d/k8s.conf"
   content: |
     net.bridge.bridge-nf-call-ip6tables = 1
     net.bridge.bridge-nf-call-iptables = 1
+    kernel.panic_on_oops = 1
+    kernel.panic = 10
+    vm.overcommit_memory = 1
```
alvaroaleman: This results in `The Linux kernel will always overcommit memory, and never check if enough memory is available. This increases the risk of out-of-memory situations, but also improves memory-intensive workloads.` according to  http://engineering.pivotal.io/post/virtual_memory_settings_in_linux_-_the_problem_with_overcommit - Do we really want that..?

kron4eg: I believe yes. That’s what kubelet wants to see, that’s what many memory intensive applications recorded (like JVM workloads or redis).

alvaroaleman: Well I guess the Kubelet will evict pod once the node has memory pressure so presumably this will not cause issues

PR #288, opened 2018-07-27
Fix Azure Validation & Finalizer assignment

alvaroaleman:

* Adds Validation for Azure Network & Subnet
* Moves the finalizer assignment after successful resource creation to avoid deadlocks
* Increases the e2e cleanup timeout as Azure may take quite some time there

Fixes e2e tests

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -179,8 +179,8 @@ func main() {
 	prometheusRegistry := prometheus.NewRegistry()
 
 	// before we acquire a lock we actually warm up caches mirroring the state of the API server
-	machineInformerFactory := machineinformers.NewSharedInformerFactory(machineClient, time.Minute*5)
-	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(kubeClient, time.Minute*5)
+	machineInformerFactory := machineinformers.NewSharedInformerFactory(machineClient, time.Minute*15)
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(kubeClient, time.Minute*15)
 	kubePublicKubeInformerFactory := kubeinformers.NewFilteredSharedInformerFactory(kubeClient, time.Second*30, metav1.NamespacePublic, nil)
 	kubeSystemInformerFactory := kubeinformers.NewFilteredSharedInformerFactory(kubeClient, time.Second*30, metav1.NamespaceSystem, nil)
 	defaultKubeInformerFactory := kubeinformers.NewFilteredSharedInformerFactory(kubeClient, time.Second*30, metav1.NamespaceDefault, nil)
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -179,8 +179,8 @@ func main() {
 	prometheusRegistry := prometheus.NewRegistry()
 
 	// before we acquire a lock we actually warm up caches mirroring the state of the API server
-	machineInformerFactory := machineinformers.NewSharedInformerFactory(machineClient, time.Second*30)
-	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(kubeClient, time.Second*30)
+	machineInformerFactory := machineinformers.NewSharedInformerFactory(machineClient, time.Minute*5)
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(kubeClient, time.Minute*5)
 	kubePublicKubeInformerFactory := kubeinformers.NewFilteredSharedInformerFactory(kubeClient, time.Second*30, metav1.NamespacePublic, nil)
 	kubeSystemInformerFactory := kubeinformers.NewFilteredSharedInformerFactory(kubeClient, time.Second*30, metav1.NamespaceSystem, nil)
 	defaultKubeInformerFactory := kubeinformers.NewFilteredSharedInformerFactory(kubeClient, time.Second*30, metav1.NamespaceDefault, nil)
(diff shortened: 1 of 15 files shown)


pkg/controller/machine/machine.go
```diff
@@ -496,10 +502,8 @@ func (c *Controller) ensureInstanceExistsForMachine(prov cloud.Provider, machine
 			if _, errNested := c.updateMachineError(machine, machinev1alpha1.InvalidConfigurationMachineError, err.Error()); errNested != nil {
 				return fmt.Errorf("failed to update machine error after failed validation: %v", errNested)
 			}
-			c.recorder.Eventf(machine, corev1.EventTypeWarning, "ValidationFailed", "Validation failed: %v", err)
 			return fmt.Errorf("invalid provider config: %v", err)
 		}
-		c.recorder.Event(machine, corev1.EventTypeNormal, "ValidationSucceeded", "Validation succeeded")
```
mrIncompetent: Does this mean we're now only emit the event for azure?
I dislike the idea of having to emit the event in the cloud-provider code - its redundant and easy to forget

alvaroaleman: https://github.com/kubermatic/machine-controller/pull/288/files#diff-87acff32b00826238809db10ac9ef562R309

It  just got moved inside the wrapper around the `Validate` we have within the machine-controller

mrIncompetent: Ah sorry - i misread my diff -  though it belongs to the azure code now

alvaroaleman: As a future enhancement I would however propose to pass down the EventRecorder to the `Create`/`Delete` funcs, because Azure takes forever to delete and as a User it is pretty hard to distinguish if its already doing something or not

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -348,19 +341,26 @@ func (p *provider) Create(machine *v1alpha1.Machine, update cloud.MachineUpdater
...
+		if !kuberneteshelper.HasFinalizer(machine, finalizerPublicIP) {
+			if machine, err = update(machine.Name, func(updatedMachine *v1alpha1.Machine) {
+				updatedMachine.Finalizers = append(updatedMachine.Finalizers, finalizerPublicIP)
+			}); err != nil {
+				return nil, err
+			}
+		}
 	}
 
+	iface, err := createNetworkInterface(context.TODO(), ifaceName, machine.UID, config, publicIP)
```
mrIncompetent: Sidequestion:
What happens if the `Create` fuctions dies after creating a network interface?
Will it reuse the already created interface on the next sync? Or will it fail with a `AlreadyExists` err?

alvaroaleman: I _think_ it will be a noop, because the underlying func from the Azure SDK is called `CreateOrUpdate`

mrIncompetent: Perfect!

alvaroaleman:
> The validation part is perfectly sensible, but moving the finalizer creation until after the resource creation can lead to resource leaks if the k8s api call fails.

> Cleanup will remove those finalizers whether or not the resources are actually there.

Kube has no transaction safety so we have to do one of the two (creating infra, adding finalizers) first. The chances of the creating infra part failing is much much higher than the chances of adding the finalizers failing, especially since the latter gets retried.

A middleground would be to check within the `Get` if the finalizers are there and if not add them but I am heavily opposed to adding them before infra creation succeeded.

WDYT?

alvaroaleman:
@kdomanski After discussing this with @mrIncompetent we settled on:

* Always call a cloud providers delete
* Expect the cloud providers delete to be implemented in a way that it can cope with the ressources already being removed

=> Now we can add the finalizers before creating ressources

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -450,16 +450,24 @@ func (p *provider) Create(machine *v1alpha1.Machine, update cloud.MachineUpdater
...
 	if err != nil {
 		return fmt.Errorf("failed to parse MachineSpec: %v", err)
 	}
 
-	glog.Infof("deleting VM %q", machine.Name)
-	if err = deleteVMsByMachineUID(context.TODO(), config, machine.UID); err != nil {
-		return fmt.Errorf("Is failed to remove public IP addresses of machine %q: %v", machine.Name, err)
+	_, err = p.Get(machine)
+	if err != nil {
+		if err != cloudprovidererrors.ErrInstanceNotFound {
```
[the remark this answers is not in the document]
alvaroaleman: Yes. Looking at the flow it seems what happens is:

* Initial Get succeeds, because otherwise we never execute the `Create`
* We execute the `Create` and fail
* Every subsequent `Get` fails because the config of the vm is invalid

I guess we must call the `DeleteVM` always except if we get a `ErrInstanceNotFound`

alvaroaleman:
@kdomanski PTAL

cmd/controller/main.go
```diff
@@ -179,8 +179,8 @@ func main() {
 	prometheusRegistry := prometheus.NewRegistry()
 
 	// before we acquire a lock we actually warm up caches mirroring the state of the API server
-	machineInformerFactory := machineinformers.NewSharedInformerFactory(machineClient, time.Second*30)
-	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(kubeClient, time.Second*30)
+	machineInformerFactory := machineinformers.NewSharedInformerFactory(machineClient, time.Minute*15)
+	kubeInformerFactory := kubeinformers.NewSharedInformerFactory(kubeClient, time.Minute*15)
```
mrIncompetent: oops

alvaroaleman: "Oops" as in..? I changed it because the current interval is so short it will re-qeue Azure machines that are still blocking a goroutine. Also IMHO if a machine needs the resync we have a bug in the controller

mrIncompetent: Opps as in, why did we ever had such a super short sync period?

PR #289, opened 2018-07-31
Remove --keep-terminated-pod-volumes=false

kron4eg:
--keep-terminated-pod-volumes is deprecated flag


alvaroaleman:
Why remove it again?

kron4eg:
@alvaroaleman it's deprecated, with functionality being removed. And it's `false` by default anyway.

PR #297, opened 2018-08-08
process only machines with a matching worker name

### LINKED ISSUE #111: running machine-controller through leader election is optional
the machine controller has been incorporated into kubermatic and is an inherent part of every cluster. 
That made local development/testing impossible as is highly possible that the machine controller which runs inside kubermatic will acquire a lock right before the local instance.

Making leader election optional seems to remedy this issue.

### LINKED ISSUE #233: Replace existing machine.k8s.io/controller annotation with a label
We currently use an annotation to determine if a given `machine-controller` should process a machine.

We should instead use a label selector, that way we can already excluded unwanted machines in the cache by using a `FilteredSharedInformerFactory`

kdomanski:
Fixes #233 
Fixes #111

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -113,7 +113,7 @@ type controllerRunOptions struct {
 	// kubeconfigProvider knows how to get cluster information stored under a ConfigMap
 	kubeconfigProvider machinecontroller.KubeconfigProvider
 
-	// name of the controller. When set the controller will only process machines with the annotation "machine.k8s.io/controller": name
+	// name of the controller. When set the controller will only process machines with the label "machine.k8s.io/controller": name
 	name string
 
 	// parentCtx carries a cancellation signal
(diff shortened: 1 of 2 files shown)


pkg/controller/machine/machine.go
```diff
@@ -327,21 +327,21 @@ func (c *Controller) syncHandler(key string) error {
 	machine := listerMachine.DeepCopy()
 
 	// step 1: check if the machine can be processed by this controller.
-	// set the annotation "machine.k8s.io/controller": my-controller
+	// set the label "machine.k8s.io/controller": my-controller
 	// and the flag --name=my-controller to make only this controller process a node
-	machineControllerName := machine.Annotations[controllerNameAnnotationKey]
+	machineControllerName := machine.Labels[controllerNameLabelKey]
```
alvaroaleman: The main idea behind using a label instead of an annotation is to be able to use a label selector within the informer factories so we do not have to build an exclusion step within the actual business logic

cmd/controller/main.go
```diff
@@ -416,3 +426,22 @@ func parseClusterDNSIPs(s string) ([]net.IP, error) {
...
 	return ips, nil
 }
+
+// return label selector to only process machines with a matching machine.k8s.io/controller label
+func labelSelector(workerName string) func(*metav1.ListOptions) {
+	return func(options *metav1.ListOptions) {
+		var req *labels.Requirement
+		var err error
+		if workerName == "" {
+			if req, err = labels.NewRequirement(controllerNameLabelKey, selection.DoesNotExist, nil); err != nil {
```
alvaroaleman: Is it possible to allow both `selection.DoesNoetExist` and empty string here?

kdomanski: It's not possible to select for both at the same time.

Issue #292, opened 2018-08-03
Limit size of logs coming from docker & kubelet


mrIncompetent:
We should limit the size of logs to 10gb

mrIncompetent:
On coreos & centos, docker logs to journald - which we limited via https://github.com/kubermatic/machine-controller/pull/301

On Ubuntu though, docker (Coming from the xenial repos) logs to file. When docker gets configured to use journald, it is not possible anymore to read container logs: https://github.com/moby/moby/issues/34066#issuecomment-314620360

We must set a logrotate config.

mrIncompetent:
Was it before or after the move to xenial ?

PR #301, opened 2018-08-20
userdata: limit journald disk usage to 10G

kdomanski:
Addresses half of #292


kdomanski:
@alvaroaleman I wrote "Addresses **half** of #292", the half in question being limiting the logs from kubelet.

alvaroaleman:
Okay, so please just write that in the PR description next time. Also I think this should be 5GB, else we end up with a total of 20GB instead of 10.

kdomanski:
It is in the PR description.

PR #302, opened 2018-08-21
Bring back node metrics

xrstf:

Node metrics (count of how many nodes currently exist) have been removed in #270 for no (to me) apparent reason. Since kubermatic/kubermatic#868 calls for a dashboard (and maybe alerts) that shows the difference between machines and nodes, we need the old node metric back.

The implementation is practically identical to the one for machine metrics, it's just using slightly different labels. I again chose to put a bunch of labels on the metric to provide data just in case we want to run analytics on it later. In a sample cluster with three nodes, the metrics look something like this:

```
# HELP machine_controller_machines
# TYPE machine_controller_machines gauge
machine_controller_machines{kubelet_version="1.10.6",os="centos",provider="digitalocean",region="fra1",size="s-1vcpu-1gb"} 1
machine_controller_machines{kubelet_version="1.10.6",os="ubuntu",provider="digitalocean",region="fra1",size="s-1vcpu-1gb"} 2

# HELP machine_controller_nodes
# TYPE machine_controller_nodes gauge
machine_controller_nodes{architecture="amd64",kubelet_version="v1.10.6",os="linux",osimage="CentOS Linux 7 (Core)",runtime="docker://1.13.1"} 1
machine_controller_nodes{architecture="amd64",kubelet_version="v1.10.6",os="linux",osimage="Ubuntu 16.04.5 LTS",runtime="docker://17.3.2"} 2
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -232,7 +232,8 @@ func main() {
 	var g run.Group
 	{
 		prometheusRegistry.MustRegister(machinecontroller.NewMachineCollector(
-			machineInformerFactory.Machine().V1alpha1().Machines().Lister(),
+			runOptions.machineLister,
+			runOptions.nodeLister,
 			kubeClient,
 		))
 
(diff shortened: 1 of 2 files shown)


xrstf:
This is obsolete, as kube-state-metrics already provides this data, but is not yet scraped by the seed-level Prometheus.

PR #305, opened 2018-08-30
Fix escaping in cloud-config

### LINKED ISSUE #304: Quotation marks break the cloud config
# User Story
As a user I want to be able to use the full range of allowed characters in passwords without breaking my cluster

## Acceptance criteria
* Any character (especially ") may be used in the cloud config fields where reasonable

# Further information
When a customer uses a " in the password, the quotation of the password breaks (ending at the ") and e.g. the controller-manager has troubles connecting to the cluster. This might also affect any other field in the cloud config, where " (and \ for escaping reasons) is used.

cbeneke:
As quotation marks may be part of the password in the cloud config,
these must be escaped in order to let the quoted string be correct.
Also backslashes are interpreted as special characters and must be
escaped likewise.
For convenience every field of the cloud-config gets replaced,
regardless of whether or not its (currently) a valid character.

When a customer uses a " in the password, the quotation of the password breaks (ending at the ") and e.g. the controller-manager has troubles connecting to the cluster. This PR addresses this issue.

Fixes #304

Currently changed in Azure and OpenStack. As far as I can tell, those should be the only ones affected.

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -547,13 +547,15 @@ func (p *provider) GetCloudConfig(spec v1alpha1.MachineSpec) (config string, nam
 
 	config = fmt.Sprintf(`
 [Global]
-auth-url = "%s"
-username = "%s"
-password = "%s"
-domain-name="%s"
-tenant-name = "%s"
-region = "%s"
-`, c.IdentityEndpoint, c.Username, c.Password, c.DomainName, c.TenantName, c.Region)
+auth-url = "%q"
+username = "%q"
+password = "%q"
+domain-name="%q"
+tenant-name = "%q"
+region = "%q"
+`, c.IdentityEndpoint, c.Username, c.Password, c.DomainName,
+		c.TenantName, c.Region)
+
 	return config, "openstack", nil
 }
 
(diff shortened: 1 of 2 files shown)


xrstf:
Can we switch to defining an inline struct with the required fields and then use encoding/json to properly marshal it? Replacing some characters seems to be a brittle solution.

alvaroaleman:
@xrstf Thats actually a very good idea, but it will require some further investigation/thinking because:

* Most providers do not use json for the cloud config but ini and AFAIK the std lib has no encoder for that
* The structs are not exported and we can not import kubernetes/kubernetes as thats too big - Maybe poke upstream, maybe just copy over?

I'd suggest we use this PR to get this fixed for the time being and create a ticket to switch it all over to annotated structs, WDYT?

xrstf:
SGTM

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -553,7 +556,9 @@ password = "%s"
 domain-name="%s"
 tenant-name = "%s"
 region = "%s"
-`, c.IdentityEndpoint, c.Username, c.Password, c.DomainName, c.TenantName, c.Region)
+`, re.Replace(c.IdentityEndpoint), re.Replace(c.Username), re.Replace(c.Password), re.Replace(c.DomainName),
```
alvaroaleman: Instead of the `strings.Replacer` approach, the `%s` in the fmt string should be changed to `%q`: `a double-quoted string safely escaped with Go syntax` as per https://golang.org/pkg/fmt/

mrIncompetent:
Structs & escaping have been implemented in https://github.com/kubermatic/machine-controller/pull/310

PR #306, opened 2018-08-30
Fix ubuntu bionic

kdomanski:
We migrated to Ubuntu Bionic, but that has a different set of Docker versions available.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/userdata/ubuntu/userdata.go
+++ b/pkg/userdata/ubuntu/userdata.go
@@ -197,7 +197,7 @@ write_files:
 
 - path: "/etc/apt/sources.list.d/docker.list"
   permissions: "0644"
-  content: deb [arch=amd64] https://download.docker.com/linux/ubuntu xenial stable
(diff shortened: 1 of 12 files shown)


pkg/userdata/ubuntu/userdata.go
```diff
@@ -197,7 +197,7 @@ write_files:
 
 - path: "/etc/apt/sources.list.d/docker.list"
   permissions: "0644"
-  content: deb [arch=amd64] https://download.docker.com/linux/ubuntu xenial stable
+  content: deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable
```
alvaroaleman: Not really happy with that because we do not always use bionic

mrIncompetent: We switched to bionic, didn't we?

kdomanski: Yes, we do use bionic now.

PR #308, opened 2018-08-30
Cluster api migrator

alvaroaleman:

Fixes #


alvaroaleman:
WRT the owner ref:

```
// List of objects depended by this object. If ALL objects in the list have
// been deleted, this object will be garbage collected. If this object is managed by a controller,
// then an entry in this list will point to this controller, with the controller field set to true.
// There cannot be more than one managing controller.
```
Source: https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta

So there may be multiple but there should only be one with controller=true.

alvaroaleman:
But I am not 100% sure what happens if the owner reference points to a GVK that does not exist anymore. I'll play around with that when adjusting the actual controller

PR #309, opened 2018-08-31
Use clusterapi in machine controller

alvaroaleman:

Fixes #


hack/run-machine-controller.sh
```diff
@@ -5,6 +5,7 @@ set -e
 make -C $(dirname $0)/.. machine-controller
 $(dirname $0)/../machine-controller \
   -kubeconfig=$(dirname $0)/../.kubeconfig \
+  -worker-count=50 \
```
mrIncompetent: :thinking:

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -136,9 +137,12 @@ func New(configVarResolver *providerconfig.ConfigVarResolver) cloud.Provider {
 	return &provider{configVarResolver: configVarResolver}
 }
 
-func (p *provider) getConfig(s runtime.RawExtension) (*config, *providerconfig.Config, error) {
+func (p *provider) getConfig(s v1alpha1.ProviderConfig) (*config, *providerconfig.Config, error) {
+	if s.Value == nil {
```
mrIncompetent: Maybe `len(s.Value) == 0` would be better, or is a initialized 0-len slice equal to `nil` ?

mrIncompetent: Nevermind, `s.Value` is not a slice

pkg/controller/machine/machine.go
```diff
@@ -420,28 +428,50 @@ func (c *Controller) cleanupMachineAfterDeletion(machine *machinev1alpha1.Machin
...
+func (c *Controller) deleteNodeForMachine(machine *clusterv1alpha1.Machine) error {
+	nodesList, err := c.nodesLister.List(labels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to list nodes: %v", err)
+	}
+
+	for _, node := range nodesList {
+		if ownerUID, exists := node.GetLabels()[NodeOwnerLabelName]; exists {
+			if ownerUID == string(machine.UID) {
+				return c.kubeClient.CoreV1().Nodes().Delete(node.Name, nil)
```
mrIncompetent: As i just found out about it: `c.nodesLister.ListWithPredicate` allows to pass a custom filter function.
So only objects, which apply to the filter function are being returned.

Just wanted to mention it, not requesting a change here.

pkg/controller/machine/machine.go
```diff
@@ -801,36 +799,32 @@ func (c *Controller) handleObject(obj interface{}) {
...
-		if err != nil {
-			utilruntime.HandleError(fmt.Errorf("error listing machines: '%v'", err))
-			return
-		}
-		for _, machine := range machines {
-			// We get triggered by node{Add,Update}, so enqeue machines if they
-			// have no nodeRef yet to make matching happen ASAP
+	var ownerUIDString string
+	var exists bool
+	if ownerUIDString, exists = object.GetLabels()[NodeOwnerLabelName]; !exists {
```
mrIncompetent: Do we have a guarantee that this always returns an initialized map?
I remember having issues with the annotations map and objects from the lister

pkg/controller/machine/machine.go
```diff
@@ -801,36 +799,32 @@ func (c *Controller) handleObject(obj interface{}) {
...
-		}
-		machine, err := c.machinesLister.Get(ownerRef.Name)
-		if err != nil {
-			glog.V(4).Infof("ignoring orphaned object '%s' of machine '%s'", object.GetSelfLink(), ownerRef.Name)
-			return
-		}
-
-		glog.V(6).Infof("Processing node: %s (machine=%s)", object.GetName(), machine.Name)
-		c.enqueueMachine(machine)
+	machinesList, err := c.machinesLister.List(labels.Everything())
```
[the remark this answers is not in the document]
mrIncompetent: Maybe i would like to know if we can move it below https://github.com/kubermatic/machine-controller/pull/309/files#diff-87acff32b00826238809db10ac9ef562R810

hack/run-machine-controller.sh
alvaroaleman: Need moar speed

pkg/controller/machine/machine.go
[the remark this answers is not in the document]
alvaroaleman: Actually that is not possible because:

* We check if labels exist
* if no: Iterate machines, enq all that have no nodeRef
* if yes: iterate machines, enq the one that has the ID from the label

Either way, we need the list

PR #310, opened 2018-09-06
Add cloud config structs

mrIncompetent:
This PR will introduce the structs used by the cloud-provider implementations in Kubernetes.

Also it will introduce functions to convert those structs to a string. This is required as the lib to parse ini-files in kubernetes can only read but not write. There are only a few libraries to write ini files, and all implement a different style of escaping - this i used a library which does not escape at all & added the escaping manually.

All introduced structs are exported, so they can be reused.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
     steps:
       - restore_cache:
           key: repo-{{ .Environment.CIRCLE_SHA1 }}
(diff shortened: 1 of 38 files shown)


mrIncompetent:
One thing to note on this PR:
It will drop support for kubernetes v1.8.
The reason is that the vSphere provider changed the format of the cloud-config from v1.8 to v1.9

PR #316, opened 2018-09-12
Improve FloatingIP allocation on OpenStack

### LINKED ISSUE #313: OpenStack: Race Condition on Floating IP assignments
When two or more clusters are created at the same time in the same OpenStack project in parallel, there is a very likely race condition when assigning Floating IPs to the newly created machines:

* There are no un-attached Floating IPs in the OpenStack project
* Cluster A is created
* Machine-controller for cluster A is created
* Cluster B is created
* Machine-controller for cluster B is created
* Machine-controller for cluster A checks if there is a free floating IP available. There is non. It creates a new floating IP id=15 and remembers its id
* Machine-controller for cluster A creates a new VM id=vm1  and waits until it is booted
* Machine-controller for cluster B checks if there is a free floating IP available, there is an unattached one with id=15 and it remembers its id
* Machine-controller for cluster B creates a new VM id=vm2 and waits until it is booted
* Machine-controller for cluster A sees that its new VM id=vm1 is booted and updates floating IP id=15 to be attached to the port of VM id=vm1
* Machine-controller for cluster B sees that its new VM id=vm2 is booted and updates floating IP id=15 to be attached to the port of VM id=vm2
* VM id=vm1 has no floating IP anymore. The machine-controller never cleans this up either.

Since it can take some time until a VM is ready, the time window for this to happen is quite large.

The described logic is implemented in https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/openstack/provider.go#L345-L466.

There is a floating IP lock mechanism, but this only works within one instances of a machine-controller and not across multiple machine-controller instances operation on the same OpenStack project.

To make this issue more unlikely, the floating IP could be retrieved, created and attached after the VM is ready. Also `osfloatingips.CreateOpts` allows to pass the VM PortID directly on creation so that only one operation to create and attach a new floating IP is needed.
There still would be a potential race condition when retrieving an existing floating IP and the updating it, but the time window would a lot smaller.

mrIncompetent:
This PR will reduce the critical time, in which a concurrent machine-controller can steal an already assigned FloatingIP.
As a cheap additional hack, a check was added to verify that the assigned FloatingIP is still assigned after 3 seconds.

Fixes #313

### DIFF
--- a/pkg/cloudprovider/provider/openstack/helper.go
+++ b/pkg/cloudprovider/provider/openstack/helper.go
@@ -4,6 +4,9 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
+
+	"github.com/golang/glog"
 
 	"github.com/gophercloud/gophercloud"
 	goopenstack "github.com/gophercloud/gophercloud/openstack"
@@ -25,7 +28,11 @@ var (
 	securityGroupCreationLock = sync.Mutex{}
 )
 
-const openstackFloatingIPErrorStatusName = "ERROR"
+const (
+	openstackFloatingIPErrorStatusName = "ERROR"
+
+	floatingReassignIPCheckPeriod = 3 * time.Second
+)
 
 func getRegion(client *gophercloud.ProviderClient, name string) (*osregions.Region, error) {
 	idClient, err := goopenstack.NewIdentityV3(client, gophercloud.EndpointOpts{})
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/openstack/helper.go
```diff
@@ -324,21 +331,65 @@ func getFreeFloatingIPs(client *gophercloud.ProviderClient, region string, float
...
+		ip = &freeFloatingIps[0]
+	}
+
+	_, err = osfloatingips.Update(netClient, ip.ID, osfloatingips.UpdateOpts{
 		PortID: &port.ID,
 	}).Extract()
+	if err != nil {
+		return fmt.Errorf("failed to update FloatingIP %s: %v", ip.ID, err)
+	}
+	secondsTook := time.Since(started).Seconds()
```
alvaroaleman: Feels to me like this should be a metric?

pkg/cloudprovider/provider/openstack/helper.go
```diff
@@ -324,21 +331,65 @@ func getFreeFloatingIPs(client *gophercloud.ProviderClient, region string, float
...
+	// We're now going to wait 3 seconds and check if the IP is still ours. If not, we're going to fail
+	// On our reference system it took ~3 seconds for a full FloatingIP allocation (Including creating a new one). It took ~600ms just for assigning one.
+	time.Sleep(floatingReassignIPCheckPeriod)
+	currentIP, err := osfloatingips.Get(netClient, ip.ID).Extract()
+	if err != nil {
+		return fmt.Errorf("failed to load FloatingIP %s after assignment has been done: %v", ip.FloatingIP, err)
+	}
+	// Verify if the port is still the one we set it to
+	if currentIP.PortID != port.ID {
+		return fmt.Errorf("FloatingIP %s got reassigned", currentIP.FloatingIP)
```
alvaroaleman: errors must start with lowercase

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -433,36 +409,63 @@ func (p *provider) Create(machine *v1alpha1.Machine, _ cloud.MachineUpdater, use
...
-				return true, nil
+	if err := waitUntilInstanceIsActive(computeClient, server.ID); err != nil {
+		defer deleteInstanceDueToFatalLogged(computeClient, server.ID)
+		return nil, fmt.Errorf("instance %s became not active: %v", server.ID, err)
+	}
+
+	// Find a free FloatingIP or allocate a new one
+	if c.FloatingIPPool != "" {
+		if err := assignFloatingIPToInstance(client, server.ID, c.FloatingIPPool, c.Region, network); err != nil {
+			defer deleteInstanceDueToFatalLogged(computeClient, server.ID)
```
alvaroaleman: What about instead making the `Create` of the openstack provider idempotent by calling `Get` before attempting to create the instance, then do the FIP stuff? I believe this is what we do in Azure and feels much more elegant to me

pkg/cloudprovider/provider/openstack/helper.go
mrIncompetent: I though about it, but decided not introduce a metric here.
The values would depend on creating a new FloatingIP or just assigning.
And creating a metric for both parts individually, would mean creating a metric for each request to OpenStack effectively.

The time above is really just for informational purpose

pkg/cloudprovider/provider/openstack/provider.go
mrIncompetent: The controller always executes a `Get` before doing the `Create`. Only if the `Get` cannot be satisfied, a `Create` gets executed.

#### Azure does 2 relevant requests:
- Create the NIC - A `Get` would **not** be satisfied in case we fail here
- Create the VM with the NIC - A `Get` would **not** be satisfied in case we fail here

The `Get` with Azure gets satisfied by finding a VM which has the Machine-UID as a tag.
Therefore only if the last request succeeds, the `Get` would be satisfied.

#### In OpenStack, the flow is a bit different.
- Create the VM -  A `Get` would **not** be satisfied in case we fail here
- Get a FloatingIP -  A `Get` **would** be satisfied in case we fail here
- Assign the Floating IP - A `Get` **would** be satisfied in case we fail here

The `Get` with OpenStack gets satisfied by finding a VM which has the Machine-UID in the VM metadata.
Therefore it would get satisfied as soon as we have been able to create the VM.


A more solid solution would probably be to introduce some kind of `EnsureVMExists` - which replaces the `Get` and `Create` function. Though this has the potential to increase the number of API calls, done to the cloud-provider, to increase by quite a bit.

alvaroaleman: Why not just extend the `Get` of the Openstack provider to not succeed in case a FIP pool was specified but the instance does not have an FIP associated?

mrIncompetent: It would be a special and non expectable implementation of a `Get` operation - as the `Get` would now also need to verify if the instance at the cloud-provider is actually according to spec.
It feels like a very broken mix of a `Get/Create` & `Ensure` approach.

If we want to go that way, we should do it proper and not partially & weird.

alvaroaleman: Okay I see. The reason I am nitpicking on this so much is that I very much dislike that we recreate an instance just because FIP allocation & assignment failed due to concurrency issues.

Maybe we can create an issue to replace the existing `GET` & `Create` with an `Ensure` in the interface? Implementation should be quite trivial.

mrIncompetent: I created https://github.com/kubermatic/machine-controller/issues/318
Though i'm not entirely sure if its such an easy implementation

pkg/cloudprovider/provider/openstack/helper.go
```diff
@@ -324,21 +331,65 @@ func getFreeFloatingIPs(client *gophercloud.ProviderClient, region string, float
...
+	defer floatingIPAssignLock.Unlock()
+
+	freeFloatingIps, err := getFreeFloatingIPs(client, region, floatingIPPool)
+	if err != nil {
+		return osErrorToTerminalError(err, "failed to get free floating ips")
+	}
+
+	var ip *osfloatingips.FloatingIP
+	if len(freeFloatingIps) < 1 {
+		if ip, err = createFloatingIP(client, region, floatingIPPool); err != nil {
```
bashofmann: As an additional optimization, the PortID could be directly passed here when creating the floating IP, so that a new Floating IP would never be un-attached

alvaroaleman: Does that mean there will be one api call for both allocation to the project and association to the instance?

bashofmann: As far as I understand yes

PR #317, opened 2018-09-13
Add machineset controller

alvaroaleman:

Fixes #

This does not yet include updated examples/README, I will do that after `machineDeployments` are getting handled as well


examples/machine-controller.yaml
```diff
@@ -10,6 +10,42 @@ spec:
...
+  version: v1alpha1
+  scope: Namespaced
+  names:
+    kind: MachineDeployment
+    plural: machinedeployments
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: clusters.cluster.k8s.io
```
alvaroaleman: The reason for creating this and the `machinedeployments.cluster.k8s.io` crd is that the `SharedInfomers` controller starts informers for this in `startInformers` and spams the logs with errors if they are not present: https://github.com/kubermatic/machine-controller/pull/317/files#diff-261786ef9d08de00190759a59afc4763

PR #323, opened 2018-09-14
Use cluster.k8s.io types, add machineSet controller

alvaroaleman:

Fixes #


pkg/machines/register.go
```diff
@@ -58,7 +60,7 @@ func CustomResourceDefinitionExists(name string, clientset apiextensionsclient.I
 }
 
 func createCustomResourceDefinition(plural, kind string, clientset apiextensionsclient.Interface) error {
-	name := plural + "." + v1alpha1.GroupName
+	name := CRDName
```
mrIncompetent: Is that really necessary?

pkg/userdata/ubuntu/userdata.go
```diff
@@ -117,18 +119,18 @@ func (p Provider) UserData(
...
 
-	var crPkg, crPkgVersion string
-	if spec.Versions.ContainerRuntime.Name == containerruntime.Docker {
-		crPkg, crPkgVersion, err = getDockerInstallCandidate(spec.Versions.ContainerRuntime.Version)
-		if err != nil {
-			return "", fmt.Errorf("failed to get docker install candidate for %s: %v", spec.Versions.ContainerRuntime.Version, err)
-		}
-	} else {
-		return "", fmt.Errorf("unknown container runtime selected '%s'", spec.Versions.ContainerRuntime.Name)
+	if pconfig.ContainerRuntimeInfo.Name != containerruntime.Docker {
```
mrIncompetent: Just as a idea:
Should we remove the possibility to specify the docker-version via Machine manifest?
IMO the controller should be very opinionated here. It would make our code a bit easier.
But this has nothing to do with this PR

pkg/machines/register.go
alvaroaleman: The reason was I wanted to export it. I can also revert this part back and just hardcode the GVK in the migration/conversion

hack/run-machine-controller.sh
```diff
@@ -5,6 +5,7 @@ set -e
 make -C $(dirname $0)/.. machine-controller
 $(dirname $0)/../machine-controller \
   -kubeconfig=$(dirname $0)/../.kubeconfig \
+  -worker-count=50 \
```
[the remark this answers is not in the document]
alvaroaleman: Yeah, increases the performance of the local E2E tests quite a bit

pkg/userdata/ubuntu/userdata.go
alvaroaleman: Sounds like a good idea but I'd prefer not to do that within this pull, its already pretty huge

pkg/machines/register.go
mrIncompetent: I'm fine with the const - definitely a improvement. But do we still need to local `name` variable then?

pkg/userdata/ubuntu/userdata.go
mrIncompetent: Yes, definitely something for a follow up!

alvaroaleman:
@mrIncompetent Removed the superfluous variable, PTAL

PR #346, opened 2018-10-10
Drain nodes before deleting them

alvaroaleman:

Blocks node deletion until the node got drained and contains no pods that are not owned by a `DaemonSet` or are a mirror pod.

Fixes #

* A lot of inspiration was taken from https://github.com/openshift/cluster-api-provider-aws/pull/60
* After this PR, the machine-controller won't be able to delete all nodes for a cluster that has one or more `PodDisruptionBudgets` set, unless the nodes contain a `kubermatic.io/skip-eviction` annotation

### DIFF
--- a/pkg/node/eviction/eviction.go
+++ b/pkg/node/eviction/eviction.go
@@ -49,7 +49,7 @@ func (ne *NodeEviction) Run() error {
 		glog.V(4).Infof("Skipping eviction for node %s as it has a %s annotation", ne.nodeName, SkipEvictionAnnotationKey)
 		return nil
 	}
-	glog.V(4).Infof("Starting to evict node %s", ne.nodeLister)
+	glog.V(4).Infof("Starting to evict node %s", ne.nodeName)
 
 	if err := ne.cordonNode(node); err != nil {
 		return fmt.Errorf("failed to cordon node %s: %v", ne.nodeName, err)


pkg/node/eviction/eviction.go
```diff
@@ -0,0 +1,222 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/kubernetes"
+	listerscorev1 "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/util/retry"
+)
+
+const (
+	timeout                   = 60 * time.Second
```
mrIncompetent: This seems fairly low - Do we have some reference for other implementations?

mrIncompetent:
@xmudrii Please have a look as well

pkg/node/eviction/eviction.go
```diff
@@ -0,0 +1,222 @@
...
+
+// Run excutes the eviction
+func (ne *NodeEviction) Run() error {
+	listerNode, err := ne.nodeLister.Get(ne.nodeName)
+	if err != nil {
+		return fmt.Errorf("failed to get node from lister: %v", err)
+	}
+	node := listerNode.DeepCopy()
+	if node.Annotations != nil {
+		if _, exists := node.Annotations[SkipEvictionAnnotationKey]; exists {
```
mrIncompetent: Do we need the nil check when we just read from the map ?

pkg/node/eviction/eviction.go
```diff
@@ -0,0 +1,222 @@
...
+	if errs := ne.evictPods(podsToEvict); len(errs) > 0 {
+		return fmt.Errorf("failed to evict pods, errors encountered: %v", errs)
+	}
+	glog.V(4).Infof("Successfully evicted all pods for node %s!", ne.nodeName)
+
+	glog.V(4).Infof("Waiting for deletion of all pods for node %s", ne.nodeName)
+	if err := ne.waitForDeletion(podsToEvict); err != nil {
+		return fmt.Errorf("failed waiting for pods of node %s to be deleted: %v", ne.nodeName, err)
+	}
+	glog.V(4).Infof("All pods of node %s were successfully deleted", ne.nodeName)
```
mrIncompetent: Maybe just log the start & end of the eviction with level 4 - any message in between should be level 6.

pkg/node/eviction/eviction.go
```diff
@@ -0,0 +1,222 @@
...
+	podsToEvict, err := ne.getFilteredPods()
+	if err != nil {
+		return fmt.Errorf("failed to get Pods to evict for node %s: %v", ne.nodeName, err)
+	}
+	glog.V(4).Infof("Found %v pods to evict for node %s", len(podsToEvict), ne.nodeName)
+
+	if errs := ne.evictPods(podsToEvict); len(errs) > 0 {
+		return fmt.Errorf("failed to evict pods, errors encountered: %v", errs)
+	}
+	glog.V(4).Infof("Successfully evicted all pods for node %s!", ne.nodeName)
```
mrIncompetent: This should be the last log message - as the pod os only "evicted" when it's gone

pkg/node/eviction/eviction.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (ne *NodeEviction) cordonNode(node *corev1.Node) error {
+	_, err := ne.updateNode(func(n *corev1.Node) {
+		n.Spec.Unschedulable = true
+	})
+	if err != nil {
+		return err
+	}
+
+	// Be paranoid and wait until the change got propagated to the lister
```
mrIncompetent: Why exactly here?

mrIncompetent:
The code looks really, really good!
Maybe this is something which we can put upstream using a dedicated controller?

pkg/node/eviction/eviction.go
alvaroaleman: Because its after the node got set to `Unschedulable` and before we call `getFilteredPods` the purpose is to be really sure that the scheduler does not schedule any pods in between setting this and after `getFilteredPods` because it doesn't have the change yet

alvaroaleman: Yes I know that this might not be 100% accurate, however the chances of a node getting scheduled in exactly this timeframe is pretty low. I would prefer not to add a 5s sleep there, but will add a comment

alvaroaleman:
@mrIncompetent PTAL

pkg/controller/machine/machine.go
```diff
@@ -461,16 +462,23 @@ func (c *Controller) ensureMachineHasNodeReadyCondition(machine *clusterv1alpha1
...
+			return fmt.Errorf("failed to evict node %s: %v", machine.Status.NodeRef.Name, err)
+		}
+	}
+
 	if err := c.deleteCloudProviderInstance(prov, machine); err != nil {
 		c.recorder.Eventf(machine, corev1.EventTypeWarning, "DeletionFailed", "Failed to delete instance at cloud provider: %v", err)
 		return err
 	}
 
-	// Delete the node object after the instance is gone
```
alvaroaleman: This check is already in `deleteNodeForMachine` so having it here was redundant

alvaroaleman: https://github.com/kubermatic/machine-controller/pull/346/files#diff-87acff32b00826238809db10ac9ef562L530

alvaroaleman: Correction: This finalizer here is not in `deleteNodeForMachine` but it gets removed by `deleteCloudProviderInstance` so at this point here it is impossible that it still exists

pkg/node/eviction/eviction.go
```diff
@@ -0,0 +1,225 @@
...
+	listerNode, err := ne.nodeLister.Get(ne.nodeName)
+	if err != nil {
+		return fmt.Errorf("failed to get node from lister: %v", err)
+	}
+	node := listerNode.DeepCopy()
+	if _, exists := node.Annotations[SkipEvictionAnnotationKey]; exists {
+		glog.V(4).Infof("Skipping eviction for node %s as it has a %s annotation", ne.nodeName, SkipEvictionAnnotationKey)
+		return nil
+	}
+	glog.V(4).Infof("Starting to evict node %s", ne.nodeLister)
```
xmudrii: Should this be `nodeName` instead of `nodeLister`?

alvaroaleman:
@mrIncompetent @xmudrii Everything addressed, PTAL

PR #348, opened 2018-10-10
Remove finalizers when instance is gone

### LINKED ISSUE #345: Finalizers get removed before the instance is really gone
When deleting a machine, the finalizer which is responsible to delete the cloud provider instance (`machine-delete-finalizer`), gets removed before we know that the instance is gone.

As cloud providers usually do the deletion async, the machine controller should only remove the finalizer when the instance is gone.

mrIncompetent:
This PR will ensure that we remove the finalizers only when we know that all actions have been successful.
Being more specific:
- Only when the instances at the cloud provider cannot be found anymore*, the deletion will be considered successful and the finalizer will be removed.
*For AWS the instance state `terminated` is considered deleted as well.
- The node object will only be removed in case the cloud provider instance is gone - this will ensure that the kubelet does not restart & registers again before the instance got shutdown.

Fixes #345

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -711,8 +711,19 @@ func (p *provider) Get(machine *v1alpha1.Machine) (instance.Instance, error) {
 		return nil, cloudprovidererrors.ErrInstanceNotFound
 	}
 
+	i := inOut.Reservations[0].Instances[0]
+	if i.State != nil && i.State.Name != nil {
+		// We consider terminated instances as deleted / don't exist anymore.
+		// The reason is that AWS takes very long >10min to gc those instances.
+		// Customers don't get billed anymore for those instances and they also don't block the deletion of security groups, etc.
+		// In AWS terms, a terminated instance is deleted.
+		if *i.State.Name == ec2.InstanceStateNameTerminated {
+			return nil, cloudprovidererrors.ErrInstanceNotFound
+		}
+	}
+
 	return &awsInstance{
-		instance: inOut.Reservations[0].Instances[0],
+		instance: i,
 	}, nil
 }
 
--- a/pkg/controller/machine/machine.go
+++ b/pkg/controller/machine/machine.go
@@ -308,10 +308,6 @@ func (c *Controller) updateMachineErrorIfTerminalError(machine *clusterv1alpha1.
 	return fmt.Errorf("%s, due to %v", errMsg, err)
 }
 
-func (c *Controller) deleteProviderInstance(prov cloud.Provider, machine *clusterv1alpha1.Machine) error {
-	return prov.Delete(machine, c.updateMachine)
-}
-
 func (c *Controller) createProviderInstance(prov cloud.Provider, machine *clusterv1alpha1.Machine, userdata string) (instance.Instance, error) {
 	// Ensure finalizer is there
 	machine, err := c.ensureDeleteFinalizerExists(machine)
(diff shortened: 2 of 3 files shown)


pkg/controller/machine/machine.go
```diff
@@ -463,36 +459,58 @@ func (c *Controller) ensureMachineHasNodeReadyCondition(machine *clusterv1alpha1
...
+		return nil
 	}
-	if err := c.deleteNodeForMachine(machine); err != nil {
-		message := fmt.Sprintf("%v. Please manually delete %s finalizer from the machine object.", err, finalizerDeleteNode)
-		c.recorder.Eventf(machine, corev1.EventTypeWarning, "DeletionFailed", "Failed to delete node: %v", err)
-		return errors.New(message)
+
+	// Retrieve the instance from the cloud provider
+	if _, err := prov.Get(machine); err != nil {
+		if err == cloudprovidererrors.ErrInstanceNotFound {
```
alvaroaleman: I really dislike this change. This should be inside the provider code, it should simply block until it can't find the instance anymore.

mrIncompetent: The deletion of instances is async for most providers we have(Not sure about vSphere) - thus its a more reliable implementation if we have that check above the providers.

alvaroaleman: Okay, valid point. Can you please then add this info to the godoc of the interface definition? That the `Get` must return `cloudprovidererrors.ErrInstanceNotFound` when the instance does not exist?

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -711,8 +711,19 @@ func (p *provider) Get(machine *v1alpha1.Machine) (instance.Instance, error) {
 		return nil, cloudprovidererrors.ErrInstanceNotFound
 	}
 
+	i := inOut.Reservations[0].Instances[0]
```
alvaroaleman: Can we do do `len` checks before accessing those items?

pkg/controller/machine/machine.go
```diff
@@ -463,36 +459,58 @@ func (c *Controller) ensureMachineHasNodeReadyCondition(machine *clusterv1alpha1
...
 
-	if machine, err = c.updateMachine(machine, func(m *clusterv1alpha1.Machine) {
-		finalizers := sets.NewString(m.Finalizers...)
-		finalizers.Delete(finalizerDeleteInstance)
-		finalizers.Delete(finalizerDeleteNode)
-		m.Finalizers = finalizers.List()
-	}); err != nil {
-		return fmt.Errorf("failed to update machine after removing the delete instance finalizer: %v", err)
+	// Delete the node object after the instance is gone
+	if sets.NewString(machine.Finalizers...).Has(finalizerDeleteInstance) {
```
alvaroaleman: IMHO we should first delete the node, before we delete the instance, especially since node deletion may imply draining

pkg/cloudprovider/provider/aws/provider.go
mrIncompetent: We do already on https://github.com/kubermatic/machine-controller/pull/348/files#diff-4c53f3e4be9b1b46a8d55115391a44eaR710

pkg/controller/machine/machine.go
mrIncompetent: The kubelet might restart & registers new - It's a safety mechanism to prevent empty node objects.

But I just looked it up in upstream, and the kubelet actually never fatals during runtime - it just loggs the error.
So unless any external entity restarts the kubelet on failed health-checks (And the gracetime period must be very, very short) we're safe.

I'll update the pull

pkg/controller/machine/machine.go
alvaroaleman: But actually this is nos that simple, because we must reprocess the machine  which will not happen by itself- Should we just rely on `resyncInterval` for that?

mrIncompetent: The controller will reprocess the machine after 10 seconds max, see: https://github.com/kubermatic/machine-controller/pull/348/files#diff-87acff32b00826238809db10ac9ef562R391

pkg/controller/machine/machine.go
mrIncompetent: I changed the order in https://github.com/kubermatic/machine-controller/pull/348/commits/1e0d1d19356be90df3c0752b189fb404bd2a35bb

mrIncompetent:
@alvaroaleman Comments addresses: PTAL

pkg/controller/machine/machine.go
alvaroaleman: @mrIncompetent  Please add the expections regarding the `InstanceNotFound` err onto the godoc of the interface in `pkg/cloudprovider/cloud/provider.go`, then we are ready to go :)

pkg/controller/machine/machine.go
alvaroaleman: Okay sorry for the back and forth but I just thought about this again. So if we delete the `node` beforehand and remove the finalizer there is a decent chance it will just re-register and stay, right?

Maybe the right order should be:

* Drain + mark node as unschedulable
* Delete the instance
* Delete the node

WDYT?

mrIncompetent: The Kubelet only creates the Node object during startup - So the chance that the Node object appears again after it has been deleted is fairly small.

Deleting the instance before we remove the Node object will eliminate that chance though.
The above order sounds about right.

PR #355, opened 2018-10-14
reserve dedicated amount of system resources for kubelet and system components

### LINKED ISSUE #343: Reserve minimal amount of CPU & memory for non-kubernetes & kubelet work
To prevent pods from killing the host - via CPU load for example - `--kube-reserved` should be specified on each kubelet.

The amount to reserve must be defined & set via flag

mrIncompetent:
Reserve dedicated amount of system resources for kubelet and system components.

Fixes #343


pkg/userdata/helper/kubelet.go
```diff
@@ -36,7 +36,9 @@ const (
 --anonymous-auth=false \
 --protect-kernel-defaults=true \
 --cluster-dns={{ .ClusterDNSIPs | join "," }} \
---cluster-domain=cluster.local`
+--cluster-domain=cluster.local \
+--kube-reserved=cpu=100m,memory=100Mi,ephemeral-storage=1Gi \
```
alvaroaleman: I am not sure if this does what you expect it to do. From the docs:

```
Optionally, kubelet can be made to enforce kube-reserved and system-reserved by specifying kube-reserved & 
system-reserved values in the same flag. Note that to enforce kube-reserved or system-reserved,
 --kube-reserved-cgroup or --system-reserved-cgroup needs to be specified respectively.
```

Further on:

```
Note that Kubelet does not create --kube-reserved-cgroup if it doesn’t exist. Kubelet will fail if an invalid cgroup is specified.
Note that Kubelet does not create --system-reserved-cgroup if it doesn’t exist. Kubelet will fail if an invalid cgroup is specified.
```

So in order for `kube-reserved` and `system-reserved` to have any effect, we would need a cgroup for kubelet/system services, which would then limit (!) those. The docs on that:

```
Once adequate monitoring and alerting is in place to track kube system daemons, attempt to enforce kube-reserved based on usage heuristics.
```

```
If absolutely necessary, enforce system-reserved over time.
```

I am in favor of not doing anything with `kube-reserverd` or `system-reserved`. If any, we could increase `eviction-hard`, which already has a non-zero default of ` imagefs.available<15%,memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5%)` but I am not sure if that is required.

Docs: https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/

mrIncompetent: Following https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/#example-scenario the flags will ensure that a pod cannot consume 100% CPU, Memory or fs. This is just meant as a protection for the system in case a single pod starts going crazy.

mrIncompetent: To extend my previous comment:
If i understood the docs correctly, all pods launched by the kubelet will receive the `Allocatable` amount of resources of the node - unless the pod has resource limits for the containers.

The `Allocatable` of the node gets calculated by:
```
      Node Capacity
---------------------------
|     kube-reserved       |
|-------------------------|
|     system-reserved     |
|-------------------------|
|    eviction-threshold   |
|-------------------------|
|                         |
|      allocatable        |
|   (available for pods)  |
|                         |
|                         |
---------------------------
```

So having a node with 4 CPU's and having `--kube-reserved=cpu=100m`, a pod can consume 3.9cpu's max.
I very deliberately not enabled the enforcement of the `kube-reserved` or `system-reserved` as it would not just reserve the given amount - it would limit the cgroups to the amount.


I'll do a manual test to verify the above - finally some crypto mining

alvaroaleman: Just to repeat myself here:

> Note that to enforce kube-reserved or system-reserved, --kube-reserved-cgroup or --system-reserved-cgroup needs to be specified respectively.

For increasing just the `--eviction-threshold` we can adjust the default of that flag, we don't need `--kube-reserved`/`--system-reserved` for that

mrIncompetent:
Just did a test, all pods can consume 100% of the total available resources - regardless whats configures with the `reserved` flags. The default eviction flag should be sufficient

PR #356, opened 2018-10-15
Detach PVC disks before deleting a Node (Vsphere cloud provider)

domcar:
While using the machine controller I found out that Persistent Disk are removed when the node they are attached to, are deleted.
It looks like that the the machine-controller does not drain the node before deleting it, this could lead to data loss. This PR detaches the (PVC) disks before deleting the node

I used the version v0.7.21 of machine controller for my tests

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/helper.go
+++ b/pkg/cloudprovider/provider/vsphere/helper.go
@@ -1,6 +1,7 @@
 package vsphere
 
 import (
+	"path/filepath"
 	"bytes"
 	"context"
 	"encoding/base64"
@@ -408,6 +409,37 @@ func generateLocalUserdataISO(userdata, name string) (string, error) {
 	}
 
 	return isoFilePath, nil
+}
+
+func detachDisksPVC (virtualMachine *object.VirtualMachine, moVM mo.VirtualMachine) error {
+    // Get VM folder path
+	glog.Infof("VM URL: %v", moVM.Config.Files.VmPathName)
+	vm_dir := filepath.Dir(moVM.Config.Files.VmPathName)
+	glog.Infof("VM folder: %v", vm_dir)
+
+	// Get devices
+    glog.Infof("Getting pvc disks")
+	vmDevices, err := virtualMachine.Device(context.TODO())
+	if err != nil {
+		return fmt.Errorf("Failed to get device list: %v", err)
+	}
+	for _, device := range vmDevices {
+		if _, ok := device.(*types.VirtualDisk); ok {
+            disk_name := device.GetVirtualDevice().DeviceInfo.GetDescription().Label
+            disk_path := device.GetVirtualDevice().Backing.(types.BaseVirtualDeviceFileBackingInfo).GetVirtualDeviceFileBackingInfo().FileName
+            disk_dir := filepath.Dir(disk_path)
+            if (disk_dir != vm_dir) {
+                glog.Infof("Detaching disk: %v disk path: %v", disk_name, disk_path)
+	            err = virtualMachine.RemoveDevice(context.TODO(), true, device)
+	            if err != nil {
+                    return fmt.Errorf("failed to detach disk: %v", err)
+	            }
+            } else {
+                    glog.Infof("Ignoring disk: %v, disk path: %v", disk_name, disk_path)
+            }
+        }
+    }
+	return nil
 }
 
 func removeFloppyDevice(virtualMachine *object.VirtualMachine) error {
--- a/pkg/cloudprovider/provider/vsphere/provider.go
+++ b/pkg/cloudprovider/provider/vsphere/provider.go
@@ -433,6 +433,16 @@ func (p *provider) Delete(machine *v1alpha1.Machine, _ cloud.MachineUpdater) err
 		return fmt.Errorf("failed to get virtual machine object: %v", err)
 	}
 
+	// Detach PVC Disks before deleting VM
+	var moVirtualMachine mo.VirtualMachine
+	pcollect := property.DefaultCollector(client.Client)
+	err = pcollect.RetrieveOne(context.TODO(), virtualMachine.Reference(), []string{"config.files"}, &moVirtualMachine)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve VM configuration: %v", err)
+	}
+	err = detachDisksPVC(virtualMachine,moVirtualMachine)
+
+
 	// We can't destroy a VM thats powered on...
 	powerOffTask, err := virtualMachine.PowerOff(context.TODO())
 	if err != nil {
@@ -530,6 +540,7 @@ func (p *provider) Get(machine *v1alpha1.Machine) (instance.Instance, error) {
 	} else {
 		glog.Warningf("vmware guest utils for machine %s are not running, can't match it to a node!", machine.Spec.Name)
 	}
+
 
 	return Server{name: virtualMachine.Name(), status: status, addresses: addresses, id: virtualMachine.Reference().Value}, nil
 }


alvaroaleman:
@domcar Thanks for your PR!

Coincidentally we added support for draining nodes just recently: #346, it is part of the latest release `v0.9.5`. We switched to a different apigroup in `v0.9.0` thought, so while the `machine-controller` does migrate all machines, third-party controllers or apis that act on machines may need an update ;)

So currently, the only situation in which your PR would make a difference is when draining is explicitly disabled.

> While using the machine controller I found out that Persistent Disk are removed when the node they are attached to, are deleted.
Can you elaborate on that, what exactly does `removed` mean here?

domcar:
@alvaroaleman That's a good news!

What I mean with `removed` is the following: when a Pod has a PVC the actual vmdk disk gets attached to the VM where the pod is running.
When deleting the node abruptly, without drain, all disks attached to the VM get deleted (removed).

Although drain is the best solution, this PR could be still a "safe mechanism" in case some disks were not detached from the VM during drain.

alvaroaleman:
> When deleting the node abruptly, without drain, all disks attached to the VM get deleted (removed).

At least the PV does not get deleted as the PVC stays, since it has a lifecycle that is different from that of the Pod (That's the whole reason the PVC/PV concept got introduced).
So this would mean that the vsphere cloud provider deletes the underlying vmdk for a non-deleted PV if it is attached to a pod on a node that gets deleted?

alvaroaleman:
> Yes, this is what happens

Okay, wow. That is a pretty severe bug in the vsphere cloud provider then. I'll try to repro and get this fixed upstream. Can you tell me the Kubernetes version on which you experienced this issue?

If we can confirm this, we'll get your PR merged.

domcar:
1.11.3

But i'm not quite sure this is a cloud provider problem, the cloud providers takes care of the node from a "vsphere" point of view, if the code tells it to delete the node it can/must just do it.
I think the `drain` command was an important missing piece of machine-controller

alvaroaleman:
> But i'm not quite sure this is a cloud provider problem, the cloud providers takes care of the node from a "vsphere" point of view, if the code tells it to delete the node it can/must just do it.

There is no code that would delete a PVC just because the associated Pod got deleted. Hence the PV the PVC is bound cant be deleted either. So the only way this could happen is if the vsphere cloud provider deleted the vmdk on node deletion even thought the PV still exists which would be a very severe bug. Are you sure this is what happened?

tonyfahrion:
Hi @alvaroaleman,

I might be wrong, but when deleting the machine resource within Kubernetes, the machine-controller deletes the VM through the vSphere API, doesn't it? Then the VMware system does the rest, which involves removing all related disks associated with the VM.

Kind regards
Tony 🙂

tonyfahrion:
> Coincidentally we added support for draining nodes just recently: #346

😍 😍 😍 😍 😍

tonyfahrion:
Nevertheless, please see this PR as a fallback-on-serious-issues. If somehow the drain does fail or don't work as expected, it could help preventing datalost on VMware vSphere.

What do you think about this PR independently of the cloud-provider topic?

alvaroaleman:
> I might be wrong, but when deleting the machine resource within Kubernetes, the machine-controller deletes the VM through the vSphere API, doesn't it? Then the VMware system does the rest, which involves removing all related disks associated with the VM.

Ah, that is a more reasonable explanation. Thanks for the info!

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -410,6 +411,37 @@ func generateLocalUserdataISO(userdata, name string) (string, error) {
 	return isoFilePath, nil
 }
 
+func detachDisksPVC (virtualMachine *object.VirtualMachine, moVM mo.VirtualMachine) error {
+    // Get VM folder path
```
alvaroaleman: Please run `go fmt` over all the code you've changed/added

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -410,6 +411,37 @@ func generateLocalUserdataISO(userdata, name string) (string, error) {
...
+	glog.Infof("VM URL: %v", moVM.Config.Files.VmPathName)
+	vm_dir := filepath.Dir(moVM.Config.Files.VmPathName)
+	glog.Infof("VM folder: %v", vm_dir)
+
+	// Get devices
+    glog.Infof("Getting pvc disks")
```
alvaroaleman: Please use `glog.V(4).Infof` here and for all other log lines you added, because this is debug logging

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -410,6 +411,37 @@ func generateLocalUserdataISO(userdata, name string) (string, error) {
...
+	vmDevices, err := virtualMachine.Device(context.TODO())
+	if err != nil {
+		return fmt.Errorf("Failed to get device list: %v", err)
+	}
+	for _, device := range vmDevices {
+		if _, ok := device.(*types.VirtualDisk); ok {
+            disk_name := device.GetVirtualDevice().DeviceInfo.GetDescription().Label
+            disk_path := device.GetVirtualDevice().Backing.(types.BaseVirtualDeviceFileBackingInfo).GetVirtualDeviceFileBackingInfo().FileName
+            disk_dir := filepath.Dir(disk_path)
+            if (disk_dir != vm_dir) {
```
alvaroaleman: Is this the only option to find out if the disk is the root volume or not? Maybe we can instead get this info from the template vm and store it somewhere on the created VM? IIRC vsphere disks have a name, roght?

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -433,6 +433,16 @@ func (p *provider) Delete(machine *v1alpha1.Machine, _ cloud.MachineUpdater) err
...
 	}
 
+	// Detach PVC Disks before deleting VM
+	var moVirtualMachine mo.VirtualMachine
+	pcollect := property.DefaultCollector(client.Client)
+	err = pcollect.RetrieveOne(context.TODO(), virtualMachine.Reference(), []string{"config.files"}, &moVirtualMachine)
+	if err != nil {
+		return fmt.Errorf("failed to retrieve VM configuration: %v", err)
+	}
+	err = detachDisksPVC(virtualMachine,moVirtualMachine)
```
alvaroaleman: Unhandled `err`

pkg/cloudprovider/provider/vsphere/helper.go
domcar: > Maybe we can instead get this info from the template vm and store it somewhere on the created VM?

I think it's better to check in real time where the disk is, otherwise we could have some problems, for example, if we read this info from the template and store it, we don't cover the case if the VM gets moved.

> Is this the only option to find out if the disk is the root volume or not?

You can get the disk name and *usually* it is the `Hard Disk 1` the root disk, but there is no assurance that `Hard Disk 1` must be the root disk, that's why I opted for this other way

alvaroaleman: What I meant is: The disks have a name, right? And at creation time we could get that name from the template vm and then store it somewhere on the created vm, maybe as tag?

Thinking about it, since we create the vm we can just get the name of the root disk directly after creating it and then store it in a tag. What do you think?

domcar: I think our approach is different, we try to be independent of what the situation was when the VM was created, we can't think of all possibilities that change the initial situation. We try to solve the problem when it is needed, we look for the actual situation (should vs is, soll vs ist Zustand).

I hope you understand what I mean.

alvaroaleman: @domcar My point is, I am not confident the location of the vmdk is a reliable way of finding out if its the rootdisk or not. This is why I would prefer something else. One option is to store the name/id of the rootdisk on the VM at creation time. Another option could be the name, because it looks like the vsphere cloud provider names all its disk something like `kubernetes-dynamic-pvc-.*` . But I am not entirely sure if we can trust in that this will never change ;)

I am fine with adding the location-based code as a fallback if all other ways of getting that info fail, but I do not want it to be the primary way of finding the root disk.

Would you be fine with that? Or are there reasons why you think that is not a good idea?

alvaroaleman: > @alvaroaleman I'll try to implement the tag at creation time and combine it with my original way as fallback.

Great, thank you!

> While trying out the version v0.9.5 the problem is there again even with drain
Oh. So the cloud provider probably does not detach on node delete, only when the pod gets started somewhere else? :thinking: 
I think on all other cloud providers, disks wont get deleted just because the instance they are attached to are deleted, so for everything except vsphere that would be finde. If I have time I'll check on that in upstream. Either way, we should definitely always detach everything before deleting nodes, just to be sure :)

domcar: @alvaroaleman After thinking on it again, probably the tag is not the best solution. It is very important to understand that a tag (or similar) cannot be a reliable way to store the information. 

From an developing point of view it would be just fine, just store the info there and retrieve. But this is fine in a ideal world where nothing changes. The infrastructure can change in time, I think we should focus on finding a solution that works at the moment when needed without relying on an information that as far as we know could outdated.
Do you understand my point of view?

alvaroaleman: Yes I do understand a tag could get lost or manipulated for various reasons, that is why we have the fallback. Your code should check if the tag points to a disk attached to the given instance and if not fall back to the location-based code.

The chances that the disk stored in the tag is not the root disk anymore but instead a pv is IMHO pretty near to impossible so I am fine taking that risk :)

domcar: That's the point, we cannot use the fallback because we don't know if the tag is correct or not: if the tag was changed/manipulated we have no way of finding out. 

In worst case scenario we use the wrong/outdated tag and risk detaching the wrong disk

alvaroaleman: Ofc we have a way of finding out, we can simply validate if the tag points to a disk attached to the current instance. If that is not the case we fall back.

The only situation where this could go wrong is if a pv gets attached to the node that has the same name as the root disk had. As mentioned earlier, I think this is near impossible to happen.

domcar: I honestly think the current solution is better: we find the disk and compare the path with the pvc.
When you say that: 
> think this is near impossible to happen.

Then my answer is that I think it's nearly impossible that a PVC is located in the same folder as the VM.
Look, I am not saying that my proposal is 100% safe, but your suggestion introduces more variability without getting to 100% reliability, in this case better to keep it simple

alvaroaleman: >  honestly think the current solution is better: we find the disk and compare the path with the pvc.

Where does your code compare anything with a pvc?

> Then my answer is that I think it's nearly impossible that a PVC is located in the same folder as the VM.

I strongly disagree. This is a current implementation detail that may or may not change. I do not want code here to depend on assumptions made in terms of how it is being used. I think the chance that the tag on the instance becomes invalid in any way is very unlikely, it would require someone manually fiddling with either the tag or the disk.

domcar: >Where does your code compare anything with a pvc?

Come on... I meant the disks

>This is a current implementation detail that may or may not change

Yes, but since PVC are independent of Nodes there is no chance that their path will be bounded to the VM path

domcar: I disagree

alvaroaleman: > I disagree

That's completely fine, it happens sometimes

Do I take from that that you do not want to write the corresponding code?

tonyfahrion: Hi @alvaroaleman, 🙂 

first of all, thank you for taking the time to tackle the challenges of this PR!

We currently try to come up with a better idea. You got me, when you said, that a VMs VMDKs might also live inside the directory, where the PVs are stored as VMDKs. This means, the current approach isn't sufficient and might brake at some point.

Using tags to store the information about system disks (we may use multiple in our templates), so we can differentiate them from PVs, would make us some trouble, as we are evaluating tags from VMs by other systems. In addition, I don't think we have to rely on a "cache" concept. The tag might not be sufficient as well (e.g. if a system disk gets attached later on).

To not leave you just with questions, I checked some possible options and I think I might have found a way to satisfy both requirements 🎉 🙂 

When a PV gets created, the vSphere Cloud Provider adds its physical location to the spec.
```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  ...
spec:
  ...
  storageClassName: vsphere
  vsphereVolume:
    fsType: ext4
    volumePath: '[datastore-name] kubevols/kubernetes-dynamic-pvc-9591e4ba-d3d9-11e8-8aa1-005056b6ce84.vmdk'
status:
  phase: Bound
```

So, I would raise the question, if we just might to test if a disk from the VM can be found inside K8s PVs? We have the full-path to match the VMDK in the PV spec, we only need a way to query the kubernetes API for a possible part of the VMDK name or if possible (I don't know the grpc API details here), filter results based on the `volumePath` content?
Currently we could extract everything between `pvc-` and `.vmdk` and set a _metadata name_ filter. If the name can be found inside k8s, it is a PV, if not - it should be save to be removed by VMware.

Might this solve our problem here?

Kind regards and have a nice weekend! 🙂

alvaroaleman: Hey @tonyfahrion 

Thanks for taking the time to elaborate on the issue :)

The approach to take the relevant info from the PVs is an excellent idea, that's how we are going to do it!

However this requires some more adjustments in order to be able to access the PV info from the providers. I'll prepare a PR for that and notify you once that is done.

And thanks again you two for finding and raising this issue and coming up with such a great solution!

tonyfahrion: Hi @alvaroaleman,

thanks! We're looking forward to it 🙂 

Happy hacking 🙂

alvaroaleman: @domcar #389 just got merged, which provides you with a `PVLister` during deletion

domcar: @alvaroaleman I'll have a look at it as soon as I can

domcar: @alvaroaleman Maybe I misunderstood, the #389 provides only a list of PVs, it doesn't detach them, correct? I thought you were going to detach them too, or you wanted me to add this functionality?

alvaroaleman: @domcar Yes, it does not detach them, just provides the required info. I assumed you wanted to work on that.

If you don't have time I am also fine taking care of it. What do you prefer?

alvaroaleman: @domcar #395 fixed the issue here :)

domcar: That's good, are you planning a release soon?

Issue #227, opened 2018-05-25
Implement admission controller for validation and defaulting


mrIncompetent:

p0lyn0mial:
related issue #224

p0lyn0mial:
this admission controller should be implemented as a webhook.
See https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#what-are-admission-webhooks

the API server that actually accepts/stores `machines` resources will make use of the webhook

mrIncompetent:
@alvaroaleman please provide your example code

alvaroaleman:
Here is an example mutating admission webhook: https://github.com/alvaroaleman/k8s-auth-injector

PR #362, opened 2018-10-18
Add a MutatingAdmissionWebook + Defaulting/Validation for MachineDeployment

alvaroaleman:

Add aa MutatingAdmissionWebook + Defaulting/Validation for MachineDeployment

Fixes #

* The defaulting and validation code  for the machineDeployment comse from the upstream clusterAPI repo, where it got removed when switching to CRDs:  https://github.com/kubernetes-sigs/cluster-api/commit/80129c0ca4eee12c47be0c04ea551fdd6f2cbaba#diff-bc9c8dee3c7f90be8876f1307cc221fdL177
* Defaulting of `MachineDeployment.Spec.Strategy` results in an invalid json patch, because its an optional struct field that is not a pointer. This can only be fixed by making it a pointer. Longer explanation here: https://github.com/kubernetes-sigs/cluster-api/pull/551
* In a follow-up I will move Defaulting+Validation of machines into the webhook

xref: #227


examples/machine-controller.yaml
```diff
@@ -189,6 +189,75 @@ spec:
               port: 8085
             periodSeconds: 5
 ---
+apiVersion: apps/v1beta2
```
alvaroaleman: Until I put this into a dedicated Deployment, the e2e tests always had some failures because of a connection reset when the APIServer tried to call the Webhook

alvaroaleman:
@mrIncompetent This passes tests now, PTAL :)

mrIncompetent:
Whats the advantage of using 2 separate docker images?
I would prefer just having a single one for simplicity

alvaroaleman:
> Whats the advantage of using 2 separate docker images?
I would prefer just having a single one for simplicity

So one image that contains two binaries? Or the hyperkube way of one binary that switches functionality based off of arg #1?

mrIncompetent:
> > Whats the advantage of using 2 separate docker images?
I would prefer just having a single one for simplicity

> So one image that contains two binaries? Or the hyperkube way of one binary that switches functionality based off of arg #1?

One image containing 2 binaries

alvaroaleman:
> One image containing 2 binaries

Done in b981212b7ee04f2fbedd8034f9e911765f963b32

PR #370, opened 2018-10-22
Fix ini escaping

mrIncompetent:
Fix ini escaping by going back to simple text templates as the used lib was not compatible.

```release-note
NONE
```


cbeneke:
lgtm

pkg/ini/escape_test.go
```diff
@@ -0,0 +1,71 @@
...
+			t.Fatalf("failed to execute cloud config template: %v", err)
+		}
+
+		after := &testData{}
+		if err := gcfg.ReadStringInto(after, buf.String()); err != nil {
+			t.Logf("\n%s", after)
+			t.Fatalf("failed to load string into config object: %v", err)
+		}
+
+		if before.Global.Password != after.Global.Password {
```
alvaroaleman: I think we should print the password that couldn't be un-/marshaled here

mrIncompetent: Done with https://github.com/kubermatic/machine-controller/pull/370/commits/b4488f171ab724afd3f2cf40fce9b36e30fe5209

PR #371, opened 2018-10-22
Move defaulting and validation into webhook

alvaroaleman:

Fixes #

In a follow-up PR I will ensure that we only default+validate once and at the highest possible level (e.G. MachineDeployment instead of Machine)

```release-note
Defaulting of machines is now done via a webhook
```


pkg/controller/machine/machine.go
```diff
@@ -585,50 +527,11 @@ func (c *Controller) deleteNodeForMachine(machine *clusterv1alpha1.Machine) erro
...
-		c.validationCacheMutex.Unlock()
-	} else {
-		glog.V(6).Infof("Skipping validation as the machine was already successfully validated before")
-	}
 	providerInstance, err := prov.Get(machine)
 
 	// case 2: retrieving instance from provider was not successful
 	if err != nil {
-		//First invalidate the validation cache to make sure we run the validation on the next sync.
-		//This might happen in case the user invalidates his provider credentials...
```
alvaroaleman: Hm. The webhook can obviously not help if someone invalidates their credentials. Maybe we should keep the validation in the controller and just remove the defaulting?

pkg/admission/admission.go
```diff
@@ -3,19 +3,32 @@ package admission
...
+}
+
+var jsonPatch = admissionv1beta1.PatchTypeJSONPatch
+
+func New(listenAddress string, coreClient kubernetes.Interface) *http.Server {
 	m := http.NewServeMux()
-	m.HandleFunc("/machinedeployments", handleFuncFactory(mutateMachineDeployments))
+	ad := &admissionData{coreClient: coreClient}
+	m.HandleFunc("/machinedeployments", handleFuncFactory(ad.mutateMachineDeployments))
+	m.HandleFunc("/machines", handleFuncFactory(ad.mutateMachines))
```
mrIncompetent: Should we also cover `MachineSet`s ? (Not in this PR of course)

alvaroaleman: Yes, will add that in a follow-up

pkg/admission/machines.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		machine.Annotations[clusterv1alpha1conversions.TypeRevisionAnnotationName] = clusterv1alpha1conversions.TypeRevisionCurrentVersion
+	}
+
+	// Default name
+	if machine.Spec.Name == "" {
+		machine.Spec.Name = machine.Name
+	}
+
+	// Default and verify .Spec on CREATE only, its expensive and not required to do it on UPDATE
```
mrIncompetent: It could help/improve the UX on update - but to avoid unnecessary calls to the cloud provider, we could maintain a internal cache.

The cache could be: 
```go
cache := map[string]bool{}
key := string(machine.UID)+hash(resolved-providerConfig)
```

Though this is an improvement and can be done in a follow up.

alvaroaleman: The reality is that updating a `machine` has no effect unless you manually delete the instance at the cloud provider which is why I have added a check above that will deny any `UPDATE` request except for `.Spec.Name` because we can not default it.

This is why its not required to do it on `UPDATE`, I will add that info to the comment

alvaroaleman: Actually it is already in the comment, see the line below: `as we disallow .Spec changes anyways`

alvaroaleman:
@mrIncompetent all comments addressed, PTAL

PR #380, opened 2018-10-29
[AWS] always use latest AMI ID

### LINKED ISSUE #357: Use latest coreos image for AWS
Currently the machine controller has a static list with AMI's for the coreos images.
This list has never been updated and is not up to date - leading to all nodes to do the autoupdate right after initial boot.

To prevent that, the machine controller should lookup the latest AMI and use it.

mrIncompetent:
This PR will use dynamic loading of AMI ID's for the relevant operating system images, instead of relying on a static hard coded map.

Fixes #357

```release-note
AWS: Always use the latest AMI available, when none is specified
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -78,12 +78,13 @@ var (
 			owner: "595879546273",
 		},
 		providerconfig.OperatingSystemCentOS: {
-			description: "CentOS Linux 7 x86_64*",
+			description: "CentOS Linux 7 x86_64 HVM EBS*",
 			// The AWS marketplace ID from AWS
 			owner: "679593333241",
 		},
 		providerconfig.OperatingSystemUbuntu: {
-			description: "Canonical, Ubuntu, 18.04*",
+			// Be as precise as possible - otherwise we might get a nightly dev build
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -70,62 +71,22 @@ var (
...
-			"us-west-1":      "ami-65e0e305",
-			"ap-northeast-2": "ami-7248e81c",
-			"ap-southeast-2": "ami-b6bb47d4",
-			"eu-west-2":      "ami-ee6a718a",
-			"us-east-2":      "ami-e1496384",
-			"us-west-2":      "ami-a042f4d8",
-			"eu-west-3":      "ami-bfff49c2",
+			description: "CentOS Linux 7 x86_64 HVM EBS*",
+			// The AWS marketplace ID from AWS
+			owner: "679593333241",
```
alvaroaleman: The owner is always the same in all regions?

mrIncompetent: Yes. Checked with the regions `eu-central-1` & `eu-west-1`

mrIncompetent: Wait, that is not correct

PR #382, opened 2018-10-30
Install docker 17.03 via binary and configure logs rotation on all systems via daemon.json

### LINKED ISSUE #292: Limit size of logs coming from docker & kubelet

mrIncompetent:
This PR will unify the docker installation on all operating systems.
Instead of installing the best(Or close to best) version via the individual OS repositories, docker will now be installed by simply downloading the release binary & creating the systemdunit.
This ensures that we always install the same version across all operating systems.

Additionally - and this was actually the initial idea - all operating systems now use the docker log-rotation via `daemon.json`

Fixes #292 

```release-note
Install docker 17.03 via binary and configure logs rotation on all systems via daemon.json
```

### DIFF
--- a/pkg/userdata/centos/testdata/kubelet-v1.10-aws.golden
+++ b/pkg/userdata/centos/testdata/kubelet-v1.10-aws.golden
@@ -57,8 +57,7 @@ write_files:
 
     
 
-    yum install -y docker-1.13.1 \
-      ebtables \
+    yum install -y ebtables \
       ethtool \
       nfs-utils \
       bash-completion \
@@ -66,6 +65,9 @@ write_files:
       socat \
       wget \
       curl \
+      libtool-ltdl \
+      libseccomp \
+      libcgroup \
       ipvsadm
 
     #setup some common directories
@@ -74,6 +76,16 @@ write_files:
     mkdir -p /etc/kubernetes/manifests
     mkdir -p /etc/cni/net.d
     mkdir -p /opt/cni/bin
+    
+    # docker
+    if [ ! -f /opt/bin/docker ]; then
+        # TODO: Support newer versions. Make sure to validate if the newer containerd version has configuration changes
+        # Newer versions of docker use containterd with a config file - containerd from 17.03 does not support that
+        # we maybe need to manage both ways(flags & config) or wait until we deprecate Kubernetes 1.11 
+        curl -L http://download.docker.com/linux/static/stable/x86_64/docker-17.03.2-ce.tgz | tar -xvzC /opt/ -f -
+        mv /opt/docker/* /opt/bin/
+        rm -rf /opt/docker
+    fi
     
     # cni
     if [ ! -f /opt/cni/bin/loopback ]; then
@@ -89,10 +101,15 @@ write_files:
         curl -Lfo /opt/bin/health-monitor.sh https://raw.githubusercontent.com/kubermatic/machine-controller/8b5b66e4910a6228dfaecccaa0a3b05ec4902f8e/pkg/userdata/scripts/health-monitor.sh
         chmod +x /opt/bin/health-monitor.sh
     fi
+    
+
+        # Make sure systemd is aware of the new units in /etc/systemd/system
+    systemctl daemon-reload
     systemctl enable --now docker
     systemctl enable --now kubelet
     systemctl enable --now --no-block kubelet-healthcheck.service
     systemctl enable --now --no-block docker-healthcheck.service
+    
 
 - path: "/opt/bin/supervise.sh"
   permissions: "0755"
@@ -113,11 +130,10 @@ write_files:
     Documentation=https://kubernetes.io/docs/home/
     
     [Service]
+    Environment="PATH=/opt/bin:/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin/"
     Restart=always
     StartLimitInterval=0
     RestartSec=10
-    
-    Environment="PATH=/opt/bin:/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin/"
     
     ExecStart=/opt/bin/kubelet $KUBELET_EXTRA_ARGS \
       --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf \
@@ -145,11 +161,6 @@ write_files:
     
     [Install]
     WantedBy=multi-user.target
-
-- path: "/etc/systemd/system/kubelet.service.d/extras.conf"
-  content: |
-    [Service]
-    Environment="KUBELET_EXTRA_ARGS=--cgroup-driver=systemd"
 
 - path: "/etc/kubernetes/cloud-config"
   content: |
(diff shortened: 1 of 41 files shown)


pkg/userdata/ubuntu/testdata/dist-upgrade-on-boot.golden
```diff
@@ -306,12 +306,17 @@ write_files:
...
   content: |
-    [Service]
-    ExecStart=
-    ExecStart=/usr/bin/dockerd -H fd:// --storage-driver=overlay2
+    {
+      "storage-driver": "overlay2",
+      "log-driver": "json-file",
+      "log-opts": {
+        "max-size": "10m",
+        "max-file": "2"
```
alvaroaleman: Is this a per-container setting?

alvaroaleman:
Is there a way we can find out via the Kube api which storage driver docker uses?

pkg/userdata/ubuntu/testdata/dist-upgrade-on-boot.golden
mrIncompetent: Unfortunately yes.

mrIncompetent:
> Is there a way we can find out via the Kube API which storage driver docker uses?

I don't think so. The Node Object does only container the container runtime version. The storage driver seems more an implementation detail, which should not even appear in the Node object.

pkg/userdata/ubuntu/testdata/dist-upgrade-on-boot.golden
alvaroaleman: Do we already limit the overall journald size?

mrIncompetent: Yes, on every system we set:
https://github.com/kubermatic/machine-controller/pull/382/files#diff-8e277877f731647a2f08ec1c985fb9abR16

alvaroaleman:
/hold

@mrIncompetent The issue with reading logs from the journald log driver only exists with the docker package from the built-in repo (`docker.io`), not with the one from the upstream repo (`docker-ce`) - Maybe just start using `docker-ce` for all Kube versions and not only 1.12?

alvaroaleman:
> Kubernetes 1.9 - 1.11 supports 17.03 as highest version.

Sort-Of, 17.03 is the highest _verified_ Docker version, that doesn't mean higher Docker versions won't work. E2Es do pass for all Kubernetes versions with Docker 1806

mrIncompetent:
Maybe going with the kops way of doing it, could be a solution?
https://github.com/kubernetes/kops/blob/master/nodeup/pkg/model/docker.go#L426

Basically downloading the release & manually setting it up.

It would decouple us from the operating system specific package management - which has up & downsides.
But it would enable us to use the latest, supported docker version available

alvaroaleman:
Doesn't help here, the problem is that we need a dynamically linked version of Docker in order for docker logs to work with the journald driver. So we basically must use packages from the repo

mrIncompetent:
@alvaroaleman PTAL

mrIncompetent:
This works finally. 
Went a non-pretty shortcut with that setup.sh script for CoreOS, but that did it :)

PTAL @alvaroaleman

pkg/userdata/ubuntu/userdata.go
```diff
@@ -133,6 +133,9 @@ hostname: {{ .MachineSpec.Name }}
 # Never set the hostname on AWS nodes. Kubernetes(kube-proxy) requires the hostname to be the private dns name
 {{ end }}
 
+groups:
```
alvaroaleman: What for do we need this?

pkg/userdata/centos/testdata/kubelet-v1.10-aws.golden
```diff
@@ -244,11 +260,95 @@ write_files:
...
+    KillMode=process
+    # restart the docker process if it exits prematurely
+    Restart=on-failure
+    StartLimitBurst=3
+    StartLimitInterval=60s
+    
+    [Install]
+    WantedBy=multi-user.target
+
+- path: /etc/systemd/system/docker.socket
```
alvaroaleman: Is there any reason to have a `docker.socket` and a `containerd.service` aside from "CoreOS does it that way"? The docker package on Ubuntu manages perfectly fine without a `containerd.service` and the socket is only useful to have socket activation which we don't need because we already explicitly declare `docker` as dependency of the Kubelet

pkg/userdata/ubuntu/userdata.go
mrIncompetent: I would like to stick as close as possible to "normal" setup of docker as possible.
Using a `docker` group is the common way of handling docker permission, no?

pkg/userdata/centos/testdata/kubelet-v1.10-aws.golden
mrIncompetent: `containerd.service`:
Even though has a compiled in containerd, docker will look on the host for a `/usr/bin/containerd` first. If one is found, that binary will be started instead of the compiled in one.
On CoreOS `/usr/bin/containerd` exists - as it gets shipped with CoreOS, but does not work together with Docker 17.04. As Thus the way via the service is needed.

`docker.socket`
That is how CoreOS does it. I can remove it

pkg/userdata/ubuntu/userdata.go
mrIncompetent: Oh wait, i just realized that we don't actually use that group anywhere

pkg/userdata/centos/testdata/kubelet-v1.10-aws.golden
mrIncompetent: Done with https://github.com/kubermatic/machine-controller/pull/382/commits/ce9322d27ec5c9706903b536d7c67294a104cdde

pkg/userdata/ubuntu/userdata.go
mrIncompetent: Removed with https://github.com/kubermatic/machine-controller/pull/382/commits/b4f71d6db2a40962ef43b1169d630ae4c4e6ea6f

alvaroaleman: We did use it on the socket ;)

pkg/userdata/centos/testdata/kubelet-v1.10-aws.golden
alvaroaleman: > containerd.service:
Even though has a compiled in containerd, docker will look on the host for a /usr/bin/containerd first. If one is found, that binary will be started instead of the compiled in one.

Yeah but that would mean that we need it on coreos only, not on the rest. I'd really like to prefer all of this as simple as possible and right now IMHO the templates are extremely hard to read already, so I'd prefer to not put anything in them which we dont absolutely need

pkg/userdata/ubuntu/userdata.go
mrIncompetent: Oh that explains a lot :D

pkg/userdata/centos/testdata/kubelet-v1.10-aws.golden
alvaroaleman: We can use a drop-in to add the additional dependency without having to alter the main unit file

mrIncompetent: Great idea, i'll adapt the pull!

mrIncompetent:
@alvaroaleman PTAL

PR #385, opened 2018-10-30
replace delete with cleanup to ensure proper deletion of all cloud provider resources

### LINKED ISSUE #383: [Azure] orphaning resources
When a machine gets deleted the controller will:
- Call the provider to check if an instance exists
- If an instance exists, the `Delete` function on the provider will be executed

The azure instance deletion consists of several cleanups:
- Instance
- Disk
- Network interfaces
- Public IP

When the first cleanup (instance) succeeds, but one of the following tasks fails, the machine will hang forever. The reason here is that the controller will check if an instance still exists - which won't return an instance. Thus the controller won't call the `Delete` on the provider implementation.

Potential solution:
Instead of relying on a successful previous `Get` on the provider implementation, the controller will only call `Delete` on the provider. 
For better understanding, the `Delete` function will be renamed to `Cleanup`

```go
    // Cleanup will delete the instance associated with the machine and all associated resources.
    // If all resources have been cleaned up, true will be returned.
    // In case the cleanup involves ansynchronous deletion of resources & those resources are not gone yet,
    // false should be returned. This is to indicate that the cleanup is not done, but needs to be called again at a later point  
    Cleanup(machine *clusterv1alpha1.Machine, update MachineUpdater) (bool, error)
```

mrIncompetent:
This PR changes the way we delete cloud provider resources.
Instead of doing a one-shot delete call, the controller will now continuously call `Cleanup` on the cloud provider implementation until it returns `true` - stating that all resources are gone.

Fixes #383

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/azure/provider.go
+++ b/pkg/cloudprovider/provider/azure/provider.go
@@ -467,10 +467,10 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloud.MachineCreateDe
 	return &azureVM{vm: &vm, ipAddresses: ipAddresses, status: status}, nil
 }
 
-func (p *provider) Delete(machine *v1alpha1.Machine, data *cloud.MachineCreateDeleteData) error {
+func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloud.MachineCreateDeleteData) (bool, error) {
 	config, _, err := p.getConfig(machine.Spec.ProviderConfig)
 	if err != nil {
-		return fmt.Errorf("failed to parse MachineSpec: %v", err)
+		return false, fmt.Errorf("failed to parse MachineSpec: %v", err)
 	}
 
 	_, err = p.Get(machine)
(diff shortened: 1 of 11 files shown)


pkg/cloudprovider/validationwrapper.go
```diff
@@ -63,8 +63,8 @@ func (w *cachingValidationWrapper) Create(m *v1alpha1.Machine, mcd *cloud.Machin
 }
 
 // Delete just calls the underlying cloudproviders Delete
-func (w *cachingValidationWrapper) Delete(m *v1alpha1.Machine, mcd *cloud.MachineCreateDeleteData) error {
-	return w.actualProvider.Delete(m, mcd)
+func (w *cachingValidationWrapper) Cleanup(m *v1alpha1.Machine, mcd *cloud.MachineCreateDeleteData) (bool, error) {
```
alvaroaleman: Comment should be `Cleanup just calls...`

test/e2e/provisioning/migrateuidscenario.go
```diff
@@ -145,15 +145,22 @@ func verifyMigrateUID(kubeConfig, manifestPath string, parameters []string, time
...
 			if i < maxTries-1 {
 				glog.V(4).Infof("Failed to delete machine %s on try %v with err=%v, will retry", machine.Name, i, err)
 				time.Sleep(10 * time.Second)
 				continue
 			}
 			return fmt.Errorf("failed to delete machine %s: %v", machine.Name, err)
 		}
+		if !done {
+			// Resources are not gone yet, check in a few seconds again
+			time.Sleep(30 * time.Second)
```
alvaroaleman: `30 seconds == a few seconds`..?

pkg/cloudprovider/validationwrapper.go
mrIncompetent: Done with https://github.com/kubermatic/machine-controller/pull/385/commits/90f5bfd702b525c9ccbf0da106bc5ae30a5c8a66

test/e2e/provisioning/migrateuidscenario.go
mrIncompetent: I set it to 10 seconds as it seems to be the common wait period in the cmd.
Also i updated the comment.

https://github.com/kubermatic/machine-controller/pull/385/commits/90f5bfd702b525c9ccbf0da106bc5ae30a5c8a66

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -479,47 +479,47 @@ func (p *provider) Delete(machine *v1alpha1.Machine, data *cloud.MachineCreateDe
...
 	}
 	if machine, err = data.Updater(machine, func(updatedMachine *v1alpha1.Machine) {
 		updatedMachine.Finalizers = kuberneteshelper.RemoveFinalizer(updatedMachine.Finalizers, finalizerPublicIP)
 	}); err != nil {
-		return err
+		return false, err
 	}
 
-	return nil
+	return true, nil
```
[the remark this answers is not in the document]
alvaroaleman: Its fine as it is

PR #391, opened 2018-11-05
AWS - Add cache for the AMI ID's

mrIncompetent:
This PR introduce a simple library which can be used for caching.
In this case, the AMI ID in the AWS provider gets cached.

The long term goal is to reduce the number of API calls to AWS.
It will speed up the machine processing.

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -145,16 +145,16 @@ func getDefaultAMIID(client *ec2.EC2, os providerconfig.OperatingSystem, region 
 	cacheLock.Lock()
 	defer cacheLock.Unlock()
 
+	filter, osSupported := amiFilters[os]
+	if !osSupported {
+		return "", fmt.Errorf("operating system %q not supported", os)
+	}
+
 	cacheKey := fmt.Sprintf("ami-id-%s-%s", region, os)
 	amiID, found := cache.Get(cacheKey)
 	if found {
 		glog.V(4).Info("found AMI-ID in cache!")
 		return amiID.(string), nil
-	}
-
-	filter, osSupported := amiFilters[os]
-	if !osSupported {
-		return "", fmt.Errorf("operating system %q not supported", os)
 	}
 
 	imagesOut, err := client.DescribeImages(&ec2.DescribeImagesInput{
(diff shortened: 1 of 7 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -89,6 +91,11 @@ var (
 			owner: "099720109477",
 		},
 	}
+
+	// cacheLock protects concurrent cache misses against a single key. This usually happens when multiple machines get created simultaneously
+	// We lock so the first access updates/writes the data to the cache and afterwards everyone reads the cached data
+	cacheLock = &sync.Mutex{}
+	cache     = gocache.New(5*time.Minute, 5*time.Minute)
```
alvaroaleman: Not sure about how I feel using a library for this. IMHO we should:

* Move the whole ami lookup into the `AddDefaults` to have it written into the `spec` to avoid the possibility that different machines with the same config from the same `MachineSet` end up using different AMIs 
* Add a validation step for the ami that checks if it exists into `Validate`
* Cache the validation for the whole `.Spec` by using a `map[string]bool` that holds a `hash` of the `.spec` and a `bool` that indicates success/failure

alvaroaleman: No, it covers all of that, `Machine` will still get defaulted and validated if its not owned by a `MachineSet`. My point is, that I would like to keep the amount of caches we built as low as possible, hence only one that validates the whole of a `MachineSpec`

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -134,7 +141,17 @@ type amiFilter struct {
...
+	defer cacheLock.Unlock()
+
+	cacheKey := fmt.Sprintf("ami-id-%s-%s", region, os)
+	amiID, found := cache.Get(cacheKey)
+	if found {
+		glog.V(4).Info("found AMI-ID in cache!")
+		return amiID.(string), nil
+	}
+
 	filter, osSupported := amiFilters[os]
```
alvaroaleman: I think the check if the os is supported should be the first one

mrIncompetent: Done with https://github.com/kubermatic/machine-controller/pull/391/commits/34fa383d134719058f7b065f840f75bcf4b9b43a

PR #399, opened 2018-11-09
Add keypair into openstack deploy param

jichenjc:
now something like:

keypair: "k1"

can be added into deploy yml of openstack machine

Add 'keypair' support to openstack nova provider
Fixes #

```release-note
now a keypair is eligible to be input to boot vm from openstack
```

### DIFF
--- a/examples/openstack-machinedeployment.yaml
+++ b/examples/openstack-machinedeployment.yaml
@@ -81,6 +81,8 @@ spec:
                 key: tenantName
             image: "Ubuntu 18.04 amd64"
             flavor: "m1.small"
+            # only required if there is need to inject keypair
+            keypair: ""
             securityGroups:
               - configMapKeyRef:
                   namespace: kube-system
--- a/pkg/cloudprovider/provider/openstack/helper.go
+++ b/pkg/cloudprovider/provider/openstack/helper.go
@@ -11,6 +11,7 @@ import (
 	"github.com/gophercloud/gophercloud"
 	goopenstack "github.com/gophercloud/gophercloud/openstack"
 	osavailabilityzones "github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
+	oskeypairs "github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/keypairs"
 	osflavors "github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
 	osimages "github.com/gophercloud/gophercloud/openstack/compute/v2/images"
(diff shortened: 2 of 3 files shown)


alvaroaleman:
Hi @jichenjc ,

Thanks for your PR! However it is already possible to inject a keypair by leveraging the `.spec.providerConfig.Value.sshPublicKeys` option: https://github.com/kubermatic/machine-controller/pull/399/files#diff-ade7da5603b0332c7452455f4f848522R48

jichenjc:
Yes, I noticed that, however, from openstack point of view, it's easier to use key pair concept even though you can use a ssh public key, but you need following to inject and use public directly actually is not supported in openstack API, from openstack API point of view, you need use keypair as it's transparent to user....

stack@k8s:~$ openstack keypair list
+------------+-------------------------------------------------+
| Name       | Fingerprint                                     |
+------------+-------------------------------------------------+
| clusterapi | 02:07:7d:b0:dd:68:80:d2:d1:4c:23:cc:3a:75:96:84 |
| k1         | 0d:dd:b2:81:3e:43:00:af:70:8b:a0:9f:ab:8e:0b:c8 |
+------------+-------------------------------------------------+
stack@k8s:~$ openstack keypair show --public-key k1
ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC/qHByJ/dg30q98qcVhKWlqakGDYtmzf7gKut8LMC2G935b1I6Fw89zEkxDcLE1xKV2LnAiO44EJ0e5/E9TCjgxNwPms7Akb38m7ZlBZREbgOqTDAlzUXMLeHSrj6lviHybCy84+eJL5f+H+/e9cIr63HWUVS9ddnrxvgQH/db6Tg62bfO6RmyY7C618vccz6Fd/M4CQo2shumKn32eKH+15/Kpf4tgmhro0NTqk7F01jmIqTp+y3DBFgxbl/fYtOlhoVKeGdibvNYgyApBFSHk5ZFf6PzAQTBUhxC14VI+wPQzN69exRZAJW+8iFbeLIJnL5If/iEIX1ypMCE6gmb Generated-by-Nova

alvaroaleman:
> Yes, I noticed that, however, from openstack point of view, it's easier to use key pair concept even though you can use a ssh public key, but you need following to inject and use public directly actually is not supported in openstack API, from openstack API point of view, you need use keypair as it's transparent to user....

Well, the reason we did it the way we did it is that its an implementation that works for _all_ supported providers. I think for end-users its a little confusing if there are two ways to pass in a ssh key

alvaroaleman:
Closing this as this hasn't seen any activity in the past four months. If you disagree, feel free to re-open.

/close

Issue #400, opened 2018-11-09
Move all existing machine-controller CI jobs to prow


alvaroaleman:
Currently the machine-controllers CI jobs mostly run in circleci. This has the drawback that its not properly integrated into drawback and with regards to the end to end test does not allow us to test external PRs for security reasons.

*Acceptance criteria:*

* All machine-controller CI jobs are executed by Prow

nikhita:
Closing as this is now done, yay!

PR #401, opened 2018-11-09
Update makefiles for migrating to prow

nikhita:
This PR updates Makefiles and adds a script for the e2e test so that we can directly use them in prowjobs.

This is a first step for moving from CircleCI to Prowjobs: https://github.com/kubermatic/machine-controller/issues/400.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ webhook: $(shell find cmd pkg -name '*.go') vendor
 		github.com/kubermatic/machine-controller/cmd/webhook
 
 lint:
+	./hack/verify-type-revision-annotation-const.sh
 	gometalinter --config gometalinter.json ./...
 
 docker-image: machine-controller webhook docker-image-nodep
@@ -105,3 +106,9 @@ deploy: examples/admission-cert.pem
 		|sed "s/__admission_cert__/$(shell cat examples/admission-cert.pem|base64 -w0)/g" \
 		|sed "s/__admission_key__/$(shell cat examples/admission-key.pem|base64 -w0)/g" \
 		|kubectl apply -f -
+
+check-dependencies:
+	# We need mercurial for bitbucket.org/ww/goautoneg, otherwise dep hangs forever
+	which hg >/dev/null 2>&1 || apt update && apt install -y mercurial
+	$$GOPATH/bin/dep version || (export DEP_RELEASE_TAG=v0.5.0; curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh)
+	dep status
--- a/hack/ci-e2e-test.sh
+++ b/hack/ci-e2e-test.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+set -e
+
+export BUILD_ID="${BUILD_ID:-$CIRCLE_BUILD_NUM}"
+
+# Install dependencies
+echo "Installing dependencies."
+apt update && apt install -y jq rsync unzip &&
+curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.10.0/bin/linux/amd64/kubectl &&
+chmod +x kubectl &&
+mv kubectl /usr/local/bin
+
+# Generate ssh keypair
+echo "Generating ssh keypairs."
+ssh-keygen -f $HOME/.ssh/id_rsa -P ''
+
+# Create environment at cloud provider
+echo "Creating environment at cloud provider."
+make -C test/tools/integration apply
+
+# Create kubeadm cluster and install machine-controller onto it
+echo "Creating kubeadm cluster and install machine-controller onto it."
+./test/tools/integration/provision_master.sh
+
+# Run e2e test
+echo "Running e2e test."
+export KUBECONFIG=$GOPATH/src/github.com/kubermatic/machine-controller/.kubeconfig &&
+go test -race -tags=e2e -parallel 240 -v -timeout 30m  ./test/e2e/... -identifier=$BUILD_ID
+
+function cleanup {
+    set +e
+
+    # Clean up machines
+    echo "Cleaning up machines."
+    ./test/tools/integration/cleanup_machines.sh
+
+    # Clean up master
+    echo "Clean up master."
+    make -C test/tools/integration destroy
+}
+trap cleanup EXIT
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -105,3 +105,12 @@ deploy: examples/admission-cert.pem
 		|sed "s/__admission_cert__/$(shell cat examples/admission-cert.pem|base64 -w0)/g" \
 		|sed "s/__admission_key__/$(shell cat examples/admission-key.pem|base64 -w0)/g" \
 		|kubectl apply -f -
+
+lint:
+	./hack/verify-type-revision-annotation-const.sh
+	gometalinter --config gometalinter.json ./...
+
+check-dependencies:
+	apt update && apt install -y mercurial
```
alvaroaleman: Can we add a check that first checks if the  binaries exist before trying to install them?

hack/e2e-test.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
```
alvaroaleman: Can we prefix the names of all scripts that are not intended to be used by developers but by CI only with `ci-`? So in this case that would mean move the script to `hack/ci-e2e-test.sh`

test/tools/integration/Makefile
```diff
@@ -1,4 +1,4 @@
-CIRCLE_BUILD_NUM ?= local
+CIRCLE_BUILD_NUM ?= local # TODO: what should this be for prow?
```
alvaroaleman: This must be unique per build, otherwise the e2e test runs into concurrency issue. Does prow provide something like a build id that we can leverage here? Worst case we could use nanosecond timestamp or something the like

Makefile
```diff
@@ -105,3 +105,12 @@ deploy: examples/admission-cert.pem
 		|sed "s/__admission_cert__/$(shell cat examples/admission-cert.pem|base64 -w0)/g" \
 		|sed "s/__admission_key__/$(shell cat examples/admission-key.pem|base64 -w0)/g" \
 		|kubectl apply -f -
+
+lint:
+	./hack/verify-type-revision-annotation-const.sh
```
alvaroaleman: There is btw already a lint target on line 39

nikhita: :woman_facepalming: thanks, will fix.

test/tools/integration/Makefile
```diff
@@ -1,3 +1,6 @@
+# prow uses BUILD_ID
+# TODO: remove CIRCLE_BUILD_NUM entirely after moving to prow
+CIRCLE_BUILD_NUM=$BUILD_ID
```
alvaroaleman: This needs a second `$` sign to be `$$BUILD_ID`, otherwise the value of `CIRCLE_BUILD_NUM` will be `UILD_ID`

nikhita: Done.

hack/ci-e2e-test.sh
```diff
@@ -0,0 +1,35 @@
...
+./test/tools/integration/provision_master.sh
+
+# Run e2e test
+echo "Running e2e test."
+export KUBECONFIG=$GOPATH/src/github.com/kubermatic/machine-controller/.kubeconfig &&
+go test -race -tags=e2e -parallel 240 -v -timeout 30m  ./test/e2e/... -identifier=$BUILD_ID
+
+# Clean up machines
+echo "Cleaning up machines."
+./test/tools/integration/cleanup_machines.sh
```
alvaroaleman: Hm, we need to put these two into an [exit trap](http://redsymbol.net/articles/bash-exit-traps/) to be sure they always get executed, even if the testing failed. Within that exit trap we should `set +e` so even if the `clean up machines` step fails, the `clean up master` step is executed

Makefile
```diff
@@ -105,3 +106,9 @@ deploy: examples/admission-cert.pem
 		|sed "s/__admission_cert__/$(shell cat examples/admission-cert.pem|base64 -w0)/g" \
 		|sed "s/__admission_key__/$(shell cat examples/admission-key.pem|base64 -w0)/g" \
 		|kubectl apply -f -
+
+check-dependencies:
+	# We need mercurial for bitbucket.org/ww/goautoneg, otherwise dep hangs forever
+	which hg >/dev/null 2>&1 || apt update && apt install -y mercurial
+	which dep >/dev/null 2>&1 || (export DEP_RELEASE_TAG=v0.5.0; curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh)
```
nikhita: @xmudrii wdyt about using this method for https://github.com/kubernetes-sigs/cluster-api-provider-digitalocean/pull/120#discussion_r232020500?

nikhita: Checked with Marko on Slack, we can't use this for `dep` because using `which dep` requires `dep` to be in the `$PATH`. `dep` is installed in `$GOPATH/bin` but it is not in `$PATH`, while using prow.

We can use this for `hg` because it is installed system-wide and is in `$PATH`.

Updated so that it can work with prow.

hack/ci-e2e-test.sh
nikhita: Done. Ptal, thanks!

test/tools/integration/Makefile
```diff
@@ -1,3 +1,6 @@
+# prow uses BUILD_ID
+# TODO: remove CIRCLE_BUILD_NUM entirely after moving to prow
+CIRCLE_BUILD_NUM=$$BUILD_ID
```
alvaroaleman: Right now this still must be `CIRCLE_BUILD_NUM ?=$$BUILD_ID` to only set if if `CIRCLE_BUILD_NUM` is unset

hack/ci-e2e-test.sh
```diff
@@ -0,0 +1,40 @@
...
+make -C test/tools/integration apply
+
+# Create kubeadm cluster and install machine-controller onto it
+echo "Creating kubeadm cluster and install machine-controller onto it."
+./test/tools/integration/provision_master.sh
+
+# Run e2e test
+echo "Running e2e test."
+export KUBECONFIG=$GOPATH/src/github.com/kubermatic/machine-controller/.kubeconfig &&
+go test -race -tags=e2e -parallel 240 -v -timeout 30m  ./test/e2e/... -identifier=$BUILD_ID
```
alvaroaleman: Until we are finished moving, this script needs a `export BUILD_NUM="${BUILD_NUM:-$CIRCLE_BUILD_NUM}"` to default `BUILD_ID` to `CIRCLE_BUILD_NUM`

nikhita: Done.

test/tools/integration/Makefile
nikhita: Done.

PR #406, opened 2018-11-13
Remove subnet concept in opensatck(Fix #404)

### LINKED ISSUE #404: Remove 'subnet' in openstack create
Don't know the purpose of 

type RawConfig struct
        Subnet           providerconfig.ConfigVarString   `json:"subnet"`

looks to me openstack doesn't support this ? we can specify network (already did) but why use subnet?
https://developer.openstack.org/api-ref/compute/?expanded=create-server-detail
seems doesn't support subnet as well...

just want to check whether some history here or I can remove them in a new PR

then we can fix https://github.com/kubermatic/machine-controller/issues/263 after the PR merged (it will be come much easier)

jichenjc:
Per Bug indicated, the subnet need to be provided inside
network through fixed ip param, not subnet...

Per bug indicated, the subnet concept is not suitable to openstack
Fixes # 404

```release-note
Subnet concept has been removed from openstack, follow up patch will reorg network into dict and make fixed_ip to replace subnet selection
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -53,7 +53,6 @@ type RawConfig struct {
 	Flavor           providerconfig.ConfigVarString   `json:"flavor"`
 	SecurityGroups   []providerconfig.ConfigVarString `json:"securityGroups"`
 	Network          providerconfig.ConfigVarString   `json:"network"`
-	Subnet           providerconfig.ConfigVarString   `json:"subnet"`
 	FloatingIPPool   providerconfig.ConfigVarString   `json:"floatingIpPool"`
 	AvailabilityZone providerconfig.ConfigVarString   `json:"availabilityZone"`
 	Region           providerconfig.ConfigVarString   `json:"region"`
@@ -74,7 +73,6 @@ type Config struct {
 	Flavor           string
 	SecurityGroups   []string
 	Network          string
-	Subnet           string
 	FloatingIPPool   string
 	AvailabilityZone string
 	Region           string
@@ -146,10 +144,6 @@ func (p *provider) getConfig(s v1alpha1.ProviderConfig) (*Config, *providerconfi
 		c.SecurityGroups = append(c.SecurityGroups, securityGroupValue)
 	}
 	c.Network, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Network)
-	if err != nil {
-		return nil, nil, nil, err
-	}
-	c.Subnet, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Subnet)
 	if err != nil {
 		return nil, nil, nil, err
 	}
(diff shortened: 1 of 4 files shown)


alvaroaleman:
Hi @jichenc, as pointed out in #404 I would prefer to figure out how to set the subnet on instance creation rather than removing the option and relying on the openstack-internal defaulting

alvaroaleman:
@jichenjc We've discussed this and it seems just removing the option is indeed the best way forward here

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -316,10 +290,6 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 		return fmt.Errorf("failed to get network %q: %v", c.Network, err)
 	}
 
-	if _, err := getSubnet(client, c.Region, c.Subnet); err != nil {
```
alvaroaleman: Could you replace this with a check that checks if the configured network as at least one subnet?

jichenjc:
can I know where to discuss this so later on I can submit questions there before open a bug.... maybe slack channel ?

alvaroaleman:
@jichenjc You can find me on the Kubernetes slack (@alvaroaleman)

pkg/cloudprovider/provider/openstack/provider.go
jichenjc: ok, I can do that

jichenjc: Another thought is we need defer those kind of logic into provider... 

I created a net without subnet and deploy will have following error: I tried in machine controller and didn't see error, however 

stack@k8s:~/devstack$ openstack server create --key-name k1 --image 55904f0b-ff9e-4b4c-91d5-0b98a4713e32 --flavor 1 --nic net-id=ad8b1a43-27eb-498a-b0db-5ce271e1050f --security-group group1 jji1
Network ad8b1a43-27eb-498a-b0db-5ce271e1050f requires a subnet in order to boot instances on. (HTTP 400) (Request-ID: req-c0c15150-5c11-4166-b0a1-29e0ffe71ca5)


  Type     Reason                Age                From                Message
  ----     ------                ----               ----                -------
  Warning  CreateInstanceFailed  47s (x3 over 67s)  machine-controller  Instance creation failed: failed to create server, due to Invalid request due to incorrect syntax or missing required parameters.
  Warning  CreateError           47s (x3 over 67s)  machine-controller  failed to create server, due to Invalid request due to incorrect syntax or missing required parameters.. Unable to create a machine.
root@kvm-017407:~/go/src/github.com/kubermatic/machine-controller# kubectl get machine -n kube-system
NAME                                           PROVIDER    OS       ADDRESS
openstack-machinedeployment-5df56b457c-stslv   openstack   ubuntu


so my thinking is
1) let's defer the creation error detect for complicate logics, this kind of logic check will be hard to maintain in k8s layer, use openstack and get its output then handle it will make sense (grammar and whether something is existing check is another story) 

2) when error occurs, get machine should output error info...

root@kvm-017407:~/go/src/github.com/kubermatic/machine-controller# kubectl get machine -n kube-system
NAME                                           PROVIDER    OS       ADDRESS
openstack-machinedeployment-5df56b457c-stslv   openstack   ubuntu

jichenjc: ok, I figured it out, need update CRD xml file...

alvaroaleman:
@jichenjc Will you update this PR with a check if the configured Opestack network has at least one subnet?

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -261,24 +255,13 @@ func (p *provider) AddDefaults(spec v1alpha1.MachineSpec) (v1alpha1.MachineSpec,
...
-		subnet, err := getDefaultSubnet(client, net, c.Region)
-		if err != nil {
-			return spec, osErrorToTerminalError(err, "error defaulting subnet")
-		}
-		if subnet != nil {
-			glog.V(4).Infof("Defaulted subnet for machine '%s' to '%s'", spec.Name, *subnet)
-			rawConfig.Subnet.Value = *subnet
-		}
+	// Try to get subnets, ensure at least one subnet exists
+	subnets, err := getSubnets(client, c.Region, rawConfig.Network.Value)
```
alvaroaleman: This is now in the `AddDefaults`, instead it should be in the `Validate` below

alvaroaleman:
Closing since there hasn't been any response within the past six months. Feel free to re-open if you want to pursue the topic
/close

PR #410, opened 2018-11-15
Add fmt check

jichenjc:
command `make fmt` can be used to do format check now

add format check 
we can use ``make fmt`` to verify code format now
Fixes #

```release-note
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ webhook: $(shell find cmd pkg -name '*.go') vendor
 lint:
 	./hack/verify-type-revision-annotation-const.sh
 	gometalinter --config gometalinter.json ./...
+	hack/verify-gofmt.sh
 
 docker-image: machine-controller webhook docker-image-nodep
 
--- a/hack/verify-gofmt.sh
+++ b/hack/verify-gofmt.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


hack/verify-gofmt.sh
```diff
@@ -0,0 +1,32 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if ! which gofmt > /dev/null; then
+  echo "Can not find gofmt"
+  exit 1
+fi
+
+diff=$(find . -name "*.go" | grep -v "\/vendor\/" | xargs gofmt -s -d 2>&1)
```
nikhita: This needs a `|| true` at the end.

Ref: https://github.com/kubernetes/kubernetes/blob/843a67b215dcfa1781820a797c4c759e42e81d3a/hack/verify-gofmt.sh#L52-L56

jichenjc:
ok, updated the code

Makefile
```diff
@@ -112,3 +112,6 @@ check-dependencies:
 	which hg >/dev/null 2>&1 || apt update && apt install -y mercurial
 	$$GOPATH/bin/dep version || (export DEP_RELEASE_TAG=v0.5.0; curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh)
 	dep status
+
+fmt:
+	hack/verify-gofmt.sh
```
alvaroaleman: @jichenjc Can you please move that into the existing `lint` Maketarget?

nikhita: How about renaming the `lint` target to `verify`?

alvaroaleman: SGTM

alvaroaleman:
We added a `go fmt` check as part of #426, hence this PR is not needed anymore

PR #409, opened 2018-11-14
Use downstream clientset for cluster-api

alvaroaleman:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

```release-note
NONE
```

PR #415, opened 2018-11-23
Update cluster-api in order to get an up-to-date typed client

alvaroaleman:

Fixes #

Supersedes #409 , we don't need that know that there is a check that verifies typed client generation upstream

```release-note
NONE
```


nikhita:
I'm guessing the e2e test is not triggered because it's only triggered when `pull-machine-controller-test` changes from failure -> successful. If this is true, this can be thought of as a bug (will follow up upstream).

/test pull-machine-controller-e2e

alvaroaleman:
> I'm guessing the e2e test is not triggered because it's only triggered when pull-machine-controller-test changes from failure -> successful. If this is true, this can be thought of as a bug (will follow up upstream).

Ah that makes sense, thanks for triggering it! But I think upstream will probably not consider this as a bug, as say we shouldn't use `run_after_success` in the first place

nikhita:
>  But I think upstream will probably not consider this as a bug, as say we shouldn't use run_after_success in the first place

Then they need to add a warning to not use run_after_success ;)

PR #421, opened 2018-12-03
fixes the way we add security groups to an instance (aws)

p0lyn0mial:
**What this PR does / why we need it**: prevents  machine controller to enter a loop in which it fails to attach an instance to an empty security group and destroys the instance immediately after.

```
NONE
```


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -466,19 +466,21 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloud.MachineCreateDe
...
-		if err != nil {
-			return nil, awsErrorToTerminalError(err, fmt.Sprintf("failed to attach instance %s to security group id's %v & delete the created instance", aws.StringValue(runOut.Instances[0].InstanceId), config.SecurityGroupIDs))
+		if modifyInstanceErr != nil {
+			_, err := ec2Client.TerminateInstances(&ec2.TerminateInstancesInput{
+				InstanceIds: []*string{runOut.Instances[0].InstanceId},
+			})
+			if err != nil {
+				return nil, awsErrorToTerminalError(err, fmt.Sprintf("failed to attach instance %s to security group id's %v due to %v & delete the created instance", aws.StringValue(runOut.Instances[0].InstanceId), config.SecurityGroupIDs, modifyInstanceErr))
+			}
+			return nil, awsErrorToTerminalError(modifyInstanceErr, fmt.Sprintf("failed to attach instance %s to security group %v", aws.StringValue(runOut.Instances[0].InstanceId), config.SecurityGroupIDs))
```
p0lyn0mial: `err` was nil so `awsErrorToTerminalError` always returned `nil` error.

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -466,19 +466,21 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloud.MachineCreateDe
...
-	_, err = ec2Client.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{
-		InstanceId: runOut.Instances[0].InstanceId,
-		Groups:     aws.StringSlice(config.SecurityGroupIDs),
-	})
-	if err != nil {
-		_, err := ec2Client.TerminateInstances(&ec2.TerminateInstancesInput{
-			InstanceIds: []*string{runOut.Instances[0].InstanceId},
+	if len(config.SecurityGroupIDs) > 0 {
+		// Change to our security group
+		_, modifyInstanceErr := ec2Client.ModifyInstanceAttribute(&ec2.ModifyInstanceAttributeInput{
```
p0lyn0mial: previously `err` was hidden and not returned to a developer, see line 476

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -466,19 +466,21 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloud.MachineCreateDe
...
+			InstanceId: runOut.Instances[0].InstanceId,
+			Groups:     aws.StringSlice(config.SecurityGroupIDs),
 		})
-		if err != nil {
-			return nil, awsErrorToTerminalError(err, fmt.Sprintf("failed to attach instance %s to security group id's %v & delete the created instance", aws.StringValue(runOut.Instances[0].InstanceId), config.SecurityGroupIDs))
+		if modifyInstanceErr != nil {
+			_, err := ec2Client.TerminateInstances(&ec2.TerminateInstancesInput{
```
p0lyn0mial: it is not clear to my why do we have to delete an instance when adding a security group fails ?

@mrIncompetent said that the nodes cannot communicate if the are no security groups. If we will leave that code as is then perhaps we could generate an `event` of type warning ?

mrIncompetent: We don't have any reconciling otherwise. 
If it fails, the user would get an instance, which does not fit the spec (In this case it might cannot communicate to other instances - Breaking all pod/service communication inside the cluster).

pkg/cloudprovider/provider/aws/provider.go
alvaroaleman: The question I have here thought is if there is any situation were not having a security group makes sense. It means the instance wont accept any incoming traffic which means it is basically useless as a kubernetes node. Maybe we should instead add a check in `Validate` that there is at least one security group defined?

Or are there other ways in which sec groups can be attached to instances @mrIncompetent ?

pkg/cloudprovider/provider/aws/provider.go
p0lyn0mial: may i know why we don't have reconciliation ? it requires too much changes in the code ? it's not practical ?

alvaroaleman: Because most settings of a machine can not be changed after creation, e.G. distro, AMI, VPC. That is why the webhook currently forbids any change to a machine.Spec. However there are a few settings that can only be set _after_ the instance was created as this one which requires us to go this route.

pkg/cloudprovider/provider/aws/provider.go
p0lyn0mial: @mrIncompetent can you please answer the question :) ?

mrIncompetent: Adding a validation step:
```go
if len(config.SecurityGroupIDs) > 0 {
    return fmt.Errorf('no security groups specified')
}
```
should be the way to go

pkg/cloudprovider/provider/aws/provider.go
p0lyn0mial: right, thanks!

pkg/cloudprovider/provider/aws/provider.go
p0lyn0mial: alright, so having at least one security group is required. I'll update the code.

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -467,18 +469,18 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloud.MachineCreateDe
...
 		Groups:     aws.StringSlice(config.SecurityGroupIDs),
 	})
-	if err != nil {
+	if modifyInstanceErr != nil {
 		_, err := ec2Client.TerminateInstances(&ec2.TerminateInstancesInput{
 			InstanceIds: []*string{runOut.Instances[0].InstanceId},
 		})
 		if err != nil {
-			return nil, awsErrorToTerminalError(err, fmt.Sprintf("failed to attach instance %s to security group id's %v & delete the created instance", aws.StringValue(runOut.Instances[0].InstanceId), config.SecurityGroupIDs))
+			return nil, awsErrorToTerminalError(err, fmt.Sprintf("failed to attach instance %s to security group id's %v due to %v & delete the created instance", aws.StringValue(runOut.Instances[0].InstanceId), config.SecurityGroupIDs, modifyInstanceErr))
```
alvaroaleman: Could you reword this to `failed to delete instance %s due to %v after attaching to security groups %v failed due to %v`? 
Right now I find the error message a little confusing

p0lyn0mial: how about  `failed to delete instance %s due to %v after attaching to security groups %v which failed due to %v` ?

alvaroaleman: SGTM

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -467,18 +469,18 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloud.MachineCreateDe
...
 		Groups:     aws.StringSlice(config.SecurityGroupIDs),
 	})
-	if err != nil {
+	if modifyInstanceErr != nil {
 		_, err := ec2Client.TerminateInstances(&ec2.TerminateInstancesInput{
 			InstanceIds: []*string{runOut.Instances[0].InstanceId},
 		})
 		if err != nil {
-			return nil, awsErrorToTerminalError(err, fmt.Sprintf("failed to attach instance %s to security group id's %v & delete the created instance", aws.StringValue(runOut.Instances[0].InstanceId), config.SecurityGroupIDs))
+			return nil, awsErrorToTerminalError(modifyInstanceErr, fmt.Sprintf("failed to delete instance %s due to %v after attaching to security groups %v", aws.StringValue(runOut.Instances[0].InstanceId), err, config.SecurityGroupIDs))
```
alvaroaleman: @p0lyn0mial It looks like you missed the last four words here

p0lyn0mial: the error will be formatted inside the function `awsErrorToTerminalError`. The documentation of that function says that but since you asked we might consider changing the name in the future or change the logic ?

alvaroaleman: ah I see, thanks for the hint. I think the logic is fine. Do you have an idea for a name that makes that behaviour more obvious?

p0lyn0mial: the one that says it formats the error, for example `awsErrorToTerminalErrorAndFormat` ?

PR #432, opened 2018-12-14
Make ValidateMachineDeployment public

thetechnick:
This PR makes the ValidateMachineDeployment public, so it can be used in other projects when importing the cluster API objects from the machine controller.

Fixes #

```release-note
NONE
```


alvaroaleman:
@thetechnick looks like there is an actual issue with this PR as the webhook is unavailable, maybe because it crashloops?

PR #444, opened 2019-01-18
Expose an instance count metric

alvaroaleman:

Fixes #

```release-note
none
```

/assign @mrIncompetent


pkg/cloudprovider/cloud/provider.go
```diff
@@ -45,11 +47,21 @@ type Provider interface {
...
 	MigrateUID(machine *clusterv1alpha1.Machine, new types.UID) error
+
+	// SetInstanceNumberForMachines purpose is to set the number of cloud provider instances that exist
+	// for each machine in the provided *GaugeVec, using "namespace/name" of the machine as identifying label value
+	// This func gets called regularly with all machines. It should be implemented in a way that makes as
+	// little API calls as possible, e.G. get all instances from the cloud provider in one call instead of
+	// doing one call per instance. It is assumed that the same set of credentials is used for all machines
+	// Providers which are implemented in a way that its not possible that there is more than one instance
+	// for a machine should just return
+	SetInstanceNumberForMachines(machines clusterv1alpha1.MachineList, metrics *prometheus.GaugeVec) error
```
alvaroaleman: What is outlined here is one possible approach that is implemented as a POC for AWS only. I already created the PR because I want to verify this is the approach we want to take before implementing this for all providers.

Concerns:
* Will we have to extend the `Provider` interface for every new metric that we add?
* For an implementer, these parts of the interface are not very easy to understand

Also: is passing in a raw `*prometheus.GaugeVec` suffient or does it make sense to pass in some interface that has a `func SetCountForMachine(machine, count)` to avoid providers setting a wrong label value?

pkg/controller/machine/machine.go
```diff
@@ -152,6 +153,45 @@ func NewMachineController(
...
+			}
+			var machineList clusterv1alpha1.MachineList
+			for _, machine := range machines {
+				machineList.Items = append(machineList.Items, *machine.DeepCopy())
+			}
+			if len(machineList.Items) < 1 {
+				return
+			}
+
+			// We assume that all machines are on the same provider
```
alvaroaleman: is this assumption appropriate? The reason for doing it like that is

* Getting the provider for a machine is rather cumbersome
* We would otherwise need to maintain a list per provider, which makes the provider abstraction feel leaky, ideally the controller wouldn't have to know anything about what providers exist

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -766,3 +767,54 @@ func setProviderConfig(rawConfig RawConfig, s v1alpha1.ProviderConfig) (*runtime
 
 	return &runtime.RawExtension{Raw: rawPconfig}, nil
 }
+
+func (p *provider) SetInstanceNumberForMachines(machines v1alpha1.MachineList, metrics *prometheus.GaugeVec) error {
+	if len(machines.Items) < 1 {
+		return nil
+	}
+	config, _, _, err := p.getConfig(machines.Items[0].Spec.ProviderConfig)
```
alvaroaleman: We assume that the credentials are identical for all machines. This means that rotating the credential will break this metric until the machine at position zero in the slice has the appropriate one.

The alternative to that would be to maintain a `map[credentialsHash]ec2Client` inside this func. Do you think that should be done?

alvaroaleman:
Result looks like this for AWS:

```
$ curl -Ss localhost:8085/metrics|grep machine_controller_instances_for_machine
# HELP machine_controller_instances_for_machine The number of cloud provider instances for a given machine
# TYPE machine_controller_instances_for_machine gauge
machine_controller_instances_for_machine{machine="kube-system/alvaro-test-7b546d8db5-rd8rq"} 1
```

pkg/cloudprovider/cloud/provider.go
mrIncompetent: We could call the cloud provider from the metrics collector: https://github.com/kubermatic/machine-controller/blob/master/pkg/controller/machine/metrics.go#L122

That way the cloud provider implementation could define the metrics "on the fly" and we don't have to dictate every implementation what metrics they need/can to use.

As the collector gets called on each scrape, we might wan't to introduce some "buffer" around that - as the `Collection` function should respond ASAP

pkg/controller/machine/machine.go
alvaroaleman: Done

pkg/cloudprovider/provider/aws/provider.go
alvaroaleman: I have implemented that as a `map[identifier]credentials`

alvaroaleman:
Looks like this now as the metric got a new name:

```
$ curl -Ss localhost:8085/metrics|grep machine_controller_aws_instances_for_machine
# HELP machine_controller_aws_instances_for_machine The number of instances at aws for a given machine
# TYPE machine_controller_aws_instances_for_machine gauge
machine_controller_aws_instances_for_machine{machine="kube-system/alvaro-test-7b546d8db5-4dpmk"}
```

With this change it is easy for all providers to register their own metrics

alvaroaleman:
PTAL @mrIncompetent

pkg/cloudprovider/provider/fake/provider.go
```diff
@@ -91,3 +91,7 @@ func (p *provider) MigrateUID(machine *v1alpha1.Machine, new types.UID) error {
 func (p *provider) MachineMetricsLabels(machine *v1alpha1.Machine) (map[string]string, error) {
 	return map[string]string{}, nil
 }
+
+func (p *provider) SetMetricsForMachines(_ v1alpha1.MachineList) error {
+	return fmt.Errorf("Not implemented")
```
mrIncompetent: Why return for some implementations a `Not implemented` errors and for some simply nil?
https://github.com/kubermatic/machine-controller/pull/444/files#diff-42c33a2522ded90fccbad92e902b319cR358

I would prefer going the `nil` way as metrics are optional and should not cause any errors

pkg/controller/machine/metrics.go
```diff
@@ -89,6 +94,56 @@ func (l *machineMetricLabels) Counter(value uint) prometheus.Counter {
 }
 
 func NewMachineCollector(lister v1alpha1.MachineLister, kubeClient kubernetes.Interface) *MachineCollector {
+
+	// Start periodically calling the providers SetMetricsForMachines in a dedicated go routine
+	skg := providerconfig.NewConfigVarResolver(kubeClient)
+	go func() {
+		metricGatheringExecutor := func() {
+			machines, err := lister.List(labels.NewSelector())
```
mrIncompetent: Use `labels.Everything()` instead for readability

PR #448, opened 2019-01-28
Update prometheus client dependency

alvaroaleman:

Fixes #

```release-note
none
```


pkg/controller/machine/metrics.go
```diff
@@ -88,7 +88,7 @@ func (l *machineMetricLabels) Counter(value uint) prometheus.Counter {
 	}, labelNames)
 
 	counter := counterVec.With(labels)
-	counter.Set(float64(value))
+	counter.Add(float64(value))
```
alvaroaleman: The current prometheus client lib doesn't have `.Set` for counters anymore. We can just use the `.Add` thought, as we declare it right before

PR #449, opened 2019-02-08
Add option to request spot instances on AWS

alvaroaleman:

Fixes #

```release-note
Add an option to create spot instances instead of normal instances on AWS
```

/assign @mrIncompetent


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -459,8 +462,14 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloud.MachineCreateDe
 		})
 	}
 
+	var instanceMarketOptions *ec2.InstanceMarketOptionsRequest
+	if config.IsSpotInstance != nil && *config.IsSpotInstance {
+		instanceMarketOptions = &ec2.InstanceMarketOptionsRequest{MarketType: aws.String("spot")}
```
mrIncompetent: nit: Theres a const for "spot" in the ec2 package: 
```go 
const (
	// MarketTypeSpot is a MarketType enum value
	MarketTypeSpot = "spot"
)
```

Do we also want to allow for configuration of `BlockDurationMinutes` ?

alvaroaleman: SGTM and I was grepping for `Spot` not for `spot` as that is what they used in their example: https://aws.amazon.com/blogs/compute/new-amazon-ec2-spot-pricing/

> Do we also want to allow for configuration of BlockDurationMinutes ?

I guess thats reasonable. If and where we actually usage that option later on is not something we have to settle on right now

mrIncompetent:
I've introduced the const for "spot" and rebased.

/lgtm
/approve

/hold
@alvaroaleman any objections?

alvaroaleman:
No

/hold cancel

mrIncompetent:
Still the hetzner apt mirror issues

PR #451, opened 2019-02-10
Adds a machine driver for Linode

### LINKED ISSUE #450: Linode provider support
Add a machine driver for https://www.linode.com/  

They've been around for 15 years and aren't going anywhere.  

(I work at Linode)

displague:

This PR adds a machine driver Linode instances.

Fixes #450

This is a first pass at a machine driver for Linode, based on the DO and Hetzner drivers.
I haven't given it a run yet, but it does compile, quite promising.

```release-note
* Adds the Linode machine driver
```


pkg/cloudprovider/provider/linode/provider.go
```diff
@@ -0,0 +1,447 @@
...
+		},
+	}
+
+	linode, err := client.CreateInstance(ctx, createRequest)
+	if err != nil {
+		return nil, doStatusAndErrToTerminalError(err)
+	}
+
+	//We need to wait until the linode really got created as tags will be only applied when the linode is running
+	err = wait.Poll(createCheckPeriod, createCheckTimeout, func() (done bool, err error) {
```
alvaroaleman: If I am not misunderstating the API docs, it should be possible to pass Tags on creation: https://developers.linode.com/api/v4#operation/createLinodeInstance

That would be much nicer because with the current approach we basically lose track of the instance if the controller crashes at the right point in time

pkg/cloudprovider/provider/linode/provider.go
```diff
@@ -0,0 +1,447 @@
...
+	for _, linode := range linodes {
+		if linode.Label == machine.Spec.Name && sets.NewString(linode.Tags...).Has(string(machine.UID)) {
+			updateOpts := linode.GetUpdateOptions()
+
+			tags := []string{string(new)}
+			if updateOpts.Tags != nil {
+				oldUID := string(machine.UID)
+				for _, existingTag := range *updateOpts.Tags {
+					if existingTag != oldUID {
+						tags = append(tags, existingTag)
```
alvaroaleman: Can we remove the old tag here?

pkg/cloudprovider/provider/linode/provider.go
```diff
@@ -0,0 +1,447 @@
...
+		return instance.StatusUnknown
+	}
+}
+
+// doStatusAndErrToTerminalError judges if the given HTTP status
+// can be qualified as a "terminal" error, for more info see v1alpha1.MachineStatus
+
+// if the given error doesn't qualify the error passed as
+// an argument will be returned
+func doStatusAndErrToTerminalError(err error) error {
```
alvaroaleman: Can we rename this to `linodeStatusAndErroToTerminalError`? :)

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -166,6 +167,24 @@ func TestHetznerProvisioningE2E(t *testing.T) {
...
 
+// TestLinodeProvisioning - a test suite that exercises Linode provider
+// by requesting nodes with different combination of container runtime type, container runtime version and the OS flavour.
+//
+// note that tests require  a valid API Token that is read from the LINODE_E2E_TEST_TOKEN environmental variable.
+func TestLinodeProvisioningE2E(t *testing.T) {
+	t.Parallel()
+
+	// test data
+	linodeToken := os.Getenv("LINODE_E2E_TESTS_TOKEN")
```
alvaroaleman: Could you add an `excludeSelector` so only Ubuntu is tested as that is the only supported OS?

https://github.com/kubermatic/machine-controller/blob/3f15ca3bcbb4c18c408be72c59f2bcd80483d81a/test/e2e/provisioning/all_e2e_test.go#L54

pkg/cloudprovider/provider/linode/provider.go
displague: Sure .. I wasn't sure if "do" here was a verb or a reference to DO.

pkg/cloudprovider/provider/linode/provider.go
displague: Unless I missed something, that's what I'm doing.

I start with `tags := []string{string(new)}` and just add back all of the other tags, except for the `oldUID` tag.

pkg/cloudprovider/provider/linode/provider.go
displague: That's what I get for borrowing from another provider. I'll take another look at this.

test/e2e/provisioning/all_e2e_test.go
displague: Will do.

pkg/cloudprovider/provider/linode/provider.go
alvaroaleman: Oh you are right, my bad

alvaroaleman:
Sorry for the test noise, it seems we have hit a quota: `resource_limit_exceeded` :disappointed: 

I have requested an increase of that quota. Also, once you consider the PR done you can manually trigger the linode e2e test by writing `/test pull-machine-controller-e2e-linode` on a distinct line

pkg/cloudprovider/provider/linode/provider.go
displague: UpdateInstance returns the full Linode instance object, I'll change this to not throw it away since it is used below.

I'll also need to change those status strings to match the Linode ones (there are consts to be used there).

pkg/cloudprovider/provider/linode/provider.go
displague: These status strings will be updated to reflect linodego constants.

displague: updated in 575403a

test/e2e/provisioning/all_e2e_test.go
displague: updated in [575403a](https://github.com/kubermatic/machine-controller/commit/575403ae478681a8e3395eeabbb8588c522309a5)

pkg/cloudprovider/provider/linode/provider.go
displague: I misread the context from my phone.. I updated the status strings.  There is no reason to keep the linode instance object returned from UpdateInstance, here.

pkg/cloudprovider/provider/linode/provider.go
displague: updated in [575403a](https://github.com/kubermatic/machine-controller/commit/575403ae478681a8e3395eeabbb8588c522309a5)

simplified the polling using a wait mechanism built into the linodego client.

pkg/cloudprovider/provider/linode/provider.go
```diff
@@ -0,0 +1,465 @@
...
+func New(configVarResolver *providerconfig.ConfigVarResolver) cloud.Provider {
+	return &provider{configVarResolver: configVarResolver}
+}
+
+type RawConfig struct {
+	Token             providerconfig.ConfigVarString   `json:"token"`
+	Region            providerconfig.ConfigVarString   `json:"region"`
+	Type              providerconfig.ConfigVarString   `json:"type"`
+	Backups           providerconfig.ConfigVarBool     `json:"backups"`
+	IPv6              providerconfig.ConfigVarBool     `json:"ipv6"`
```
alvaroaleman: The IPV6 option is not used anywhere or did I miss it?

pkg/apis/cluster/v1alpha1/conversions/testdata/machinesv1alpha1machine/linode.yaml
```diff
@@ -0,0 +1,26 @@
...
+  metadata:
+    name: node1
+  providerConfig:
+    sshPublicKeys:
+      - "<< YOUR_PUBLIC_KEY >>"
+    cloudProvider: "linode"
+    cloudProviderSpec:
+      token: "token"
+      region: eu-west
+      size: g6-standard-2
```
alvaroaleman: These fixtures are not really accurate, the `Linode` ProviderSpec doesn't have a `size` property. Maybe just remove them?

pkg/cloudprovider/provider/linode/provider.go
```diff
@@ -0,0 +1,465 @@
...
+}
+
+type Config struct {
+	Token             string
+	Region            string
+	Type              string
+	Backups           bool
+	IPv6              bool
+	PrivateNetworking bool
+	Monitoring        bool
```
alvaroaleman: The `Monitoring` option isn't used anywhere or did I miss it?

displague: Nice catch. Linode's has a few options for this, alerting on CPU,net,disk usage threaholds, auto rebooting on shutdown/panic (Lassie), and Longview (graphs and stats, process/package/service monitoring).  

For the sake of this PR, that field should be removed.

pkg/cloudprovider/provider/linode/provider.go
displague: Linodes always have IPv6 enabled at the service level. It can be disabled within the OS.

This field should be removed.

displague: Linodes always have IPv6 enabled at the service level. It can be disabled within the OS.

This field should be removed.

pkg/apis/cluster/v1alpha1/conversions/testdata/machinesv1alpha1machine/linode.yaml
displague: The Linode value that makes sense here is `type` (instance type), which I've paralleled to size in other projects.
g6-nanode-1 (1gb ram, 1vcpu, 25gb SSD)
g6-standard-2 (..)
...
g6-highmem-16 (...)
...
g6-dedicated-1 (...)
...

alvaroaleman:
/lgtm
/approve

Thanks a lot for this @displague !

PR #463, opened 2019-03-04
Add Google Cloud provider

themue:

Supporting of Google Cloud with the machine provider.

Fixes #

```release-note
Add support of Google Cloud as provider
```


pkg/cloudprovider/provider/gce/cloudconfig.go
```diff
@@ -0,0 +1,93 @@
+//
+// Google Cloud Provider for the Machine Controller
+//
+
+package gce
+
+//-----
+// Imports
+//-----
```
mrIncompetent: I'm not a big fan of such comments.
For me this is just noise and decreases readability as `import` keyword is expressive enough for me.
This applies for all "segment" comments in this PR - i'm just commenting on a single bit to have a single conversation.

But this is just my personal opinion.
@alvaroaleman Any opinion on this?

pkg/cloudprovider/provider/gce/cloudconfig.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/Masterminds/sprig"
+)
+
+//-----
+// Constants
+//-----
+
+// cloudConfigTemplate renders the cloud-config in gcfg format. All
+// fields are optional, that's why containing the ifs and the explicit newlines.
+// TODO(mue) Check for mandatory fields and better ways to drop empty fields,
```
mrIncompetent: AFAIK there is no field which is mandatory

pkg/cloudprovider/provider/gce/cloudconfig_test.go
```diff
@@ -0,0 +1,53 @@
...
+	}{
+		{
+			name: "minimum test",
+			config: &cloudConfig{
+				Global: global{
+					ProjectID: "my-project-id",
+					LocalZone: "my-zone",
+				},
+			},
+			contents: "[global]\n" +
```
mrIncompetent: Please use a multi line string using backticks. It improves readability

pkg/cloudprovider/provider/gce/service.go
```diff
@@ -0,0 +1,129 @@
...
+		return svc.RegionOperations.Get(projectID, region, op.Name).Do()
+	default:
+		return svc.GlobalOperations.Get(projectID, op.Name).Do()
+	}
+}
+
+// waitOperation waits for a GCE operation to be completed or timed out.
+func (svc *service) waitOperation(projectID string, op *compute.Operation, timeout time.Duration) (err error) {
+	started := time.Now()
+	waiting := 100 * time.Millisecond
```
mrIncompetent: Instead if building you own polling function i would recommend to use something which already exists:
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L306

pkg/cloudprovider/provider/gce/service.go
```diff
@@ -0,0 +1,129 @@
...
+	"google.golang.org/api/compute/v1"
+)
+
+//-----
+// Constants
+//-----
+
+const (
+	// Timeouts.
+	timeoutNormal = 10 * time.Minute
```
mrIncompetent: Having a global timeout of 10 minutes does not really look right for all API calls we do

pkg/cloudprovider/provider/gce/provider.go
```diff
@@ -0,0 +1,335 @@
...
+func New(configVarResolver *providerconfig.ConfigVarResolver) *Provider {
+	return &Provider{
+		resolver: configVarResolver,
+	}
+}
+
+// AddDefaults reads the MachineSpec and applies defaults for provider specific fields
+func (p *Provider) AddDefaults(spec v1alpha1.MachineSpec) (v1alpha1.MachineSpec, error) {
+	// TODO(mue) Check for defaults to apply.
+	return spec, nil
```
mrIncompetent: Set:
- disk to `pd-standard`
- diskSize to `25`

pkg/cloudprovider/provider/gce/cloudconfig.go
themue: Will reduce it down to the currently used ones. Also helps to drop these `if` statements in the template.

pkg/cloudprovider/provider/gce/cloudconfig_test.go
themue: Got a different opinion when the rest of the construct, like here, is nested and indented. But it's okay.

mrIncompetent: Good point, feel free to keep it like it is :)

pkg/cloudprovider/provider/gce/service.go
themue: Nice, did not know. Mine has been only stolen. 😄

pkg/cloudprovider/provider/gce/service.go
themue: Changed to interval of 5 seconds and timeout of 5 minutes as seen in other project.

pkg/cloudprovider/provider/gce/cloudconfig.go
themue: Done.

themue:
> Should the open TODO's in the code be handled later?
> If so, please create an issue & communicate it with @kdomanski

TODOs are done now.

examples/gce-machinedeployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+            zone: "europe-west3-a"
+            # See https://cloud.google.com/compute/docs/machine-types
+            machineType: "n1-standard-2"
+            # In GB
+            diskSize: 50
+            # Can be 'pd-standard' or 'pd-ssd'
+            diskType: "pd-standard"
+            labels:
+              "kubernetes_cluster": "my-cluster"
+          # Can be 'ubuntu', 'coreos' or 'centos'
```
mrIncompetent: Not correct anymore. CentOS is not supported

pkg/cloudprovider/provider/gce/provider.go
```diff
@@ -0,0 +1,335 @@
...
+		Disks:             disks,
+		Labels:            labels,
+		ServiceAccounts: []*compute.ServiceAccount{
+			{
+				Email: cfg.jwtConfig.Email,
+				Scopes: []string{
+					compute.ComputeScope,
+					compute.DevstorageReadOnlyScope,
+					"https://www.googleapis.com/auth/monitoring",
+					"https://www.googleapis.com/auth/logging.write",
```
mrIncompetent: Are there any consts for those?

examples/gce-machinedeployment.yaml
themue: Gnah, got me.

pkg/cloudprovider/provider/gce/provider.go
themue: Haven't found anywhere, not in `compute` nor elsewhere. But will look again.

mrIncompetent: Maybe you want to temporarily disable the vendor pruning, that helps sometimes finding such things

themue: Looks bad, only those few defined here: https://godoc.org/google.golang.org/api/compute/v1#pkg-constants.

themue: Found packages for logging with scope constants and `cloud.google.com/go/monitoring/apiv3`. The latter is only exporting `DefaultAuthScopes()` returning a slice of all scopes the package needs.

PR #465, opened 2019-03-06
Add AnnotationMachineUninitialized to allow external initialization

alvaroaleman:

Fixes #

```release-note
none
```


pkg/controller/machine/machine.go
```diff
@@ -64,6 +64,11 @@ const (
 	FinalizerDeleteInstance = "machine-delete-finalizer"
 	FinalizerDeleteNode     = "machine-node-delete-finalizer"
 
+	// AnnotationMachineUninitialized indicates that a machine is not yet
+	// ready to be worked on by the machine-controller. The machine-controller
+	// will ignore all machines that have this anotation with any value
+	AnnotationMachineUninitialized = "machine-controller.kubermatic.io/needs-initialization"
```
mrIncompetent: Can we rename this to `initializers` and use it as comma separated string?
Otherwise we only allow a single initialization :/

alvaroaleman: The value of this field can be a comma separated string. Its key is intended to indicate that this specific controller will not work on machines that have it. Just in case we ever have another controller that does things on machine objects

mrIncompetent: Vague assumption: This is a replacement for `initializers`.

If that is the case, we must rename this field. Currently its called `needs-initialization` which implies that this field is a `boolean`. Which means only a single consumer can use it. 

As we're the owner of that field, we must dictate the usage of it - otherwise implementations will use it the way they like and we risk conflicts.

Rename + Make it clear that this is a comma separated field.

alvaroaleman: Updated, PTAL

PR #466, opened 2019-03-12
e2e tests: respect local KUBECONFIG env before using hardcoded path

kdomanski:
Allows easily running the tests against any environment when running them locally. Previously, the hardcoded path required copying/linking the kubeconfig to `$(go env GOPATH)/src/github.com/kubermatic/machine-controller/.kubeconfig`

```release-note
NONE
```

### DIFF
--- a/test/e2e/provisioning/helper.go
+++ b/test/e2e/provisioning/helper.go
@@ -110,10 +110,13 @@ func testScenario(t *testing.T, testCase scenario, cloudProvider string, testPar
 	// only used by OpenStack scenarios
 	scenarioParams = append(scenarioParams, fmt.Sprintf("<< OS_IMAGE >>=%s", openStackImages[testCase.osName]))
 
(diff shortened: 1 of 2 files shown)


test/e2e/provisioning/helper.go
```diff
@@ -110,10 +110,13 @@ func testScenario(t *testing.T, testCase scenario, cloudProvider string, testPar
 	// only used by OpenStack scenarios
 	scenarioParams = append(scenarioParams, fmt.Sprintf("<< OS_IMAGE >>=%s", openStackImages[testCase.osName]))
 
-	gopath := os.Getenv("GOPATH")
```
mrIncompetent: This must be the other way around, first check the kubeconfig in the repo path, otherwise you might break the existing behavior.

Also please update the Readme

kdomanski: Done.

PR #470, opened 2019-03-21
--cloud-provider=external

### LINKED ISSUE #440: Support passing --cloud-provider=external to worker kubelets
It would be nice if machine-controller supports passing `--cloud-provider=external` flag to workers it creates. That flag configures kubelet to work with external Cloud Controller Manager (CCM) and is required for all nodes if external CCM is deployed.

One of decisions we have to make regarding this issue is how we want to expose this option. From the past discussions, we have the following options:
* Add a CLI flag
* Use an environment variable
* If it is known that cloud provider uses external CCM, then set `--cloud-provider=external` automatically
* Build a Webhook-based approach

In my opinion, the third approach (assuming external CCM) is bad because we don't know is master configured to work with CCM (as we don't manage masters, but that could be potentially checked), but also cloud providers with in-tree implementation have also an out-of-tree implementation.

**Acceptance criteria**:
* [ ] An option to pass `--cloud-provider=external` flag to kubelets is implemented

**Notes:**

Relevant documentation: [Kubernetes Cloud Controller Manager](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/)

As part of [KEP 0002](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cloud-provider/0002-cloud-controller-manager.md) and [KEP 0013](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cloud-provider/0013-build-deploy-ccm.md#migrating-to-a-ccm-build), all in-tree CCMs are deprecated and are switching to external CCMs, so setting this flag will be required in the future. More details can also be found in [the `pkg/cloudprovider` package](https://github.com/kubernetes/kubernetes/tree/master/pkg/cloudprovider).

kron4eg:

Fixes #440 

```release-note
new flag `-external-cloud-provider` which will initialize `kubelet --cloud-provider=external`
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -68,14 +68,15 @@ import (
 )
 
 var (
-	masterURL          string
-	kubeconfig         string
-	clusterDNSIPs      string
-	listenAddress      string
-	profiling          bool
-	name               string
-	joinClusterTimeout string
-	workerCount        int
+	masterURL             string
+	kubeconfig            string
+	clusterDNSIPs         string
+	listenAddress         string
+	profiling             bool
+	name                  string
+	joinClusterTimeout    string
+	workerCount           int
+	externalCloudProvider bool
 )
 
 const (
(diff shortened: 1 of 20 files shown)


kron4eg:
```* hcloud_ssh_key.default: hcloud_ssh_key.default: Get https://api.hetzner.cloud/v1/ssh_keys/521992: dial tcp: lookup api.hetzner.cloud on 10.96.0.10:53: server misbehaving```

/test pull-machine-controller-e2e-ubuntu-upgrade

PR #472, opened 2019-03-27
Support for network configuration of GCE instances

kron4eg:
In order to better support network placement of nodes

```release-note
GCE network placement controls
```

### DIFF
--- a/examples/gce-machinedeployment.yaml
+++ b/examples/gce-machinedeployment.yaml
@@ -49,7 +49,8 @@ spec:
             machineType: "n1-standard-2"
             # In GB
             diskSize: 50
-            # ether of two network/subnetwork could be specified
+            # The name or self_link of the network and subnetwork to attach this interface to;
+            # either of both can be provided, otherwise default network will taken
             # in case if both empty — default network will be used
             network: "my-cool-network"
             subnetwork: "my-cool-subnetwork"
--- a/examples/gce-machinedeployment.yaml
+++ b/examples/gce-machinedeployment.yaml
@@ -49,6 +49,10 @@ spec:
             machineType: "n1-standard-2"
             # In GB
             diskSize: 50
+            # ether of two network/subnetwork could be specified
+            # in case if both empty — default network will be used
+            network: "my-cool-network"
+            subnetwork: "my-cool-subnetwork"
             # Can be 'pd-standard' or 'pd-ssd'
             diskType: "pd-standard"
             labels:
--- a/pkg/cloudprovider/provider/gce/config.go
+++ b/pkg/cloudprovider/provider/gce/config.go
@@ -167,7 +167,7 @@ func newConfig(resolver *providerconfig.ConfigVarResolver, spec v1alpha1.Provide
 
 	cfg.subnetwork, err = resolver.GetConfigVarStringValue(cpSpec.Subnetwork)
 	if err != nil {
-		return nil, fmt.Errorf("cannot retrieve network: %v", err)
+		return nil, fmt.Errorf("cannot retrieve subnetwork: %v", err)
 	}
 
 	return cfg, nil
--- a/pkg/cloudprovider/provider/gce/config.go
+++ b/pkg/cloudprovider/provider/gce/config.go
@@ -133,6 +133,8 @@ func newConfig(resolver *providerconfig.ConfigVarResolver, spec v1alpha1.Provide
 		providerConfig: providerConfig,
 		labels:         cpSpec.Labels,
 		diskSize:       cpSpec.DiskSize,
+		network:        cpSpec.Network.Value,
+		subnetwork:     cpSpec.Subnetwork.Value,
 	}
 
 	cfg.serviceAccount, err = resolver.GetConfigVarStringValueOrEnv(cpSpec.ServiceAccount, envGoogleServiceAccount)
--- a/pkg/cloudprovider/provider/gce/config.go
+++ b/pkg/cloudprovider/provider/gce/config.go
@@ -133,8 +133,6 @@ func newConfig(resolver *providerconfig.ConfigVarResolver, spec v1alpha1.Provide
 		providerConfig: providerConfig,
 		labels:         cpSpec.Labels,
 		diskSize:       cpSpec.DiskSize,
-		network:        cpSpec.Network.Value,
-		subnetwork:     cpSpec.Subnetwork.Value,
 	}
 
 	cfg.serviceAccount, err = resolver.GetConfigVarStringValueOrEnv(cpSpec.ServiceAccount, envGoogleServiceAccount)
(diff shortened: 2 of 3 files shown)


examples/gce-machinedeployment.yaml
```diff
@@ -49,6 +49,10 @@ spec:
             machineType: "n1-standard-2"
             # In GB
             diskSize: 50
+            # ether of two network/subnetwork could be specified
```
themue: Not yet clear here about the options and how I would specify e.g. only subnetwork. How would I have to specify it? Expect a convention like with `global/networks/default` for the network.

examples/gce-machinedeployment.yaml
```diff
@@ -49,6 +49,10 @@ spec:
...
+            # in case if both empty — default network will be used
+            network: "my-cool-network"
```
themue: Maybe just a dumb question from my side, not yet took a deeper look into network in GCE. Are there naming conventions?

pkg/cloudprovider/provider/gce/config.go
```diff
@@ -111,6 +113,8 @@ type config struct {
 	machineType    string
 	diskSize       int64
 	diskType       string
+	network        string
```
themue: Where do you retrieve these values from cloud provider spec?

pkg/cloudprovider/provider/gce/service.go
```diff
@@ -40,15 +44,23 @@ func connectComputeService(cfg *config) (*service, error) {
 
 // networkInterfaces returns the configured network interfaces for an instance creation.
 func (svc *service) networkInterfaces(cfg *config) ([]*compute.NetworkInterface, error) {
+	network := cfg.network
+
+	if cfg.network == "" && cfg.subnetwork == "" {
```
themue: I see I have to take a deeper look at GCE networking. How would it look like if network is empty and subnetwork has value? 🤔

examples/gce-machinedeployment.yaml
kron4eg: based on https://www.terraform.io/docs/providers/google/r/compute_instance.html#network

>network - (Optional) The name or self_link of the network to attach this interface to. Either network or subnetwork must be provided.

>subnetwork - (Optional) The name or self_link of the subnetwork to attach this interface to. The subnetwork must exist in the same region this instance will be created in. Either network or subnetwork must be provided.

so it's OK to have ether of them

examples/gce-machinedeployment.yaml
kron4eg: names could be ether "name" or selflink. both works.

pkg/cloudprovider/provider/gce/config.go
kron4eg: ups, I probably missed that

kron4eg: fixed

pkg/cloudprovider/provider/gce/service.go
kron4eg: it will pass subnetwork only to the instance, which is OK

pkg/cloudprovider/provider/gce/config.go
```diff
@@ -123,32 +127,41 @@ func newConfig(resolver *providerconfig.ConfigVarResolver, spec v1alpha1.Provide
 	if err != nil {
 		return nil, err
 	}
+
 	// Setup configuration.
 	cfg := &config{
 		providerConfig: providerConfig,
+		labels:         cpSpec.Labels,
+		diskSize:       cpSpec.DiskSize,
+		network:        cpSpec.Network.Value,
```
themue: Why not using the resolver like below? Any hint is welcome to learn more about our environment. 😅

pkg/cloudprovider/provider/gce/service.go
themue: Thx.

examples/gce-machinedeployment.yaml
themue: Probably double entry, not sure if browser got suggestion. Would write

```suggestion
            # Name or self_link of the network and subnetwork to attach this interface to; either
            # of two network and subnetwork can be provided; if both are empty default network
            # will be taken
```

pkg/cloudprovider/provider/gce/service.go
```diff
@@ -40,15 +44,23 @@ func connectComputeService(cfg *config) (*service, error) {
...
+
 	ifc := &compute.NetworkInterface{
 		AccessConfigs: []*compute.AccessConfig{
 			{
 				Name: "External NAT",
 				Type: "ONE_TO_ONE_NAT",
 			},
 		},
-		Network: "global/networks/default",
+		Network:    network,
```
mrIncompetent: This means the user needs to enter the 'full path' to the actual network?
Like the default network is `global/networks/default`

pkg/cloudprovider/provider/gce/config.go
kron4eg: Resolver will try to look for secret/configmap and once it see there is no such thing for net/subnet it will use .Value as fallback. So I've used that property %)

But you're right, those now are resolved properly and could contain secret/configmap references in spec.

pkg/cloudprovider/provider/gce/service.go
kron4eg: ether of "name" or "self_links" are documented as working solution.

mrIncompetent: Oh great, thanks!

pkg/cloudprovider/provider/gce/config.go
```diff
@@ -123,32 +127,49 @@ func newConfig(resolver *providerconfig.ConfigVarResolver, spec v1alpha1.Provide
...
-	cfg.labels = cpSpec.Labels
+
+	cfg.network, err = resolver.GetConfigVarStringValue(cpSpec.Network)
+	if err != nil {
+		return nil, fmt.Errorf("cannot retrieve network: %v", err)
+	}
+
+	cfg.subnetwork, err = resolver.GetConfigVarStringValue(cpSpec.Subnetwork)
+	if err != nil {
+		return nil, fmt.Errorf("cannot retrieve network: %v", err)
```
themue:
```suggestion
		return nil, fmt.Errorf("cannot retrieve subnetwork: %v", err)
```

pkg/cloudprovider/provider/gce/service.go
kron4eg: full self_links will be fed from terraform as it has only `.self_link` property as a way to address networks/subnetworks, but simple names will work too.

pkg/cloudprovider/provider/gce/config.go
kron4eg: fixed 😅

PR #473, opened 2019-03-28
Add UserData plugins

themue:

Providers for OS specific UserData are now external running child processes. So they later can be exchanged by ones in different and non-public repositories.

Fixes kubermatic/kubermatic#2987

```release-note
Implement UserData provider as external exchangeable processes
```


pkg/userdata/manager/plugin.go
```diff
@@ -0,0 +1,155 @@
...
+	envPluginDir = "MACHINE_CONTROLLER_USERDATA_PLUGIN_DIR"
+
+	// pluginPrefix has to be the prefix of all plugin filenames.
+	pluginPrefix = "machine-controller-userdata-"
+)
+
+// Plugin looks for the plugin executable and calls it for
+// each request.
+type Plugin struct {
+	os      providerconfig.OperatingSystem
```
alvaroaleman: Why have the `os` be part of the Plugin? The plugins are stored in a hashmap whose key is the OS so this is redundant

pkg/userdata/manager/plugin.go
```diff
@@ -0,0 +1,155 @@
+//
+// UserData plugin manager.
+//
+
+package manager
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
```
alvaroaleman: please use `glog` instead of `log`

pkg/userdata/plugin/data.go
```diff
@@ -0,0 +1,43 @@
...
+	"net"
+
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	clusterv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+const (
+	// EnvRequest names the environment variable signalling the
+	// plugin which request the caller wants to have answered.
+	EnvRequest = "REQUEST"
```
alvaroaleman: Can we put these into a package below `pkg/apis` so consuming doesn't result in importing the whole `userdataplugin` package?

alvaroaleman: Actually thinking about this, is there even a reason why they need to be public?

pkg/userdata/plugin/plugin.go
```diff
@@ -0,0 +1,123 @@
...
+	) (string, error)
+}
+
+// Handler cares dispatching of the RPC calls to the given Provider.
+type Handler struct {
+	provider Provider
+}
+
+// UserData receives the UserData RPC message and calls the provider.
+func (h *Handler) UserData(req *UserDataRequest, resp *UserDataResponse) error {
```
alvaroaleman: This is unused and I am a bit confused as to why the linter doesn't complain about that :thinking:

themue: That's why PR is WiP. This type as well as many of the log usages and overly detailed scriptings have just been for debugging or testing. This handler here has i.e. needed for the `net.RPC` solution and will now be removed. The linter doesn't complain because it's exported and may be used by any importer in future. The linter normally only complains private unused types, functions, and variables.

cmd/userdata/centos/provider_test.go
```diff
@@ -43,29 +47,34 @@ UK2ZnINJRcJpB8iRCaCxE8DdcUF0XqIEq6pA272snoLmiXLMvNl3kYEdm+je6voD
 sH9BBH38/SzUmAN4QHSPy1gjqm00OAE8NaYDkh/bzE4d7mLGGMWp/WE3KPSu82HF
 kPe6XoSbiLm/kxk32T0=
 -----END CERTIFICATE-----`
+)
 
-type fakeCloudConfigProvider struct {
+// stubCloudConfigProvider simulates cloud config provider for test.
+type stubCloudConfigProvider struct {
```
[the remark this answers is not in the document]
themue: If it's in k8s can do so. But it's funny. If it shall be a mock, then why don't they name it "mock". Also based on https://martinfowler.com/articles/mocksArentStubs.html I found our implementations action more like stubs.

But as mentioned, if k8s does it as a standard I'll follow.

cmd/userdata/centos/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+	flag.BoolVar(&debug, "debug", false, "Switch for enabling the plugin debugging")
+	flag.Parse()
+
+	// Instantiate provider and start plugin.
+	var provider = &Provider{}
+	var p = plugin.New(provider, debug)
+
+	if err := p.Run(); err != nil {
+		fmt.Printf("error running CentOS plugin: %v", err)
```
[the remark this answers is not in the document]
themue: Sure, here and elsewhere.

hack/ci-e2e-test.sh
```diff
@@ -33,6 +33,13 @@ mv kubectl /usr/local/bin
 echo "Building machine-controller and webhook"
 make machine-controller webhook
 
+# Copy individual plugins with success control.
+echo "Copying machine-controller plugins"
```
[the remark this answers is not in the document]
themue: Yep, only has been explicit for testing.

cmd/userdata/centos/main.go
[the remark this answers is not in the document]
themue: Oh, btw., simple `Print` here had the reason that the calling manager will catch it and cares for logging.

pkg/userdata/plugin/data.go
themue: It's used inside the `plugin` package as well as in the `manager` package. Surely could declare it private and redundant as it is only interesting between these two packages to communicate.

themue: Thought about it. They also share the data types to serialise so API package might make sense. Otherwise I would expect by the linker to not keep unreferenced types of `userdata/plugin` in `machine-controller` binary.

pkg/userdata/plugin/plugin.go
```diff
@@ -0,0 +1,105 @@
...
+		provider: provider,
+		debug:    debug,
+	}
+}
+
+// Run looks for the given request and executes it.
+func (p *Plugin) Run() error {
+	reqCmd := os.Getenv(plugin.EnvRequest)
+	switch reqCmd {
+	case plugin.EnvUserDataRequest:
```
alvaroaleman: Can we maybe just use an `if req.CMD != EnvUserDataRequest { return someErr }\nreturn p.handleUserDataRequest`?

There is just this one case and its unlikely to become more, which is why the current approach of using a switch statement and factoring out the error handling in a distinct func seems needlessly complicated to me

themue: Issue also wanted a _Version_ request. Not yet sure about a use case for it. IMHO can add it later if needed. So I'll simplify it.

pkg/admission/machines.go
```diff
@@ -92,9 +92,9 @@ func (ad *admissionData) defaultAndValidateMachineSpec(spec *clusterv1alpha1.Mac
 		return fmt.Errorf("failed to get cloud provider %q: %v", providerConfig.CloudProvider, err)
 	}
 
-	// Verify operating system
-	if _, err := userdata.ForOS(providerConfig.OperatingSystem); err != nil {
-		return fmt.Errorf("failed to get OS '%s': %v", providerConfig.OperatingSystem, err)
+	// Verify operating system.
+	if !manager.Supports(providerConfig.OperatingSystem) {
```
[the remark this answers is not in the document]
alvaroaleman: ?

pkg/controller/machine/machine.go
```diff
@@ -52,12 +44,20 @@ import (
...
 	machinescheme "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/scheme"
 	clusterlistersv1alpha1 "sigs.k8s.io/cluster-api/pkg/client/listers_generated/cluster/v1alpha1"
+
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider"
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider/cloud"
+	cloudprovidererrors "github.com/kubermatic/machine-controller/pkg/cloudprovider/errors"
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider/instance"
+	"github.com/kubermatic/machine-controller/pkg/node/eviction"
+	"github.com/kubermatic/machine-controller/pkg/providerconfig"
+	"github.com/kubermatic/machine-controller/pkg/userdata/manager"
```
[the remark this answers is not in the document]
alvaroaleman: ?

pkg/controller/machine/machine.go
```diff
@@ -52,12 +44,20 @@ import (
...
+	"github.com/kubermatic/machine-controller/pkg/userdata/plugin"
```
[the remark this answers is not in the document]
alvaroaleman: ?

pkg/userdata/centos/centos.go
```diff
@@ -0,0 +1,36 @@
+package centos
```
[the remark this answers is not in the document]
alvaroaleman: ?

pkg/userdata/manager/manager.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// debug contains the debug flag, default is false.
+	debug bool
+
+	// plugins contains the registered plugins.
+	plugins map[providerconfig.OperatingSystem]*Plugin
+)
+
+// init  checks the plugin debug flag.
+func init() {
```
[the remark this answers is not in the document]
alvaroaleman: ?

pkg/userdata/manager/manager.go
```diff
@@ -0,0 +1,86 @@
...
+	var found bool
+	if p, found = plugins[os]; !found {
+		return nil, ErrPluginNotFound
+	}
+
+	return p, nil
+}
+
+// Supports answers if the userdata manager supports the
+func Supports(os providerconfig.OperatingSystem) bool {
```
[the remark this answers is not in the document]
alvaroaleman: Or am i missing something?

pkg/userdata/plugin/plugin.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// Plugin implements a convenient helper to map the request to the given
+// provider and return the response.
+type Plugin struct {
+	provider Provider
+	debug    bool
+}
+
+// New creates a new plugin. Debug flag is not yet handled.
```
alvaroaleman: The second part of the comment is wrong

pkg/admission/machines.go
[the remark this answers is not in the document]
themue: `Supports()` doesn't care for supported OSses but if the `init()` has found a plugin for the given OS. Only a convenient wrapper instead of returning `ErrPluginNotFound`.  So `Supports()` is the wrong name, yes. Better would be `Found()`.

pkg/controller/machine/machine.go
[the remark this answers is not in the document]
themue: Can do, sure.

pkg/userdata/centos/centos.go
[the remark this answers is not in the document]
themue: Following the standard with `/cmd` for commands and `pkg` for libs. As the provider are not intended to be used as libs elsewhere I moved them to their command packages too. Dunno why I had the `Config` still in `pkg`. Will move the providers there too.

pkg/userdata/manager/manager.go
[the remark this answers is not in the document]
themue: Hehe, returns to a draft I've implemented first. 😅

pkg/userdata/manager/manager.go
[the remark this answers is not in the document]
themue: It's only a nicer variant of what's done with `_, err := userdata.ForOS(...); err != nil` so far.

pkg/controller/machine/machine.go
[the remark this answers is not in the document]
themue: Yip.

pkg/admission/admission.go
```diff
@@ -10,23 +10,28 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/mattbaird/jsonpatch"
-
 	admissionv1beta1 "k8s.io/api/admission/v1beta1"
```
alvaroaleman: Please keep the newline, we usually keep kubernetes imports in a dedicated group

pkg/admission/machines.go
```diff
@@ -5,16 +5,13 @@ import (
 	"fmt"
 
 	"github.com/golang/glog"
+	admissionv1beta1 "k8s.io/api/admission/v1beta1"
```
alvaroaleman: k8s imports are usually in the same import group

pkg/controller/machine/machine.go
```diff
@@ -53,12 +45,20 @@ import (
...
 	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
-
 	"sigs.k8s.io/cluster-api/pkg/apis/cluster/common"
 	clusterv1alpha1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	clusterv1alpha1clientset "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset"
 	machinescheme "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/scheme"
 	clusterlistersv1alpha1 "sigs.k8s.io/cluster-api/pkg/client/listers_generated/cluster/v1alpha1"
+
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider"
```
alvaroaleman: While splitting our own imports from the third party imports into an own group is a good thing, we usually order it like this:

```
Go SDK
external packages which do not apply to other rules (Like github.com/golang/glog, etc.)
github.com/kubermatic
k8s.io/*
```

Ref: https://github.com/kubermatic/kubermatic/blob/9c59130831f63a2ebe703d7ae328fd9459112d7a/docs/code-style.md#imports

themue: Good to know, shout have a code style guide document for it. Only know the general rule

```
import (
    stdlib

    3rd party packages

    own packages
)
```

and don't know why importing k8s should be treated differently. We often even mix own and k8s packages in imports.

pkg/admission/machines.go
themue: See comment below, but will do so in future. Would like a code style document for it.

pkg/controller/machine/machine.go
alvaroaleman: The code style document is linked above

pkg/userdata/manager/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+	if resp.Err != "" {
+		return "", fmt.Errorf("%s", resp.Err)
+	}
+	return resp.UserData, nil
+}
+
+// findPlugin tries to find the executable of the plugin.
+func (p *Plugin) findPlugin(name string) error {
+	filename := pluginPrefix + name
+	glog.Infof("looking for plugin '%s'", filename)
```
alvaroaleman: Please move this to level 3 as this is debug logging. `%q` does btw more or less the same as `'%s'`

pkg/userdata/manager/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+		return err
+	}
+	dirs = append(dirs, workingDir)
+	path := os.Getenv("PATH")
+	pathDirs := strings.Split(path, string(os.PathListSeparator))
+	dirs = append(dirs, pathDirs...)
+	// Now take a look.
+	for _, dir := range dirs {
+		command := dir + string(os.PathSeparator) + filename
+		glog.Infof("checking '%s'", command)
```
alvaroaleman: Please move this to level 3 as this is debug logging. `%q` does btw more or less the same as `'%s'`

pkg/userdata/manager/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+		fi, err := os.Stat(command)
+		if err != nil {
+			if os.IsNotExist(err) {
+				continue
+			}
+			return fmt.Errorf("error when looking for %q: %v", command, err)
+		}
+		if fi.IsDir() || (fi.Mode()&0111 == 0) {
+			glog.Infof("found '%s', but is no executable", command)
```
alvaroaleman: Please move this to level 3 as this is debug logging. `%q` does btw more or less the same as `'%s'`

pkg/userdata/manager/plugin.go
```diff
@@ -0,0 +1,146 @@
...
+			continue
+		}
+		p.command = command
+		glog.Infof("found '%s'", command)
```
alvaroaleman: Please move this to level 3 as this is debug logging. `%q` does btw more or less the same as `'%s'`

pkg/userdata/plugin/plugin.go
```diff
@@ -0,0 +1,90 @@
...
+	return p.printResponse(resp)
+}
+
+// printResponse marshals the respons and prints it to stdout.
+func (p *Plugin) printResponse(resp interface{}) error {
+	bs, err := json.Marshal(resp)
+	if err != nil {
+		return err
+	}
+	fmt.Printf("%s", string(bs))
```
alvaroaleman: This is an unhandled error, I am a bit surprised the linter doesn't compain about it

Dockerfile
```diff
@@ -2,7 +2,6 @@ FROM alpine:3.7
 
 RUN apk add --no-cache ca-certificates cdrkit
 
-COPY machine-controller /usr/local/bin
-COPY webhook /usr/local/bin
+COPY machine-controller machine-controller-userdata-* webhook /usr/local/bin
```
alvaroaleman: The `/usr/local/bin` need a trailing `/`, otherwise this errors. This is also the reason the e2e tests fail

pkg/userdata/manager/plugin.go
themue: More or less. `%q` double-quotes and escapes while `%s` takes the bytes of the sting uninterpreted. But can change it.

pkg/userdata/manager/plugin.go
themue: Hmm, has comment been deleted? Wrote the both are almost the same. `%q` encodes the string and puts in in double-quotes while `%s` simply takes the bytes of the string as they are. But will change.

pkg/userdata/plugin/plugin.go
themue: I have to admit I almost never have seen an error check of a `fmt` function and also never seen a linter complaining about it. But will simply pass it.

Dockerfile
themue: Grmpflx! 🤬 Yes, already once stumbled about it and didn't remembered it here. Gnah! Good finding.

mrIncompetent:
/hold cancel

Great work :)

PR #476, opened 2019-04-02
 Enable `CPUAccounting` and `MemoryAccounting` for kubelet.service

nikhita:
To avoid the error encountered in https://github.com/kubernetes/kubernetes/issues/56850

For the solution, see https://github.com/kubernetes/kubernetes/issues/56850#issuecomment-406241077

Enabling these causes systemd to create the missing `/system.slice/*.service` cgroups for all services.

This allows the kubelet `/stats/summary` API to report the correct `systemContainer` metrics for the runtime and kubelet.

**Release Note**:
```release-note
NONE
```


alvaroaleman:
@nikhita Is this an issue on Ubuntu only? In the linked issue the OS mentioned is CentOS, not Ubuntu

nikhita:
> @nikhita Is this an issue on Ubuntu only? In the linked issue the OS mentioned is CentOS, not Ubuntu

I think it should be CentOS only because this issue arises because the kubelet RPM doesn't come with these settings by default. But I wasn't sure on which OS this bug wasn't encountered while using machine-controller... @mrIncompetent was it CentOS?

mrIncompetent:
Yes, only CentOS was affected.

Is it safe to set the setting for all operating systems?
If so, we should not use a drop-in

themue:
After dropping changes for CoreOS and Ubuntu only CentOS keeps failing. Have to check exact cgroup setting.

alvaroaleman:
@themue @mrIncompetent Did we find an answer to the question of "Should we enable this for all distributions"? The test failures in the e2e tests were apparently unrelated to the OS:

> After dropping changes for CoreOS and Ubuntu only CentOS keeps failing. Have to check exact cgroup setting.

themue:
@alvaroaleman Not yet. Wanted to handle this topic and the linting of yaml in two individual issues.

alvaroaleman:
"Linting of yaml"?

IMHO it would made have a lot of sense to find out where we want to activate this before merging anything.

themue:
So far would see it as part of the unit tests. The unit tests right now compare the template output. Sadly they had all had the same indentation problem, so the tests didn't failed. 

For this time it's fixed but I would in future first see a linter check the test data and afterwards the unit test doing the comparison like today. In case the test data is fine but the unit tests fail doing the diffing we would get it it shown.

alvaroaleman:
Okay. Can you please create tickets for both topics and move the ticket for checking if we want CPUAccounting and MemoryAccounting for Ubuntu and CoreOS into the current sprint?

themue:
Sure, that's my plan. 😉

PR #492, opened 2019-04-15
Dont show leader election log by default

### LINKED ISSUE #424: remove/disable leader election info log
As operator of Kubermatic, I get every 2 seconds the following log line in the `machine-controller`:
```
cluster-9k6tgrg7hj | I1203 17:15:33.863797       1 leaderelection.go:199] successfully renewed lease kube-system/machine-controller
```
This info don't give any benefit for anyone. Only a error message when something breaks helps. So I would suggest to remove this log message or move it to a debug level. 

The issue is may a follow up from https://github.com/kubermatic/kubermatic/pull/2471:

alvaroaleman:

Disables logging of leader election related messages by default. Additionally moves all our debug log to level 3, as level 4 contains the leader election log

Fixes #424 

```release-note
none
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -548,9 +548,9 @@ func readinessChecks(kubeconfigProvider machinecontroller.KubeconfigProvider) ma
 // createRecorder creates a new event recorder which is later used by the leader election
 // library to broadcast events
 func createRecorder(kubeClient *kubernetes.Clientset) record.EventRecorder {
-	glog.V(4).Info("creating event broadcaster")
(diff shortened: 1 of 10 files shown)


mrIncompetent:
This does not adhere to the [logging guidelines](https://github.com/kubernetes/community/blob/b3349d5b1354df814b67bbdee6890477f3c250cb/contributors/devel/logging.md)

The proper solution would be to check if all logging we do adheres the guidelines & set the verbosity to `-v=2` on the Deployment.

I'm going to approve this anyway, since it fixes the issue.
/lgtm
/approve

PR #493, opened 2019-04-15
Packet provider

deitch:

Add Packet support to the kubermatic machine-controller

Followed the document instructions at [here](https://github.com/kubermatic/machine-controller/blob/master/docs/howto-provider.md) . 

I am fairly certain we missed something. Also not sure the easiest way to test it. Would request we do the following:

1. Have someone do a quick look at it to see if there is something glaringly obvious.
2. Have someone write here the best way to test just this against Packet. 

```release-note
Add Packet cloud support.
```

### DIFF
--- a/pkg/cloudprovider/provider/packet/provider.go
+++ b/pkg/cloudprovider/provider/packet/provider.go
@@ -1,0 +1,504 @@
+package packet
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/golang/glog"
+	"github.com/packethost/packngo"
+
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider/cloud"
+	cloudprovidererrors "github.com/kubermatic/machine-controller/pkg/cloudprovider/errors"
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider/instance"
+	"github.com/kubermatic/machine-controller/pkg/providerconfig"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/cluster-api/pkg/apis/cluster/common"
+	"sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+const (
+	machineUIDTag       = "kubermatic-machine-controller:machine-uid"
+	defaultBillingCycle = "hourly"
+)
+
+// New returns a Packet provider
+func New(configVarResolver *providerconfig.ConfigVarResolver) cloud.Provider {
+	return &provider{configVarResolver: configVarResolver}
+}
+
+type RawConfig struct {
+	APIKey       providerconfig.ConfigVarString   `json:"apiKey"`
+	ProjectID    providerconfig.ConfigVarString   `json:"projectID"`
+	BillingCycle providerconfig.ConfigVarString   `json:"billingCycle"`
+	InstanceType providerconfig.ConfigVarString   `json:"instanceType"`
+	Facilities   []providerconfig.ConfigVarString `json:"facilities"`
+	OS           providerconfig.ConfigVarString   `json:"os"`
+	Tags         []providerconfig.ConfigVarString `json:"tags"`
+}
+
+type Config struct {
+	APIKey       string
+	ProjectID    string
+	BillingCycle string
+	InstanceType string
+	Facilities   []string
+	OS           string
+	Tags         []string
+}
+
+// because we have both Config and RawConfig, we need to have func for each
+// ideally, these would be merged into one
+func (c *Config) populateDefaults() {
+	if c.BillingCycle == "" {
+		c.BillingCycle = defaultBillingCycle
+	}
+}
+
+func (c *RawConfig) populateDefaults() {
+	if c.BillingCycle.Value == "" {
+		c.BillingCycle.Value = defaultBillingCycle
+	}
+}
+
+type provider struct {
+	configVarResolver *providerconfig.ConfigVarResolver
+}
+
+func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *RawConfig, *providerconfig.Config, error) {
+	if s.Value == nil {
+		return nil, nil, nil, fmt.Errorf("machine.spec.providerconfig.value is nil")
+	}
+	pconfig := providerconfig.Config{}
+	err := json.Unmarshal(s.Value.Raw, &pconfig)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	rawConfig := RawConfig{}
+	if err = json.Unmarshal(pconfig.CloudProviderSpec.Raw, &rawConfig); err != nil {
+		return nil, nil, nil, err
+	}
+
+	c := Config{}
+	c.APIKey, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.APIKey, "PACKET_API_KEY")
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"apiKey\" field, error = %v", err)
+	}
+	c.ProjectID, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.ProjectID)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"projectID\" field, error = %v", err)
+	}
+	c.InstanceType, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.InstanceType)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"instanceType\" field, error = %v", err)
+	}
+	c.BillingCycle, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.BillingCycle)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"billingCycle\" field, error = %v", err)
+	}
+	for i, tag := range rawConfig.Tags {
+		tagValue, err := p.configVarResolver.GetConfigVarStringValue(tag)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to read the value for the Tag at index %d of the \"tags\" field, error = %v", i, err)
+		}
+		c.Tags = append(c.Tags, tagValue)
+	}
+	for i, facility := range rawConfig.Facilities {
+		facilityValue, err := p.configVarResolver.GetConfigVarStringValue(facility)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to read the value for the Tag at index %d of the \"facilities\" field, error = %v", i, err)
+		}
+		c.Facilities = append(c.Facilities, facilityValue)
+	}
+
+	// ensure we have defaults
+	c.populateDefaults()
+
+	return &c, &rawConfig, &pconfig, err
+}
+
+func (p *provider) getPacketDevice(machine *v1alpha1.Machine) (*packngo.Device, *packngo.Client, error) {
+	c, _, _, err := p.getConfig(machine.Spec.ProviderSpec)
+	if err != nil {
+		return nil, nil, cloudprovidererrors.TerminalError{
+			Reason:  common.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Failed to parse MachineSpec, due to %v", err),
+		}
+	}
+
+	client := getClient(c.APIKey)
+	device, err := getDeviceByTag(client, c.ProjectID, generateTag(string(machine.UID)))
+	if err != nil {
+		return nil, nil, err
+	}
+	return device, client, nil
+}
+
+func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
+	c, _, pc, err := p.getConfig(spec.ProviderSpec)
+	if err != nil {
+		return fmt.Errorf("failed to parse config: %v", err)
+	}
+
+	if c.APIKey == "" {
+		return errors.New("apiKey is missing")
+	}
+	if c.InstanceType == "" {
+		return errors.New("instanceType is missing")
+	}
+	if c.ProjectID == "" {
+		return errors.New("projectID is missing")
+	}
+
+	_, err = getNameForOS(pc.OperatingSystem)
+	if err != nil {
+		return fmt.Errorf("invalid/not supported operating system specified %q: %v", pc.OperatingSystem, err)
+	}
+
+	client := getClient(c.APIKey)
+
+	if len(c.Facilities) == 0 || c.Facilities[0] == "" {
+		return fmt.Errorf("must have at least one non-blank facility")
+	}
+
+	// get all valid facilities
+	facilities, _, err := client.Facilities.List(nil)
+	if err != nil {
+		return fmt.Errorf("failed to list facilities: %v", err)
+	}
+	// ensure our requested facilities are in those facilities
+	if missingFacilities := itemsNotInList(facilityProp(facilities, "Code"), c.Facilities); len(missingFacilities) > 0 {
+		return fmt.Errorf("unknown facilities: %s", strings.Join(missingFacilities, ","))
+	}
+
+	// get all valid plans a.k.a. instance types
+	plans, _, err := client.Plans.List(nil)
+	if err != nil {
+		return fmt.Errorf("failed to list instance types / plans: %v", err)
+	}
+	// ensure our requested plan is in those plans
+	validPlanNames := planProp(plans, "Name")
+	if missingPlans := itemsNotInList(validPlanNames, []string{c.InstanceType}); len(missingPlans) > 0 {
+		return fmt.Errorf("unknown instance type / plan: %s, acceptable plans: %s", strings.Join(missingPlans, ","), strings.Join(validPlanNames, ","))
+	}
+
+	return nil
+}
+
+func (p *provider) Create(machine *v1alpha1.Machine, _ *cloud.MachineCreateDeleteData, userdata string) (instance.Instance, error) {
+	c, _, pc, err := p.getConfig(machine.Spec.ProviderSpec)
+	if err != nil {
+		return nil, cloudprovidererrors.TerminalError{
+			Reason:  common.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Failed to parse MachineSpec, due to %v", err),
+		}
+	}
+
+	client := getClient(c.APIKey)
+
+	imageName, err := getNameForOS(pc.OperatingSystem)
+	if err != nil {
+		return nil, cloudprovidererrors.TerminalError{
+			Reason:  common.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Invalid operating system specified %q, details = %v", pc.OperatingSystem, err),
+		}
+	}
+
+	serverCreateOpts := &packngo.DeviceCreateRequest{
+		Hostname:     machine.Spec.Name,
+		UserData:     userdata,
+		ProjectID:    c.ProjectID,
+		Facility:     c.Facilities,
+		BillingCycle: c.BillingCycle,
+		Plan:         c.InstanceType,
+		OS:           imageName,
+		Tags: []string{
+			generateTag(string(machine.UID)),
+		},
+	}
+
+	device, res, err := client.Devices.Create(serverCreateOpts)
+	if err != nil {
+		return nil, packetErrorToTerminalError(err, res, "failed to create server")
+	}
+
+	return &packetDevice{device: device}, nil
+}
+
+func (p *provider) Cleanup(machine *v1alpha1.Machine, _ *cloud.MachineCreateDeleteData) (bool, error) {
+	instance, err := p.Get(machine)
+	if err != nil {
+		if err == cloudprovidererrors.ErrInstanceNotFound {
+			return true, nil
+		}
+		return false, err
+	}
+
+	c, _, _, err := p.getConfig(machine.Spec.ProviderSpec)
+	if err != nil {
+		return false, cloudprovidererrors.TerminalError{
+			Reason:  common.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Failed to parse MachineSpec, due to %v", err),
+		}
+	}
+
+	client := getClient(c.APIKey)
+
+	res, err := client.Devices.Delete(instance.(*packetDevice).device.ID)
+	if err != nil {
+		return false, packetErrorToTerminalError(err, res, "failed to delete the server")
+	}
+
+	return false, nil
+}
+
+func (p *provider) AddDefaults(spec v1alpha1.MachineSpec) (v1alpha1.MachineSpec, error) {
+	_, rawConfig, _, err := p.getConfig(spec.ProviderSpec)
+	if err != nil {
+		return spec, err
+	}
+	rawConfig.populateDefaults()
+	spec.ProviderSpec.Value, err = setProviderSpec(*rawConfig, spec.ProviderSpec)
+	if err != nil {
+		return spec, err
+	}
+	return spec, nil
+}
+
+func (p *provider) Get(machine *v1alpha1.Machine) (instance.Instance, error) {
+	device, _, err := p.getPacketDevice(machine)
+	if err != nil {
+		return nil, err
+	}
+	if device != nil {
+		return &packetDevice{device: device}, nil
+	}
+
+	return nil, cloudprovidererrors.ErrInstanceNotFound
+}
+
+func (p *provider) MigrateUID(machine *v1alpha1.Machine, newID types.UID) error {
+	device, client, err := p.getPacketDevice(machine)
+	if err != nil {
+		return err
+	}
+	if device == nil {
+		glog.Infof("No instance exists for machine %s", machine.Name)
+		return nil
+	}
+
+	// go through existing labels, make sure that no other UID label exists
+	tags := make([]string, 0)
+	for _, t := range device.Tags {
+		// filter out old UID tag(s)
+		if _, err := getTagUID(t); err != nil {
+			tags = append(tags, t)
+		}
+	}
+
+	// create a new UID label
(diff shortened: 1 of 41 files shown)


deitch:
Fixed formatting issues.

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,420 @@
...
+	if len(c.Facilities) == 0 || c.Facilities[0] == "" {
+		return fmt.Errorf("must have at least one non-blank facility")
+	}
+
+	// get all valid facilities
+	facilities, _, err := client.Facilities.List(nil)
+	if err != nil {
+		return fmt.Errorf("failed to list facilities: %v", err)
+	}
+	// ensure our requested facilities are in those facilities
```
alvaroaleman: Is the list an expensive API call? Could we instead just GET the ones we are interested in?

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,420 @@
...
+		return fmt.Errorf("unknown facilities: %s", strings.Join(missingFacilities, ","))
+	}
+
+	// get all valid plans a.k.a. instance types
+	plans, _, err := client.Plans.List(nil)
+	if err != nil {
+		return fmt.Errorf("failed to list instance types / plans: %v", err)
+	}
+	// ensure our requested plan is in those plans
+	if missingPlans := itemsNotInList(planProp(plans, "Slug"), []string{c.InstanceType}); len(missingPlans) > 0 {
```
alvaroaleman: Same here, what about just getting the one we are interested in?

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+	return "", providerconfig.ErrOSNotSupported
+}
+
+func getClient(apiKey string) *packngo.Client {
+	return packngo.NewClientWithAuth("kubermatic", apiKey, nil)
+}
+
+func generateTag(ID string) string {
+	return fmt.Sprintf("gardener:%s:%s", machineUIDTag, ID)
```
alvaroaleman: That does not feel right ;)

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,430 @@
...
+
+	tag := generateTag(string(machine.UID))
+	client := getClient(c.APIKey)
+	device, err := getDeviceByTag(client, c.ProjectID, tag)
+	if err != nil {
+		return err
+	}
+
+	// append to our tags
+	tags := append(device.Tags, tag)
```
alvaroaleman: We need to actually replace the old one, ideally with one API call so we cant end up with both a new and the old tag

pkg/cloudprovider/provider/packet/provider.go
deitch: Actually, there is no option to GET a facility. The only option is to list them. It has them in a static table, so it isn't as if it is collecting them from lots of places. It is (or should be, as I understand it), a simple `N(1)` call.

pkg/cloudprovider/provider/packet/provider.go
deitch: Same for `Plans` as for `Facilities`. It is a simple `N(1)` op, and there is no `GET` for it.

pkg/cloudprovider/provider/packet/provider.go
deitch: I am more than happy to change it to something else. Suggestion?

pkg/cloudprovider/provider/packet/provider.go
deitch: Can you explain it some more? What are we looking to do?

pkg/cloudprovider/provider/packet/provider.go
alvaroaleman: Change the `gardener` for `kubermatic` or just `machine-controller`?

pkg/cloudprovider/provider/packet/provider.go
alvaroaleman: We use the UID of the Kubernetes object as unique identifier of the object in the providers API. That means the object in the providers API must have exactly one UID tag/label whatever it offers. If I don't misunderstand your code, it will currently put a second UID tag on the instance when `MigrateUID` gets called

alvaroaleman:
The tests fail btw because of 

```
# github.com/kubermatic/machine-controller/pkg/cloudprovider/provider/packet
pkg/cloudprovider/provider/packet/provider.go:18:2: imported and not used: "github.com/kubermatic/machine-controller/vendor/k8s.io/apimachinery/pkg/runtime"
```

pkg/cloudprovider/provider/packet/provider.go
deitch: Doh! I copied/pasted a log message from somewhere else! Definite facepalm!

deitch:
> The tests fail btw because of

Strange that it didn't pick it up locally. Oh well.

pkg/cloudprovider/provider/packet/provider.go
deitch: Tags at packet aren't key/value pairs, but just arbitrary strings. You easily can make it key/value by making the string, e.g., 

```yml
- "Name:Alvaro"
- "Purpose:Success"
```

etc.

In this case, it might have other tags, perhaps even unrelated to this specific UID tag. So I am appending this new tag to the existing tags, whatever they may be.

However, I see how you might get two or more actual UID tags (independent of whatever other tags there are). I am going to change it so that it ensures you only get one of the UID tags.

deitch: OK done.

deitch:
My local tests all are working well against actual Packet servers.

deitch:
My local test had the following errors. See [this gist](https://gist.github.com/deitch/09396774e5816bdd79b211bdd8755fb9)

Some direction please?

alvaroaleman:
@deitch That just means the tests timed out. You can get some info about where they were when they timed out by running them with the `-v` flag

deitch:
I was. Here is the command-line you suggested I run:

```
$ go test -race -tags=e2e -parallel 24 -v -timeout 20m ./test/e2e/... -identifier avi -run TestPacketProvisioningE2E
```

alvaroaleman:
Oh, you need to enable glog for that by appending `-args -logtostderr`

But at the end of the day these tests just create a machinedeployment, wait for a machine, wait for a node for the machine, wait for the node to be ready, delete everything, wait for everything to be gone. Its probably easier to just do that yourself and see where it hangs?

deitch:
> But at the end of the day these tests just create a machinedeployment, wait for a machine, wait for a node for the machine, wait for the node to be ready, delete everything, wait for everything to be gone. Its probably easier to just do that yourself and see where it hangs?

Yup. My local runs (and the e2e tests) created machines and assigned the right tags. How do I check where it blew up?

alvaroaleman:
@deitch Do they also appear as Kubernetes node (`kubectl get node`)? If yes, does the node become `ready`? If not, just SSH onto the instance and check what `setup.service`

Here the e2e tests already fail in the very beginning btw:

```
    --- FAIL: TestPacketProvisioningE2E/ubuntu-1.13.5 (1.83s)
        helper.go:135: verify failed due to error=failed to verify creation of node for MachineDeployment: Internal error occurred: admission webhook "machinedeployments.machine-controller.kubermatic.io" denied the request: defaulting or validation failed: validation failed: unknown instance type / plan: t1.small
```

deitch:
That is helpful. I think I found something.

deitch:
Fixed that issue and pushed .

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,455 @@
...
+
+func (s *packetDevice) ID() string {
+	return s.device.ID
+}
+
+func (s *packetDevice) Addresses() []string {
+	// returns addresses in CIDR format
+	var addresses []string
+	for _, ip := range s.device.Network {
+		addresses = append(addresses, fmt.Sprintf("%s/%d", ip.Address, ip.CIDR))
```
alvaroaleman: This must not contain the CIDR, just the address. For providers that don't have a cloud provider, we match the node by IP address and the nodes ip addresses dont have a CIDR in them

deitch: Got it. Fixed.

deitch:
My guess is that these are going to continue to fail until #504 gets merged in, since I changed the `userdata` in this repo as well, but not the tests. Once that is in, I will rebase here.

Either way, my own tests are not working, see lots of logs and config info in the [gist](https://gist.github.com/deitch/570f39c0127d58ff0f4e1927cdfcd278)

deitch:
Per the Slack discussion, I can confirm that [this line](https://github.com/kubermatic/machine-controller/blob/master/test/e2e/provisioning/migrateuidscenario.go#L93):

```go
prov.Create(machine, machineCreateDeleteData, "#cloud-config")
```

is causing a `422` by the Packet API, because the userdata appears to be invalid. It is quite content with a newline at the end, though, so adding that here.

deitch:
Rebased with correct userdata based on #504 . Restarting packet tests

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,455 @@
...
+	return "", "", nil
+}
+
+func (p *provider) MachineMetricsLabels(machine *v1alpha1.Machine) (map[string]string, error) {
+	labels := make(map[string]string)
+
+	c, _, _, err := p.getConfig(machine.Spec.ProviderSpec)
+	if err == nil {
+		labels["size"] = c.InstanceType
+		labels["facilities"] = strings.Join(c.Facilities, ",")
```
alvaroaleman: Not completely happy about this, as it doesn't actually tell the user in which facility the instance ended up being if there was more than one submitted. But I am not sure what a better solution is, as doing an API call every time `MachineMetricsLabels` gets called is definitely too expensive

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	client := getClient(c.APIKey)
+
+	if len(c.Facilities) == 0 || c.Facilities[0] == "" {
+		return fmt.Errorf("must have at least one non-blank facility")
+	}
+
+	// get all valid facilities
+	facilities, _, err := client.Facilities.List(nil)
```
alvaroaleman: Not really something to do with the PR, but the `response Body`/`Request URL`/`Curl` fields in your API docs about this are pretty empty: https://www.packet.com/developers/api/#retrieve-all-facilities

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,455 @@
...
+	imageName, err := getNameForOS(pc.OperatingSystem)
+	if err != nil {
+		return nil, cloudprovidererrors.TerminalError{
+			Reason:  common.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Invalid operating system specified %q, details = %v", pc.OperatingSystem, err),
+		}
+	}
+
+	serverCreateOpts := &packngo.DeviceCreateRequest{
+		Hostname:     machine.Spec.Name,
```
alvaroaleman: Does the hostname have to be unique per project?

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,455 @@
...
+func (p *provider) Get(machine *v1alpha1.Machine) (instance.Instance, error) {
+	device, _, err := p.getPacketDevice(machine)
+	if err != nil {
+		return nil, err
+	}
+	if device != nil {
+		return &packetDevice{device: device}, nil
+	}
+
+	return nil, cloudprovidererrors.ErrInstanceNotFound
```
alvaroaleman: If the API can not find the instance it does return an empty error and a nil `device`...?

alvaroaleman:
The tests btw failed  because of some gcr issue that broke the kubeadm installation:
```
[ERROR ImagePull]: failed to pull image k8s.gcr.io/kube-scheduler:v1.14.0: output: Error response from daemon: Get https://k8s.gcr.io/v2/kube-scheduler/manifests/v1.14.0: Get https://k8s.gcr.io/v2/token?scope=repository%3Akube-scheduler%3Apull&service=k8s.gcr.io: net/http: TLS handshake timeout
```

and 

```
[ERROR ImagePull]: failed to pull image k8s.gcr.io/kube-controller-manager:v1.14.0: output: Error response from daemon: Get https://k8s.gcr.io/v2/kube-controller-manager/manifests/v1.14.0: Get https://k8s.gcr.io/v2/token?scope=repository%3Akube-controller-manager%3Apull&service=k8s.gcr.io: net/http: TLS handshake timeout
```

pkg/cloudprovider/provider/packet/provider.go
deitch: Right. It is an interesting feature of Packet's API that you can specify one, multiple or "any" facility. See [here](https://www.packet.com/developers/api/#create-a-device):

```
The facilities attribute specifies in what datacenter you wish to create the device.

You can either specify a single facility { "facility": "f1" } , or you can instruct to create the device in the best available datacenter { "facility": "any" }. Additionally it is possible to set a prioritized location selection.

For example { "facility": ["f3", "f2", "any"] } will try to assign to the facility f3, if there are no available f2, and so on. If "any" is not specified for "facility", the request will fail unless it can assign in the selected locations.
```

When you actually do `Create()` the device, it returns a `device` as well. I don't know if that returns the _requested_ facilities or _actual_ facility. I have some go code with a quick experiment to find out. 

If it does, could we not just retrieve it, since we save the device at the end of `Create()`?

pkg/cloudprovider/provider/packet/provider.go
deitch: Yes, I agree 100%. Going to post internally about this (and the `userdata` issue we saw earlier).

pkg/cloudprovider/provider/packet/provider.go
deitch: Confirmed. The return from the call _does_ tell you what facility it is in, and we save that:

```go
func (p *provider) Create(machine *v1alpha1.Machine, _ *cloud.MachineCreateDeleteData, userdata string) (instance.Instance, error) {
        // ...
        device, res, err := client.Devices.Create(serverCreateOpts)  
        // This "device" contains it in `device.Facility.Name`
        // ...
        // and we return it here
        return &packetDevice{device: device}, nil
}
```

pkg/cloudprovider/provider/packet/provider.go
deitch: No, it does not. Why?

deitch: And to verify, I just created two servers with the exact same hostname.

pkg/cloudprovider/provider/packet/provider.go
deitch: Oh, did I get it backwards? Oops. Fixing.

pkg/cloudprovider/provider/packet/provider.go
alvaroaleman: Perfect

pkg/cloudprovider/provider/packet/provider.go
alvaroaleman: Its not that easy, because we would need to store this state somewhere, probably in `.Status.ProviderStatus`. We can leave that as a follow-up thought.

pkg/cloudprovider/provider/packet/provider.go
deitch: No, wait, that makes sense. It returns `nil` for the device and `cloudprovidererrors.ErrInstanceNotFound` for the error. I took this from all of the other cloud providers, see [here](https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/aws/provider.go#L602), [here](https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/azure/provider.go#L558), [here](https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/hetzner/provider.go#L288), etc.

pkg/cloudprovider/provider/packet/provider.go
deitch: Makes sense. We do _have_ it, though, so that is a good first step. It is easier to find a place to store something we have, than to find a way to get that which we do not. :-)

pkg/cloudprovider/provider/packet/provider.go
alvaroaleman: Okay it actually does work, because `getDeviceByTag` returns a `nil` error if it can't find the device. I personally find that subpar, because this means all users of `p.getPacketDevice` must check if the returned `device` is non-nil (Which they currently don't do). What about making `getDeviceByTag` return an error if it cant find the device?

deitch: I am having a hard time following your logic. I will try to work it through here.

1. `getDeviceByTag()` (internal only): gets the device that matches the tag, or `nil` if none is found. Returns a non-nil `error` if there is a real error. 
2. `p.getPacketDevice()` (internal only): calls `getDeviceByTag()`. It really just wraps `getDeviceByTag()` while providing some additional convenience (I probably could subsume `getDeviceByTag()` into it, since it no longer is used elsewhere, but I like small, easy-to-understand functions). It returns the device if found, `nil` if not; `error` only if there was an error.
3. The internal convenience function `p.getPacketDevice()` is consumed by `p.Get()` and `p.MigrateUID()`, which check that... 

Ah, I see. You are saying that the API from `p.getPacketDevice()` is "return `nil` if no device found", but not everyone is respecting that, particularly `p.MigrateUID()`.

OK, that is a bug in my implementation of `p.MigrateUID()`. I will fix that.

I would like to leave `p.getPacketDevice()` as is, as I tend to view 3 distinct use cases:

* found the device: `device` set; `err = nil`
* had a comms or auth error: `device = nil`; `err` set
* no error but no matching device: `device = nil`; `err = nil`.

deitch: I took inspiration from the Hetzner and AWS code: `glog` a message and return `nil`.

deitch: Updated, `lint` and `build` all look fine.

alvaroaleman: Agree except for `no error but no matching device: device = nil; err = nil.`. That is counterintuitive, if I get no error I expect the `device` to be `non-nil`. Also it makes for redundant nil-checks everywhere `p.getPacketDevice` is used.

Where in the Hetzner or AWS code do you see a `nil` error return when no instance was found?

deitch: > Agree except for no error but no matching device: device = nil; err = nil.. That is counterintuitive, if I get no error I expect the device to be non-nil

To each their own. Since this is internal to the provider, we can agree to disagree. 

> Where in the Hetzner or AWS code do you see a nil error return when no instance was found?

In `MigrateUID()` [AWS here](https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/aws/provider.go#L646) and [Hetzner here](https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/hetzner/provider.go#L310)

deitch:
Formatting was messed up. I fixed and repushed.

pkg/cloudprovider/provider/packet/provider.go
alvaroaleman: `MigrateUID` is a special case and only returns a nil error if `p.Get` returned `err == cloudprovidererrors.ErrInstanceNotFound`.

This is really prone to result in bugs in the future. Why would I assume that if I call `p.getPacketDevice()` and `err == nil` instance can also be nil and has to be checked separately?

alvaroaleman: Like what would you think of a rest api that returned for a `GET` call a http/200 and no body?

deitch: > Why would I assume that if I call p.getPacketDevice() and err == nil device can also be nil and has to be checked separately?
 
That is pretty common, actually. 

> Like what would you think of a rest api that returned for a GET call a http/200 and no body?
 
It depends. `GET /items/1234` _should_ return a `404` for not found, and a `200` for found. But `GET /items` or a search like `GET /items?name=jim` should return `404` only if the path `GET /items` does not exist. If none matches `name=jim`, then it should return a `200` with an empty set. That has been a longstanding debate among REST implementers/builders for about as long as Fielding's paper has been out (almost 30 years), but I always came down on that side.

I think, though, that we are spending too much time on this. It is a purely internal private function. It works, is clearly defined. Let's just leave it?

alvaroaleman:
/lgtm
/approve
/hold

so @mrIncompetent can have a look

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,459 @@
...
+	c.InstanceType, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.InstanceType)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"instanceType\" field, error = %v", err)
+	}
+	c.BillingCycle, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.BillingCycle)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"billingCycle\" field, error = %v", err)
+	}
+	if c.BillingCycle == "" {
+		c.BillingCycle = defaultBillingCycle
```
mrIncompetent: Defaulting should potentially be moved into `AddDefaults`.

Have a look at the AWS code for a reference: https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/aws/provider.go#L299

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,459 @@
...
+	}
+	for i, tag := range rawConfig.Tags {
+		tagValue, err := p.configVarResolver.GetConfigVarStringValue(tag)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to get the %d value of \"tags\" field, error = %v", i, err)
```
mrIncompetent: A potential error message would look like:
`failed to get the 0 value of "tags" field, error = something really bad happened`

Can we make this more human readable?

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,459 @@
...
+		}
+		c.Tags = append(c.Tags, tagValue)
+	}
+	for i, facility := range rawConfig.Facilities {
+		facilityValue, err := p.configVarResolver.GetConfigVarStringValue(facility)
+		if err != nil {
+			return nil, nil, nil, fmt.Errorf("failed to get the %d value of \"facilities\" field, error = %v", i, err)
```
mrIncompetent: See the comment regarding the tags

pkg/cloudprovider/provider/packet/provider.go
```diff
@@ -0,0 +1,459 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to list instance types / plans: %v", err)
+	}
+	// ensure our requested plan is in those plans
+	validPlanNames := planProp(plans, "Name")
+	if missingPlans := itemsNotInList(validPlanNames, []string{c.InstanceType}); len(missingPlans) > 0 {
+		return fmt.Errorf("unknown instance type / plan: %s, acceptable plans: %s", strings.Join(missingPlans, ","), strings.Join(validPlanNames, ","))
+	}
+
+	return nil
```
mrIncompetent: Should we also validate the BillingCycle ?

pkg/cloudprovider/provider/packet/provider.go
deitch: I originally had it in `AddDefaults()`, but that doesn't get exercised in the same way. I am fine with moving it into `AddDefaults()` _also_, and then having a single internal private function that validates and adds defaults, and that one gets called _both_ by `AddDefaults()` _and_ by the private `getConfig()`, if you prefer that. But I definitely want it here as well.

Let me know.

pkg/cloudprovider/provider/packet/provider.go
deitch: Happy to take a better suggestion. I took this straight from the Hetzner example, [here](https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/hetzner/provider.go#L85)

pkg/cloudprovider/provider/packet/provider.go
deitch: Same as the above. Happy to take a suggestion.

pkg/cloudprovider/provider/packet/provider.go
deitch: I am good for that. The default, obviously, works, as we put in a sane default, but good to validate. Will be updated in 5 mins.

deitch: Ah, I remember now why I did it. There isn't an API to validate a billing cycle. You put it in the `Create()` request, and if it isn't valid, it rejects it.

pkg/cloudprovider/provider/packet/provider.go
mrIncompetent: Having it both places sounds good :+1:

pkg/cloudprovider/provider/packet/provider.go
mrIncompetent: What do you think about `failed to load the value for the Tag at index %d, error = %v` ?

pkg/cloudprovider/provider/packet/provider.go
deitch: Was a bit of a pain, but got it in. I don't love the duplication of code, but until this becomes easier, as you mentioned in the Slack, this will do.

pkg/cloudprovider/provider/packet/provider.go
deitch: Did something almost like that. Should be good enough.

PR #494, opened 2019-04-15
Add CPU and memory accounting cgroup settings for CoreOS and Ubuntu

### LINKED ISSUE #491: Check if cgroup fix is needed for CoreOS and Ubuntu too
The documented OS in https://github.com/kubernetes/kubernetes/issues/56850 is CentOS which is fixed in the Machine Controller too by adding `/etc/systemd/system/kubelet.service.d/11-cgroups.conf` with

```
[Service]
CPUAccounting=true
MemoryAccounting=true
```

through the UserData provider.

- [x] Check if CoreOS needs the same fix
- [x] Check if Ubuntu needs the same fix
- [x] Check if it would harm to add the same fix to CoreOS and Ubuntu to keep the code base homogenous

themue:

cgroup settings are already changed for CentOS due to issue https://github.com/kubernetes/kubernetes/issues/56850. Now adding same change for CoreOS and Ubuntu to keep code base homogenous.

Changes allow the *kubelet /stats/summary API* to report the correct *systemContainer* metrics for the runtime and kubelet.

Fixes #491 

```release-note
NONE
```

### DIFF
--- a/pkg/userdata/coreos/provider.go
+++ b/pkg/userdata/coreos/provider.go
@@ -311,6 +311,15 @@ storage:
           [Service]
           Environment=DOCKER_OPTS=--storage-driver=overlay2
 
+    - path: /etc/systemd/system/kubelet.service.d/11-cgroups.conf
+      filesystem: root
+      mode: 0644
+      contents:
+        inline:
+          [Service]
+          CPUAccounting=true
+          MemoryAccounting=true
+
     - path: /opt/bin/download.sh
       filesystem: root
       mode: 0755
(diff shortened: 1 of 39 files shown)


pkg/userdata/coreos/provider.go
```diff
@@ -311,6 +311,15 @@ storage:
           [Service]
           Environment=DOCKER_OPTS=--storage-driver=overlay2
 
+    - path: /etc/systemd/system/kubelet.service.d/11-cgroups.conf
```
alvaroaleman: If we add this for all OS, can you please just change the `kubelet.service` unitfile instead of creating a drop-in?

themue: Sounds good. The original fix #476 has been a copy of the solution of https://github.com/kubernetes/kubernetes/issues/56850 . As far as my checks have shown there are no global default settings of these two switches (all commented out) as well no settings in the `kubelet.service` right now.

PR #504, opened 2019-04-21
Add swapoff to centos so kubelet starts

deitch:

Disable swap on lentos so Kubelet starts. 

We already do it for Ubuntu [here](https://github.com/kubermatic/machine-controller/blob/master/pkg/userdata/ubuntu/provider.go#L223-L225) for exactly the same reason, so we should do it for centos as well. As discussed with @mrIncompetent and @thz in Slack.

Using exact same language/code for centos as Ubuntu.

```release-note
Disable swap on centos like on ubuntu
```

### DIFF
--- a/pkg/userdata/centos/provider.go
+++ b/pkg/userdata/centos/provider.go
@@ -168,6 +168,12 @@ write_files:
     systemctl restart systemd-modules-load.service
     sysctl --system
 
+    # Make sure we always disable swap - Otherwise the kubelet won't start
+    cp /etc/fstab /etc/fstab.orig
+    cat /etc/fstab.orig | awk '$3 ~ /^swap$/ && $1 !~ /^#/ {$0="# commented out by cloudinit\n#"$0} 1' > /etc/fstab.noswap
+    mv /etc/fstab.noswap /etc/fstab
+    swapoff -a
+
     {{ if ne .CloudProvider "aws" }}
     # The normal way of setting it via cloud-init is broken:
     # https://bugs.launchpad.net/cloud-init/+bug/1662542
--- a/pkg/userdata/centos/testdata/kubelet-v1.12-vsphere.golden
+++ b/pkg/userdata/centos/testdata/kubelet-v1.12-vsphere.golden
@@ -59,6 +59,12 @@ write_files:
     systemctl restart systemd-modules-load.service
     sysctl --system
 
+    # Make sure we always disable swap - Otherwise the kubelet won't start
+    cp /etc/fstab /etc/fstab.orig
+    cat /etc/fstab.orig | awk '$3 ~ /^swap$/ && $1 !~ /^#/ {$0="# commented out by cloudinit\n#"$0} 1' > /etc/fstab.noswap
+    mv /etc/fstab.noswap /etc/fstab
+    swapoff -a
+
     
     # The normal way of setting it via cloud-init is broken:
     # https://bugs.launchpad.net/cloud-init/+bug/1662542
--- a/pkg/userdata/centos/testdata/kubelet-v1.10-aws.golden
+++ b/pkg/userdata/centos/testdata/kubelet-v1.10-aws.golden
@@ -56,6 +56,12 @@ write_files:
     systemctl restart systemd-modules-load.service
     sysctl --system
 
+    # Make sure we always disable swap - Otherwise the kubelet won't start
+    cp /etc/fstab /etc/fstab.orig
+    cat /etc/fstab.orig | awk '$3 ~ /^swap$/ && $1 !~ /^#/ {$0="# commented out by cloudinit\n#"$0} 1' > /etc/fstab.noswap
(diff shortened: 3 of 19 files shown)


alvaroaleman:
Actually this will fail tests, you need to update the fixtures with `go test. /... - update`

deitch:
Good catch. I need to update the expected output in the `centos/testdata/*.golden` files. Doing now.

deitch:
FYI, the error messages are not helpful. They tell you it is from `helper.go`, but not from `centos`, or which `.golden` file doesn't match.

deitch:
Updated.

alvaroaleman:
@deitch are you sure centos manages its swap via a systemd target?

deitch:
Sure? No. I remember it that way, but been a while since I did anything serious with centos.

deitch:
> are you sure centos manages its swap via a systemd target?

Actually, it doesn't. It uses the `systemd-fstab-generator` to generate the actual units, so does Ubuntu (well, both use `systemd`, so it makes sense). Either way, it should work.

pkg/userdata/centos/provider.go
```diff
@@ -168,6 +168,10 @@ write_files:
     systemctl restart systemd-modules-load.service
     sysctl --system
 
+    # Make sure we always disable swap - Otherwise the kubelet won't start
+    systemctl mask swap.target
```
thz: Can you mask the swap target to prevent fstab generator creating swap units?
Also you might prevent other units depending on swap.target.
I would propose to remove swap entries from fstab and keep the `swapoff -a` to spare a reboot.

deitch: To be fair, that is a valid suggestion. However, we already are doing this with Ubuntu (for better or for worse). 

I would propose that we leave this as is, bring centos in line with Ubuntu, and then have a single issue (and eventual PR) to discuss the issues there and align them.

alvaroaleman: I'd argue "its just as subpar elsewhere" is not a very good argument ;) What about instead just updating Ubuntu as well?

deitch: > argue "its just as subpar elsewhere" is not a very good argument

:-)
I wouldn't disagree. 

> What about instead just updating Ubuntu as well?

I guess we can. I am a little worried about it, but future bugs are part of the game. `sed -i` is your friend...

deitch: Oh that is going to be a nasty piece of `sed` or `awk`. You need to comment out (but not delete) any lines that contain `swap` in the third column, while not double-commenting out things.

alvaroaleman: Why comment out and not just remove?

deitch: Two reasons:

1. It isn't really any easier to remove than to comment. You still want to filter only on the 3rd column precisely equal to `swap`.
2. You want to leave it as close to the original as possible, so that if anyone needs to analyze it, you have some form of trail. An SRE can sit them completely confused as to how the template `fstab` is different on this class of instances from every other, and "what in the world changed this?!?". With a comment, it is as close as possible - the original text is there, but commented out - and you can add a comment indicating what commented it.

deitch: Look at latest updated.

thz: `noauto` is an alternative to commenting out.

deitch: Yes, good point. I think I prefer the explicit comment out though. We aren't trying to say, "don't automatically mount this swap on boot", we are trying to say, "don't ever mount swap on this host."

pkg/userdata/centos/testdata/kubelet-v1.12-vsphere.golden
```diff
@@ -59,6 +59,11 @@ write_files:
     systemctl restart systemd-modules-load.service
     sysctl --system
 
+    # Make sure we always disable swap - Otherwise the kubelet won't start
+    mv /etc/fstab /etc/fstab.orig
+    cat /etc/fstab.orig | awk '$3 ~ /^swap$/ && $1 !~ /^#/ {$0="# commented out by cloudinit\n#"$0} 1' > /etc/fstab
```
thz: Can we please create a /etc/fstab.noswap first and then mv it atomically onto /etc/fstab?
If anything breaks after your first mv you'll have a system without any /etc/fstab

deitch: Yes, I like that idea. Let's do that. Update coming in 5 mins...

deitch: Updated.

pkg/userdata/centos/provider.go
deitch: Updated.

deitch:
Back in for review.

PR #508, opened 2019-04-25
add timeout for node eviction to skip evictions which won't succeed

### LINKED ISSUE #481: Draining of nodes should be killed after a certain time
Right now the machine-controller drains nodes before deleting them.
If a node has pods which cannot be drained (Container runtime has issues) the machine-controller won't delete that node.

To avoid keeping broken nodes, the drain process should be limited in time.
When the drain process reaches the timeout, the drain should be considered succeeded.

mrIncompetent:
This adds a flag to the machine-controller to skip the eviction for machines which have a deletion timestamp which is older than 2 hours.
The duration can be configured via flag.

Fixes #481

```release-note
Add 2 hour timeout to the node eviction
```


alvaroaleman:
/lgtm
/approve

Tests wont pass until #509 is in

PR #513, opened 2019-04-30
Preparation for yamllint integration in infra

### LINKED ISSUE #490: Linting for YAML files
Right now the output of the YAML templates inside the UserData providers is done by comparing the template output agains test data files. In case of illegal YAML in both, e.g. by copy and paste, the unit tests don't detect this failure.

- [x] Check YAML linters
- [x] Integrate lint run of test data before according test of the template
- [ ] Failing lint documents ➡️ failure in test data
- [ ] Working lint but failing comparison documents ➡️ failure in template

themue:

Based on issue #490 `infra` will be extended by `yamllint` to check the YAML files used in the userdata golden tests. So for this PR

- Renamed all `*.golden` to `*.yaml`
- Tested all YAML files with `yamllint` on command line and removed errors (only warnings remain)
- Added cleaning of lines with only tabs and spaces as content
- Fixed minor template related issues

Now `yamllint` and userdata unit tests work.

Fixes #490 (only here, according branch in infra has to be tested)

```release-note
NONE
```


alvaroaleman:
@themue Could you actually create the job as optional on the infra repo already (`optional: true`) and explicitly call it here via

```
/test jobname
```
so we can verify this works as expected?

themue:
@alvaroaleman Sure, sounds good. Do we have some kind of automated full tests from time to time including all optional parts? Simply to make sure that this, and other, optional tests aren't forgotten over time and get the same troubles again.

alvaroaleman:
> @alvaroaleman Sure, sounds good. Do we have some kind of automated full tests from time to time including all optional parts? Simply to make sure that this, and other, optional tests aren't forgotten over time and get the same troubles again.

No, optional tests are optional, as the name suggests. Its supposed to be made mandatory after this PR merges

pkg/userdata/centos/provider.go
```diff
@@ -121,7 +122,8 @@ func (p Provider) UserData(
 	if err != nil {
 		return "", fmt.Errorf("failed to execute user-data template: %v", err)
 	}
-	return b.String(), nil
+	clean := regexp.MustCompile(`(?m)^[ \t]+$`).ReplaceAllString(b.String(), "")
```
mrIncompetent: 1. Won't this also break data which requires tabs ?

If not, add a comment explaining why its necessary

pkg/userdata/coreos/provider.go
```diff
@@ -112,13 +113,12 @@ func (p Provider) UserData(
 	if err != nil {
 		return "", fmt.Errorf("failed to execute user-data template: %v", err)
 	}
-
-	return b.String(), nil
+	clean := regexp.MustCompile(`(?m)^[ \t]+$`).ReplaceAllString(b.String(), "")
```
mrIncompetent: Same as https://github.com/kubermatic/machine-controller/pull/513/files#r280372518

On top: If that is necessary & we use it in multiple places, should we extract it into a dedicated function to avoid the duplication?

pkg/userdata/ubuntu/provider.go
```diff
@@ -122,8 +123,8 @@ func (p Provider) UserData(
 	if err != nil {
 		return "", fmt.Errorf("failed to execute user-data template: %v", err)
 	}
-
-	return b.String(), nil
+	clean := regexp.MustCompile(`(?m)^[ \t]+$`).ReplaceAllString(b.String(), "")
```
mrIncompetent: https://github.com/kubermatic/machine-controller/pull/513/files#r280372518

pkg/userdata/centos/provider.go
themue: Doesn't remove leading tabs in front of content. Only reduces lines with only tabs and/or spaces (generated by a template plugin) to an empty line.

pkg/userdata/coreos/provider.go
themue: It's only a one-liner. But yes, the function name may be better than a comment and also the function comment in implementation can be more descriptive.

PR #516, opened 2019-05-03
Simplify updateMachine and fix finalizer handling

alvaroaleman:

Fixes a bug in the finalizer handling ans simplifies `updateMachine`. Also makes `updateMachine` able to handle updates without changes, which can happen during finalizer handling

Fixes #

```release-note
none
```

### DIFF
--- a/pkg/controller/machine/machine.go
+++ b/pkg/controller/machine/machine.go
@@ -307,42 +307,35 @@ func (c *Controller) getNodeByNodeRef(nodeRef *corev1.ObjectReference) (*corev1.
 }
 
 func (c *Controller) updateMachine(machine *clusterv1alpha1.Machine, modify func(*clusterv1alpha1.Machine)) (*clusterv1alpha1.Machine, error) {
-	var updatedMachine *clusterv1alpha1.Machine
-
 	// Both machine and updatedMachine can be nil later on, so we store the namespace and name here
 	namespace := machine.Namespace
 	name := machine.Name
 	err := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
-		var getErr, retryErr error
+		var err error
 
-		// Apply modifications
-		// machine could be nil here if the get after a failed update failed
-		if machine == nil {
-			machine, getErr = c.machineClient.ClusterV1alpha1().Machines(namespace).Get(name, metav1.GetOptions{})
-			if getErr != nil {
-				return getErr
-			}
+		machine, err = c.machinesLister.Machines(namespace).Get(name)
+		if err != nil {
+			return err
 		}
+		machine = machine.DeepCopy()
+
+		// Modify machine, only try to UPDATE when the modification results in a change
+		unmodifiedMachine := machine.DeepCopy()
 		modify(machine)
-
-		// Update the machine, if that fails, get the latest version from the api
-		// we deliberately try to update first via the provided machine object
-		// to not have at least one get call per update as we do a lot of updates
-		// which gets us ratelimited very quickly which then creates issues when there is an actual conflict
-		updatedMachine, retryErr = c.machineClient.ClusterV1alpha1().Machines(namespace).Update(machine)
-		if retryErr != nil {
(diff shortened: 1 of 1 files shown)


pkg/controller/machine/machine.go
```diff
@@ -307,42 +307,34 @@ func (c *Controller) getNodeByNodeRef(nodeRef *corev1.ObjectReference) (*corev1.
...
+		var err error
 
-		// Apply modifications
-		// machine could be nil here if the get after a failed update failed
-		if machine == nil {
-			machine, getErr = c.machineClient.ClusterV1alpha1().Machines(namespace).Get(name, metav1.GetOptions{})
-			if getErr != nil {
-				return getErr
-			}
+		machine, err = c.machineClient.ClusterV1alpha1().Machines(namespace).Get(name, metav1.GetOptions{})
```
alvaroaleman: The drawback of this is that we always do at least one GET, but IMHO this is negligible, compared to the simplification. Also this drawback will vanish once we switch to controller-runtime, as the clients reader interface will then be lister-backed.

pkg/controller/machine/machine.go
```diff
@@ -307,42 +307,34 @@ func (c *Controller) getNodeByNodeRef(nodeRef *corev1.ObjectReference) (*corev1.
...
+		if err != nil {
+			return err
 		}
+
+		// Modify machine, only try to UPDATE when the modification results in a change
+		unmodifiedMachine := machine.DeepCopy()
 		modify(machine)
+		if equal := equality.Semantic(unmodifiedMachine, machine); equal {
```
alvaroaleman: This can happen during finalizer handling, see below

pkg/controller/machine/machine.go
```diff
@@ -555,6 +547,7 @@ func (c *Controller) deleteCloudProviderInstance(prov cloud.Provider, machine *c
 	}
 
 	machine, err = c.updateMachine(machine, func(m *clusterv1alpha1.Machine) {
+		finalizers := sets.NewString(m.Finalizers...)
```
alvaroaleman: Re-using the `finalizers` set we created before calling the providers `Cleanup` is pretty problematic, because:

* We pass `machineCreateDeleteData` into `Cleanup` which contains a `MachineUpdater`
* This allows providers to Delete provider-specific finalizers
* If that happened, we will re-add them because we use a `finalizers` set that was created _before_ we called `Cleanup`

Since this may result in a no-op updateFunc, I have changed `updateMachine` to be able to cope with that case

pkg/controller/machine/machine.go
mrIncompetent: Why don't we use the lister?

pkg/controller/machine/machine.go
```diff
@@ -307,42 +307,34 @@ func (c *Controller) getNodeByNodeRef(nodeRef *corev1.ObjectReference) (*corev1.
...
-			}
+		machine, err = c.machineClient.ClusterV1alpha1().Machines(namespace).Get(name, metav1.GetOptions{})
+		if err != nil {
+			return err
 		}
+
+		// Modify machine, only try to UPDATE when the modification results in a change
+		unmodifiedMachine := machine.DeepCopy()
 		modify(machine)
+		if equal := equality.Semantic.DeepEqual(unmodifiedMachine, machine); equal {
```
mrIncompetent:
```go
if equality.Semantic.DeepEqual(unmodifiedMachine, machine) {
```

alvaroaleman:
@mrIncompetent Comment addressed, PTAL

mrIncompetent:
@alvaroaleman https://github.com/kubermatic/machine-controller/pull/516#discussion_r280673543

pkg/controller/machine/machine.go
alvaroaleman: Oh wow, yeah ofc. That one was too obvious.

alvaroaleman:
This time I actually addressed all comments, PTAL

PR #517, opened 2019-05-03
Make vsphere root disk size extendable

### LINKED ISSUE #376: Make root disk size of vsphere instances configurable
As a user of the machine-controller I want to be able to configure the size of a vsphere root disk.

*AC*

* The vsphere root disk size can be configured
* There is an e2e test in place that tests this functionality

alvaroaleman:

Makes the vsphere root disk size extendable. 
Fixes #376 

```release-note
It is now possible to configure the root disk size of vSphere machines
```

/hold
Until vsphere e2e tests are re-activated and I can write an e2e test for this

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/helper.go
+++ b/pkg/cloudprovider/provider/vsphere/helper.go
@@ -23,6 +23,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"os/exec"
 	"text/template"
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -78,20 +75,6 @@ func createClonedVM(ctx context.Context, vmName string, config *Config, dc *obje
 		targetVMFolder = datacenterFolders.VmFolder
 	}
 
-	// Create snapshot of the template VM if not already snapshotted.
-	snapshot, err := findSnapshot(ctx, templateVM, snapshotName)
```
alvaroaleman: For some reason I don't know the disk resizing silently doesn't happen when cloning a snapshot.

Since the only advantage of the snapshot cloning is that theoretically the templaveVM could be powered on, its IMHO fine to get rid of all that code and directly clone the templateVM

alvaroaleman:
I've manually verified this works :)

/assign @mrIncompetent

PR #518, opened 2019-05-03
Re-Activate vsphere tests

alvaroaleman:
Re-enables the vsphere e2e tests :)

Fixes kubermatic/kubermatic#3398

```release-note
none
```


pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -398,6 +405,24 @@ func (p *provider) Create(machine *v1alpha1.Machine, _ *cloud.MachineCreateDelet
 		return nil, fmt.Errorf("error when waiting for vm powerOn task: %v", err)
 	}
 
+	// Add a custom field to indicate to our Get that creation succeeded
+	// If the field is not set, Get will delete the instance
```
alvaroaleman: The creation can fail at various stages and we may not have a reference to the cloned VM even if it gets created, e.G. when the clone itself errord or timed out. This is the best I could think of to make sure either everything suceeds or we retry. LMK what you think

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -531,19 +560,74 @@ func (p *provider) Get(machine *v1alpha1.Machine) (instance.Instance, error) {
...
+	// Check if the creationCompleteFieldName is set to machine.Spec.Name.
+	// If that is not the case, the creation didn't complete successfully and
+	// we must delete the instance so it gets recreated
+	fieldValue, err := getValueForField(ctx, virtualMachine, creationCompleteFieldName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get value for field: %v", err)
+	}
+	if fieldValue != machine.Spec.Name {
+		// TODO: This should leverage .Cleanup, but we can currently not do that,
+		// because .Cleanup needs cloud.MachineCreateDeleteData which we don't have here
```
mrIncompetent: We should at least extract the deletion logic interaction into a dedicated function so we reuse it

alvaroaleman: we need to make the `detach PVs` part configurable then, because we dont have a client to get PVs in the `Get`

PR #519, opened 2019-05-05
Default AWS ami in AddDefaults instead of Create

alvaroaleman:

Sets the default AWS AMI in `AddDefaults` instead of `Create`. This has two advantages:

* Reproducibility: All machines for a given `MachineDeployment` will always use the same AMI
* Insigts: It is possible to find out which AMI was used if it got defaulted

Fixes #

```release-note
none
```


alvaroaleman:
Hm. AWS api seems to slow to do this in the webhook :(

PR #523, opened 2019-05-06
Allow to specify projectID using PACKET_PROJECT_ID

nikhita:
/assign @alvaroaleman 

```release-note
NONE
```


nikhita:
/hold

While we should be able to specify projectID using this env var (we do that for our tests), tests seem to be failing in the kubermatic PR even with this change. I will dig in later about this.

nikhita:
/hold cancel

Just updating vendor directory won't help because the image will still point to an older tag/commit.

PR #525, opened 2019-05-07
Fix aws filters to include productCode for CentOS

alvaroaleman:

Fixes #

```release-note
Fixed an issue with the image discovery for CentOS on AWS
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -95,7 +95,8 @@ var (
 		providerconfig.OperatingSystemCentOS: {
 			description: "CentOS Linux 7 x86_64 HVM EBS*",
 			// The AWS marketplace ID from AWS
-			owner: "679593333241",
+			owner:       "679593333241",
+			productCode: "aw0evgkw8e5c1q413zgy5pjce",
 		},
 		providerconfig.OperatingSystemUbuntu: {
 			// Be as precise as possible - otherwise we might get a nightly dev build
@@ -154,6 +155,7 @@ type Config struct {
 type amiFilter struct {
 	description string
 	owner       string
+	productCode string
 }
 
 func getDefaultAMIID(client *ec2.EC2, os providerconfig.OperatingSystem, region string) (string, error) {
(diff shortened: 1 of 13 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -95,7 +95,8 @@ var (
 		providerconfig.OperatingSystemCentOS: {
 			description: "CentOS Linux 7 x86_64 HVM EBS*",
 			// The AWS marketplace ID from AWS
-			owner: "679593333241",
+			owner:       "679593333241",
```
thz: is it necessary to use the id instead of "aws-marketplace"? the latter would be better readable.

alvaroaleman: Well, for CoreOS and Ubuntu the owner ids are actually something else than the marketplace. We could probably use both for readability? I am not sure if the name is unique

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -187,8 +189,21 @@ func getDefaultAMIID(client *ec2.EC2, os providerconfig.OperatingSystem, region
 				Name:   aws.String("root-device-type"),
 				Values: aws.StringSlice([]string{"ebs"}),
 			},
+			{
+				Name:   aws.String("architecture"),
+				Values: aws.StringSlice([]string{"x86_64"}),
```
thz: would it make sense to move the arch decision up to the caller and not hardcode it in getDefaultAMIID?

pkg/cloudprovider/provider/aws/provider.go
thz: I just want to make sure the owner is immediately understood. and "aws-marketplace" is less arbitrary id.

thz: Oh well - there is the comment above. So my comment is just about cosmetics.

pkg/cloudprovider/provider/aws/provider.go
alvaroaleman: The filters are just as hardcoded and right now we support amd64 only

thz: Just figured that there is no arm offering for it in marketplace anyways.

PR #526, opened 2019-05-07
vSphere VM cleanup after ISO failure

### LINKED ISSUE #377: Vsphere: Instance does not get cleaned up when ISO generation or attaching fails
Vsphere creates and uploads an ISO file for Ubuntu and CentOS.

If the creation or the upload or the iso fails, we are in a half-created state from which we can not recover. Instead we should cleanup, so another Create will be attempted.

themue:

In case `uploadAndAttachISO()` fails the already created VM gets destroyed.

Fixes #377 

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/provider.go
+++ b/pkg/cloudprovider/provider/vsphere/provider.go
@@ -385,6 +385,9 @@ func (p *provider) Create(machine *v1alpha1.Machine, _ *cloud.MachineCreateDelet
 		}()
 
 		if err := uploadAndAttachISO(ctx, finder, virtualMachine, localUserdataIsoFilePath, config.Datastore); err != nil {
+			if osErr := os.Remove(localUserdataIsoFilePath); osErr != nil {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -385,6 +385,14 @@ func (p *provider) Create(machine *v1alpha1.Machine, _ *cloud.MachineCreateDelet
 		}()
 
 		if err := uploadAndAttachISO(ctx, finder, virtualMachine, localUserdataIsoFilePath, config.Datastore); err != nil {
+			// Destroy VM to avoid a leftover.
+			destroyTask, err := virtualMachine.Destroy(ctx)
+			if err != nil {
+				return nil, fmt.Errorf("failed to destroy vm %s: %v", virtualMachine.Name(), err)
```
alvaroaleman: Can you for both possible errors please also add the info that the uploadAndAttachIso failed and the error that happened there?

PR #528, opened 2019-05-09
Expand CloudConfig for GCE

nikhita:
Ref: https://github.com/kubermatic/kubermatic/pull/3350#discussion_r281632201

/cc @alvaroaleman @themue 
/assign @alvaroaleman 

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/gce/cloudconfig_test.go
+++ b/pkg/cloudprovider/provider/gce/cloudconfig_test.go
@@ -38,11 +38,15 @@ func TestCloudConfigAsString(t *testing.T) {
 				Global: GlobalOpts{
 					ProjectID: "my-project-id",
 					LocalZone: "my-zone",
+					MultiZone: true,
+					Regional:  true,
 				},
 			},
 			contents: "[global]\n" +
 				"project-id = \"my-project-id\"\n" +
-				"local-zone = \"my-zone\"\n",
+				"local-zone = \"my-zone\"\n" +
+				"multizone = true\n" +
+				"regional = true\n",
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/gce/cloudconfig_test.go
```diff
@@ -38,11 +38,15 @@ func TestCloudConfigAsString(t *testing.T) {
...
+					MultiZone: true,
+					Regional:  true,
 				},
 			},
 			contents: "[global]\n" +
 				"project-id = \"my-project-id\"\n" +
-				"local-zone = \"my-zone\"\n",
+				"local-zone = \"my-zone\"\n" +
+				"multizone = true\n" +
+				"regional = true\n",
```
alvaroaleman: Do we need both of these? What's the difference between multi-zone and regional?

nikhita: > What's the difference between multi-zone and regional?

Referring to the [docs](https://cloud.google.com/kubernetes-engine/docs/how-to/creating-a-cluster#create_multi-zone):

- multizone clusters: "A multi-zone cluster runs nodes in multiple zones within the same region. All nodes in a single-zone or multi-zone cluster are controlled by the same cluster master.
- regional clusters: "Regional clusters distribute Kubernetes resources across multiple zones within a region. Regional cluster create three cluster masters across three zones and, by default, create nodes in three zones, or in as many zones as desired."

> Do we need both of these?

And from the code in Kubernetes, if [either multizone or regional is enabled](https://github.com/kubernetes/kubernetes/blob/014a2d2552ba36135dbb292d6ba0831b8023524a/pkg/cloudprovider/providers/gce/gce.go#L336-L340), `managedZones` uses all zones. Multizone is mainly used only for this purpose, but `regional` is used a [config](https://github.com/kubernetes/kubernetes/blob/014a2d2552ba36135dbb292d6ba0831b8023524a/pkg/cloudprovider/providers/gce/gce.go#L516) for the gce instance interface.

I think we can remove multizone...but I'd like more eyes on this for confirming.

alvaroaleman: Actually thinking about it, adding the option to the config struct wont harm, even if we end up not using it

/lgtm
/approve

PR #530, opened 2019-05-10
Add TokenURL in GCE CloudConfig

nikhita:
Ref: https://github.com/kubernetes/kubernetes/blob/release-1.14/pkg/cloudprovider/providers/gce/gce.go#L305

I'm not adding it to the CloudProviderConfig because I intend to default it to `nil` in Kubermatic.

/cc @alvaroaleman @mrIncompetent 

```release-note
NONE
```


nikhita:
/test pull-machine-controller-e2e-deployment-upgrade

> Job pod was evicted by the cluster.

Weird.

PR #531, opened 2019-05-13
VSphere: Use the machine UID for identification instead of name

### LINKED ISSUE #379: [vSphere] 2 clusters having a node with the same name, leads to concurrency issues
On all providers the controller uses the machine UID as tag/label on the cloud provider instance to securely identify the correct instance.
On vSphere there's no such mechanism, thus having 2 clusters with each 1 node called `node-1` will lead to issues in the controller.

alvaroaleman:

Fixes #379

```release-note
none
```

/assign  @mrIncompetent

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/provider.go
+++ b/pkg/cloudprovider/provider/vsphere/provider.go
@@ -53,6 +53,9 @@ const (
 	// the machine to indicate creation succeeded.
 	// If this is not set correctly, .Get will delete the instance
 	creationCompleteFieldName = "kubernetes-worker-complete"
+	// machineUIDFieldName is the name of the field in which we
+	// store the machines UID
+	machineUIDFieldName = "kubernetes-machine-uid"
 )
 
 type provider struct {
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -773,3 +771,82 @@ func (p *provider) MachineMetricsLabels(machine *v1alpha1.Machine) (map[string]s
 func (p *provider) SetMetricsForMachines(machines v1alpha1.MachineList) error {
 	return nil
 }
+
+func get(ctx context.Context, machine *v1alpha1.Machine, datacenterFinder *find.Finder, config *Config, operatingSystem providerconfig.OperatingSystem) (*object.VirtualMachine, error) {
```
mrIncompetent: The function name does not really match its content :confused: 

How about extracting the "cleanup" code here into a `DeleteBrokenInstances` function which takes `virtualMachineList` as argument ?

Something like:
- fetch all vm's
- cleanup broken instances
- filter based on the VM fields

I would really like to split the retrieval & cleanup code as much as possible to avoid misunderstandings when reading the code.

alvaroaleman: > https://github.com/kubermatic/kubermatic/pull/3484

How would you call it?

> How about extracting the "cleanup" code here into a DeleteBrokenInstances function which takes virtualMachineList as argument

I would like to first change the `cloudprovider` interface so that the `Get` also gets the `MachineCreateDeleteData`, because then we call call `Cleanup` in here. Any objections to that?

mrIncompetent: > How would you call it?

For the time being:
`getAndCleanup` ?

> I would like to first change the cloudprovider interface so that the Get also gets the MachineCreateDeleteData, because then we call call Cleanup in here. Any objections to that?

sgtm

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -590,74 +607,51 @@ func (p *provider) Get(machine *v1alpha1.Machine, _ *cloudprovidertypes.Provider
 	if err != nil {
 		return nil, fmt.Errorf("failed to get datacenter finder: %v", err)
 	}
-	virtualMachine, err := finder.VirtualMachine(ctx, machine.Spec.Name)
+	virtualMachineList, err := finder.VirtualMachineList(ctx, machine.Spec.Name)
```
alvaroaleman: We can not re-use `p.Get` here because it just returns a single machine and doesn't care about the creation complete part, but we need all machines here and we need to destroy the ones whose creation didn't complete

alvaroaleman:
@mrIncompetent This is ready for review again now

PR #532, opened 2019-05-14
Openstack validation: Force passing in image and flavor

alvaroaleman:

Fixes the Openstack validation to now allow an empty image and flavor. It is possible that there is an image or flavor without a name in the openstack API which will then be used.

/assign @nikhita 

Fixes #

```release-note
none
```


alvaroaleman:
I don't think there is a valid reason and will lead to an expected result if someone uses an unnamed image or flavor...

/hold cancel

PR #533, opened 2019-05-14
GCE public IP by default

kron4eg:
This PR fixes breaking change introduced in #510 

```release-note
GCE public IP by default
```

### DIFF
--- a/pkg/cloudprovider/provider/gce/config.go
+++ b/pkg/cloudprovider/provider/gce/config.go
@@ -214,12 +214,12 @@ func newConfig(resolver *providerconfig.ConfigVarResolver, spec v1alpha1.Provide
 
 	cfg.multizone, err = resolver.GetConfigVarBoolValue(cpSpec.MultiZone)
 	if err != nil {
-		return nil, fmt.Errorf("failed to retrieve assignPublicIPAddress: %v", err)
+		return nil, fmt.Errorf("failed to retrieve multizone: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/gce/config.go
```diff
@@ -202,19 +202,24 @@ func newConfig(resolver *providerconfig.ConfigVarResolver, spec v1alpha1.Provide
 		return nil, fmt.Errorf("cannot retrieve preemptible: %v", err)
 	}
 
-	cfg.assignPublicIPAddress, err = resolver.GetConfigVarBoolValue(cpSpec.AssignPublicIPAddress)
-	if err != nil {
-		return nil, fmt.Errorf("failed to retrieve assignPublicIPAddress: %v", err)
+	// make it true by default
+	cfg.assignPublicIPAddress = true
+
+	if cpSpec.AssignPublicIPAddress != nil {
```
alvaroaleman: Actually its way simpler to just set the default as you did by a `cfg.assignPublicIPAddress = true`, there is no reason to have AssignPublicIPAddress be a `*providerconfig.ConfigVarBool` instead of a literal

kron4eg: isn't `GetConfigVarBoolValue()` would return `(false, error)` in case when there is no `assignPublicIPAddress` and effectively overwrite `cfg.assignPublicIPAddress` to false?

alvaroaleman: Good point.

/lgtm
/approve

PR #542, opened 2019-05-20
Do not try to evict node when no available target is present

### LINKED ISSUE #541: Node eviction waits until timeout when all nodes in status deletion
Since our latest update our clusters hang in deletion until the node eviction timeout is reached. This happens since the machine controller tries to evict all nodes and does not find a suitable target. The machine Controller should check if there are suitable targets to evict onto in the `shouldEvict()` check, to not block deletion if the whole cluster is about to be deleted.

cbeneke:
This prevents the machineController from evicting nodes when no
other undeleted (deletionTimestamp == nil) node is present.

When the machine controller gets requested to delete all nodes it tries to evict all nodes without a suitable target. This blocks the deletion of all nodes until the default 2h timeout.
This PR adds a check to shouldEvict(), if there are machines in non-deleting status (deletionTimestamp == nil) and does not evict otherwise.

Fixes #541 

Maybe this should be flagged with an option to turn off, if the cluster e.g. has non-machine controller managed nodes (?).

```release-note
Machines are no longer evicted when all machines are marked for deletion
```

### DIFF
--- a/pkg/controller/machine/machine.go
+++ b/pkg/controller/machine/machine.go
@@ -477,7 +477,7 @@ func (c *Controller) ensureMachineHasNodeReadyCondition(machine *clusterv1alpha1
 	})
 }
 
-// evictIfNecessary checks if the machine has a node and evicts it if necessary
+// shouldEvict checks if the machine has a node and needs to be evicted before deletion
 func (c *Controller) shouldEvict(machine *clusterv1alpha1.Machine) (bool, error) {
 	// If the deletion got triggered a few hours ago, skip eviction.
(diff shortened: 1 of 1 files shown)


alvaroaleman:
@mrIncompetent and me discussed this, the approach we want to take is to trigger this when:

* All machines have a valid NodeRef to avoid the case that a machine is in the process of being provisioned while we do this
* There are no nodes with `.Spec.Unschedulable = false` (AKA a node that is probably managed outside of the machine-controller and could serve as a scheduling target)

alvaroaleman:
This was resolved in #562, closing
/close

PR #552, opened 2019-05-31
vSphere: add missing fields in example

kron4eg:
Signed-off-by: Artiom Diomin <kron82@gmail.com>

Example didn't reflected all the possible config option accessible in vSphere

Fixes #

```release-note
vSphere: add missing fields in example
```

### DIFF
--- a/examples/vsphere-machinedeployment.yaml
+++ b/examples/vsphere-machinedeployment.yaml
@@ -37,20 +37,28 @@ spec:
             - "<< YOUR_PUBLIC_KEY >>"
           cloudProvider: "vsphere"
           cloudProviderSpec:
-            templateVMName: ubuntu-template
             # Can also be set via the env var 'VSPHERE_USERNAME' on the machine-controller
             username: '<< VSPHERE_USERNAME >>'
             # Can also be set via the env var 'VSPHERE_ADDRESS' on the machine-controller
             # example: 'https://your-vcenter:8443'. '/sdk' gets appended automatically
             vsphereURL: '<< VSPHERE_ADDRESS >>'
-            datacenter: 'Datacenter'
             # Can also be set via the env var 'VSPHERE_PASSWORD' on the machine-controller
             password:
               secretKeyRef:
                 namespace: kube-system
                 name: machine-controller-vsphere
                 key: password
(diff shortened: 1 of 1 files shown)


examples/vsphere-machinedeployment.yaml
```diff
@@ -37,20 +37,25 @@ spec:
...
             password:
               secretKeyRef:
                 namespace: kube-system
                 name: machine-controller-vsphere
                 key: password
-            cluster: "test-cluster"
+            datacenter: datacenter1
+            templateVMName: ubuntu-template
+            # Optional
+            templateNetName: ''
```
alvaroaleman: Could you add an explanation, that `templateNetName` is the name of the network the interface on the template-machine is attached to that should be used for the `vmNetName` and also that its only required if the template-machine has more than one network?

examples/vsphere-machinedeployment.yaml
```diff
@@ -37,20 +37,25 @@ spec:
...
+            vmNetName: network1
```
alvaroaleman: This is also optional

examples/vsphere-machinedeployment.yaml
```diff
@@ -37,20 +37,25 @@ spec:
...
+            # Optional
+            folder: folder1
```
alvaroaleman: Are you sure, does it work when this is unset?

kron4eg: Quite sure, in kubeone example I didn't used it and it's all good: https://github.com/kubermatic/kubeone/blob/master/examples/terraform/vsphere/outputs.tf#L57

examples/vsphere-machinedeployment.yaml
kron4eg: done

examples/vsphere-machinedeployment.yaml
kron4eg: done

examples/vsphere-machinedeployment.yaml
kron4eg: done

kron4eg:
@alvaroaleman we can /override it

PR #559, opened 2019-06-04
Add proxy support

mrIncompetent:
This PR adds support for custom http proxies & custom image registries.
Both flags can be used independently.

The PR has been manually tested with CoreOS & Ubuntu - We don't have a test env for CentOS atm.

```release-note
Add support for custom http proxies & images registries
```


docs/network-restrictions.md
```diff
@@ -0,0 +1,34 @@
...
+
+If nodes only have access via a HTTP proxy, you can let the machine-controller configure all new nodes to use this proxy.
+For this the following flag must be set on the machine-controller side:
+```bash
+-node-http-proxy="http://192.168.1.1:3128"
+```
+This will set the following environment variables via /etc/environment on all nodes (lower & uppercase):
+- HTTP_PROXY
+- HTTPS_PROXY
+- NO_PROXY (Will always be set to `localhost,127.0.0.1`)
```
alvaroaleman: I think being able to add hosts here will be required in most envs. We could probably also have a smarter default, e.G: all rfc1918 ranges and the link-local ranges. WDYT @thz?

pkg/userdata/centos/provider.go
```diff
@@ -277,6 +290,19 @@ write_files:
   content: |
 {{ containerRuntimeHealthCheckSystemdUnit | indent 4 }}
 
+{{- if .NodeImageRegistry }}
+- path: /run/containers/registries.conf
+  permissions: "0644"
+  content: |
+    INSECURE_REGISTRY="--insecure-registry {{ .NodeImageRegistry }}"
```
alvaroaleman: Uhm. That this configures the registry as insecure should be at the very least documented and was not all obvious to me when reading the readme.

pkg/userdata/centos/provider.go
```diff
@@ -277,6 +290,19 @@ write_files:
...
+{{- end}}
+
+- path: /etc/systemd/system/docker.service.d/environment.conf
```
alvaroaleman: Can we avoid rendering this when no proxy is set?

alvaroaleman: You can set `EnvironmentFile=-/path/to/file`, then systemd will start the unit even when the env file is absent

pkg/userdata/centos/provider.go
mrIncompetent: Done with https://github.com/kubermatic/machine-controller/pull/559/commits/afd9104ec6e27674aaf8885e1a8e107b8c309e48

pkg/userdata/centos/provider.go
mrIncompetent: Done with https://github.com/kubermatic/machine-controller/pull/559/commits/afd9104ec6e27674aaf8885e1a8e107b8c309e48

docs/network-restrictions.md
kron4eg: Indeed, we already have examples of using NO_PROXY

cmd/controller/main.go
```diff
@@ -192,6 +199,8 @@ func main() {
 	flag.BoolVar(&profiling, "enable-profiling", false, "when set, enables the endpoints on the http server under /debug/pprof/")
 	flag.BoolVar(&externalCloudProvider, "external-cloud-provider", false, "when set, kubelets will receive --cloud-provider=external flag")
 	flag.DurationVar(&skipEvictionAfter, "skip-eviction-after", 2*time.Hour, "Skips the eviction if a machine is not gone after the specified duration.")
+	flag.StringVar(&nodeHttpProxy, "node-http-proxy", "", "If set, this proxy will be configured on all nodes.")
+	flag.StringVar(&nodeImageRegistry, "node-image-registry", "", "If set, this image registry will be used for pulling all required images on the node.")
```
alvaroaleman: Can we rename this to `insecure-registry`? That is what it is after all, it doesn't change any behaviour about what gets pulled from which location, it just adds an insecure registry. Also we might want this to be a string slice and not a string

pkg/userdata/coreos/provider.go
```diff
@@ -198,9 +214,13 @@ systemd:
         TimeoutStartSec=5min
         CPUAccounting=true
         MemoryAccounting=true
+{{- if .NodeHttpProxy }}
+        Environment=KUBELET_IMAGE=docker://{{ .NodeImageRegistry }}/machine-controller/hyperkube-amd64:{{ .HyperkubeImageTag }}
```
alvaroaleman: Instead of re-using this magic variable which other than that adds an insecure-registry, can we add an explicit config option for the registry+namespace+name of the hyperkube image in the CoreOS option?

docs/network-restrictions.md
mrIncompetent: @kron4eg Do you have a link to such an example ?

cmd/controller/main.go
mrIncompetent: If set, it changes the source for the pause & hyperkube container.

alvaroaleman: Similiarily to hyperkube container, pause container should be a dedicated option, probably a flag?

Right now we make the assumptions that:

* User has a registry and its insecure
* User puts hyperkube image into that registry instead of pulling it from the internet (which would be possible, if we did set the proxy flags for docker)
* User puts pause image into that registry instead of pulling it from the internet (which again could be done by setting up the proxy config for Docker)

pkg/userdata/coreos/provider.go
mrIncompetent: I'll add a flag to pass through the image for hyperkube & the pause container

docs/network-restrictions.md
mrIncompetent: @kron4eg Maybe just post the potential content of that field instead of a link ?

thz: Having all rfc1918 ranges in NO_PROXY by default is perhaps a bit too much. But 127.0.0.1 and link-local makes sense. Perhaps also the k8s cluster domain(s).

cmd/controller/main.go
```diff
@@ -192,12 +199,17 @@ func main() {
 	flag.BoolVar(&profiling, "enable-profiling", false, "when set, enables the endpoints on the http server under /debug/pprof/")
 	flag.BoolVar(&externalCloudProvider, "external-cloud-provider", false, "when set, kubelets will receive --cloud-provider=external flag")
 	flag.DurationVar(&skipEvictionAfter, "skip-eviction-after", 2*time.Hour, "Skips the eviction if a machine is not gone after the specified duration.")
+	flag.StringVar(&nodeHTTPProxy, "node-http-proxy", "", "If set, this proxy will be configured on all nodes.")
+	flag.StringVar(&nodeNoProxy, "node-no-proxy", ".svc,.cluster.local,localhost,127.0.0.1,127.0.0.0/", "If set, this proxy will be configured on all nodes.")
```
mrIncompetent: @thz PTAL

pkg/userdata/coreos/provider.go
mrIncompetent: Done with https://github.com/kubermatic/machine-controller/pull/559/commits/649786bd2d98d7c54b81975417e047a9a654ab42

cmd/controller/main.go
mrIncompetent: Done with https://github.com/kubermatic/machine-controller/pull/559/commits/649786bd2d98d7c54b81975417e047a9a654ab42

docs/network-restrictions.md
mrIncompetent: Done with https://github.com/kubermatic/machine-controller/pull/559/commits/649786bd2d98d7c54b81975417e047a9a654ab42

alvaroaleman: @thz Do you have an example usecase where it does make sense to not have the rfc1918 networks in no_proxy?

docs/network-restrictions.md
```diff
@@ -0,0 +1,45 @@
...
+```bash
+-node-http-proxy="http://192.168.1.1:3128"
+```
+This will set the following environment variables via /etc/environment on all nodes (lower & uppercase):
+- HTTP_PROXY
+- HTTPS_PROXY
+
+`NO_PROXY` can be configured using a dedicated flag:
+```bash
+-node-no-proxy="10.0.0.0/24"
```
alvaroaleman: Is it possible to do this with CIDR ranges? I always thought no_proxy has to be a suffix which would mean this never matches as no ip address ends with `10.0.0.0/24`

pkg/userdata/centos/provider.go
```diff
@@ -40,31 +37,24 @@ import (
...
 // UserData renders user-data template to string.
-func (p Provider) UserData(
-	spec clusterv1alpha1.MachineSpec,
-	kubeconfig *clientcmdapi.Config,
-	cloudConfig string,
-	cloudProviderName string,
-	clusterDNSIPs []net.IP,
-	externalCloudProvider bool,
-) (string, error) {
+func (p Provider) UserData(req plugin.UserDataRequest) (string, error) {
```
alvaroaleman: :+1:

pkg/userdata/coreos/provider.go
```diff
@@ -200,9 +201,13 @@ systemd:
...
         MemoryAccounting=true
-        Environment=KUBELET_IMAGE=docker://k8s.gcr.io/hyperkube-amd64:{{ .HyperkubeImageTag }}
+{{- if .HTTPProxy }}
+        Environment=KUBELET_IMAGE=docker://{{ .HyperkubeImage }}:v{{ .KubeletVersion }}
+{{- else }}
+        Environment=KUBELET_IMAGE=docker://k8s.gcr.io/hyperkube-amd64:v{{ .KubeletVersion }}
+{{- end }}
         Environment="RKT_RUN_ARGS=--uuid-file-save=/var/cache/kubelet-pod.uuid \
-          --insecure-options=image \
+          --insecure-options=image{{if .InsecureHyperkubeImage }},http{{ end }} \
```
alvaroaleman: Shouldn't the condition gate the whole arg and not only the args value?

pkg/userdata/ubuntu/testdata/vsphere-proxy.yaml
```diff
@@ -0,0 +1,360 @@
...
+ssh_pwauth: no
+ssh_authorized_keys:
+- "ssh-rsa AAABBB"
+
+write_files:
+- path: "/etc/environment"
+  content: |
+    PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games"
+    HTTP_PROXY=http://192.168.100.100:3128
+    http_proxy=http://192.168.100.100:3128
```
alvaroaleman: Nit: Can be simplified as `http_proxy=$HTTP_PROXY`

mrIncompetent: Variable expansion does not work in /etc/environment

pkg/userdata/coreos/provider.go
mrIncompetent: `image` is always required as the hyperkube images in `k8s.gcr.io` are not signed & signing is not default with docker.

I'm following the docker logic here, where `insecure` means http.

docs/network-restrictions.md
mrIncompetent: Was an assumption as that's supported in the `net` package of golang: https://github.com/golang/net/blob/master/http/httpproxy/proxy.go#L39

cc @thz

cmd/controller/main.go
thz: there is a trailing slash probably meant to be `/8`.

thz: proxy authentication is handled as url-embedded credentials, right?

docs/network-restrictions.md
thz: of course. let's say you are in a 10.1.0.0/16 CIDR vpc and you want to talk to that other 10.2.0.0/16 vpc directly by aws peering.

thz: Aehm. Not _directly_ but via the proxy.

mrIncompetent:
> Is there no image available w/o the architecture hardcoding? gcr does support the manifests at least.

According to the [Makefile](https://github.com/kubernetes/kubernetes/tree/master/cluster/images/hyperkube) no.
Also that should not be scope of this PR as it does not change anything on that side.

cmd/controller/main.go
mrIncompetent: Yes

docs/network-restrictions.md
alvaroaleman: Okay, but that still doesn't work for the rest of the world, sample:

```
$ curl https://8.8.8.8
curl: (60) SSL certificate problem: self signed certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
alvaro@t470s-[2019-06-07-12:23]:~
$ export https_proxy=127.0.0.1
alvaro@t470s-[2019-06-07-12:23]:~
$ curl https://8.8.8.8
curl: (7) Failed to connect to 127.0.0.1 port 1080: Connection refused
alvaro@t470s-[2019-06-07-12:23]:~
$ export no_proxy=8.8.0.0/16
alvaro@t470s-[2019-06-07-12:24]:~
$ curl https://8.8.8.8
curl: (7) Failed to connect to 127.0.0.1 port 1080: Connection refused
```

And we use curl to download things and need it to work with a proxy there or not?

mrIncompetent: Good point :+1: 
There is no rfc on how the `NO_PROXY` variable should be used.
Though it seems everyone supports IP's & domain names.

I'll adjust the code & doc to reflect that only IP's & domain names must be specified.

mrIncompetent:
@thz @alvaroaleman PTAL

docs/network-restrictions.md
```diff
@@ -0,0 +1,47 @@
...
+`NO_PROXY` can be configured using a dedicated flag:
+```bash
+-node-no-proxy="10.0.0.1"
+```
+
+`-node-http-proxy` & `-node-no-proxy` must only contain IP addresses and/or domain names.
+
+# Using a custom image registry
+
+Except for custom workload, the kubelet requires access to the "pause" container.
```
alvaroaleman: custom workload?

mrIncompetent: Own pods you deploy into the cluster. 
I wanted to somehow express that this registry does not modify your custom pods/deployments/etc. you still need to change your manifests to point to the registry.

thz:
> /lgtm
> /approve
> /hold
> 
> in case @thz wants to have a look

/hold cancel

PR #560, opened 2019-06-04
Use omitempty for providerConfigs

alvaroaleman:

Fixes #

```release-note
none
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -113,24 +113,24 @@ var (
 )
 
 type RawConfig struct {
-	AccessKeyID     providerconfig.ConfigVarString `json:"accessKeyId"`
-	SecretAccessKey providerconfig.ConfigVarString `json:"secretAccessKey"`
+	AccessKeyID     providerconfig.ConfigVarString `json:"accessKeyId,omitempty"`
+	SecretAccessKey providerconfig.ConfigVarString `json:"secretAccessKey,omitempty"`
 
 	Region           providerconfig.ConfigVarString `json:"region"`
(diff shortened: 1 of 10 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -113,24 +113,24 @@ var (
...
 
 type RawConfig struct {
-	AccessKeyID     providerconfig.ConfigVarString `json:"accessKeyId"`
-	SecretAccessKey providerconfig.ConfigVarString `json:"secretAccessKey"`
+	AccessKeyID     providerconfig.ConfigVarString `json:"accessKeyId,omitempty"`
+	SecretAccessKey providerconfig.ConfigVarString `json:"secretAccessKey,omitempty"`
 
-	Region           providerconfig.ConfigVarString `json:"region"`
-	AvailabilityZone providerconfig.ConfigVarString `json:"availabilityZone"`
+	Region           providerconfig.ConfigVarString `json:"region,omitempty"`
```
mrIncompetent: According to https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required only optional fields should use `omitempty`.

alvaroaleman: But none of these fields is required as they can be populated via the env

mrIncompetent: Only a subset of the fields are being fetched from the env.
Region is an example

alvaroaleman: But it could still come from a configmap or secret or not?

mrIncompetent: Which means the field is not "empty"

alvaroaleman:
@mrIncompetent Updated, PTAL

PR #565, opened 2019-06-06
Add NodeTags to GCE CloudConfig

nikhita:
`NodeTags` is the list of tags to use on firewall rules for load balancers. This is necessary for LBs to work on GCP.

For https://github.com/kubermatic/kubermatic/pull/3572

Ref:

- https://github.com/kubernetes/kubernetes/blob/ef7808fec57284582c7bf3fc834570e5769df83e/staging/src/k8s.io/legacy-cloud-providers/gce/gce.go#L121

- https://github.com/kubernetes/kubernetes/blob/ef7808fec57284582c7bf3fc834570e5769df83e/staging/src/k8s.io/legacy-cloud-providers/gce/gce_loadbalancer_external.go#L979-L981

```release-note
NONE
```

/assign @alvaroaleman

### DIFF
--- a/pkg/cloudprovider/provider/gce/cloudconfig_test.go
+++ b/pkg/cloudprovider/provider/gce/cloudconfig_test.go
@@ -43,6 +43,7 @@ func TestCloudConfigAsString(t *testing.T) {
 					TokenURL:       "nil",
 					MultiZone:      true,
 					Regional:       true,
+					NodeTags:       []string{"tag1", "tag2"},
 				},
 			},
 			contents: "[global]\n" +
@@ -52,7 +53,9 @@ func TestCloudConfigAsString(t *testing.T) {
 				"subnetwork-name = \"my-cool-subnetwork\"\n" +
 				"token-url = \"nil\"\n" +
 				"multizone = true\n" +
-				"regional = true\n",
+				"regional = true\n" +
+				"node-tags = \"tag1\"\n" +
+				"node-tags = \"tag2\"\n",
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/provider/gce/cloudconfig_test.go
```diff
@@ -52,7 +53,9 @@ func TestCloudConfigAsString(t *testing.T) {
 				"subnetwork-name = \"my-cool-subnetwork\"\n" +
 				"token-url = \"nil\"\n" +
 				"multizone = true\n" +
-				"regional = true\n",
+				"regional = true\n" +
+				"node-tags[] = \"tag1\"\n" +
```
nikhita: https://www.npmjs.com/package/ini#usage has a good example for arrays :)

alvaroaleman: Ini is not standardized, do we have other occurences of slices? If not, could you verify that this actually works with the ini parser that is being used in k/k?

nikhita: Updated. Ref: `Example (Multivalue)` here - https://godoc.org/gopkg.in/gcfg.v1

```
package main

import (
	"fmt"
	gcfg "gopkg.in/gcfg.v1"
	"log"
)

func main() {
	cfgStr := `; Comment line
[section]
nodeTags="tag1"
nodeTags="tag2"`

	cfg := struct {
		Section struct {
			NodeTags []string
		}
	}{}
	err := gcfg.ReadStringInto(&cfg, cfgStr)
	if err != nil {
		log.Fatalf("Failed to parse gcfg data: %s", err)
	}
	fmt.Println(cfg.Section.NodeTags)
}
```

gives

```
[tag1 tag2]
```

alvaroaleman: /lgtm
/approve

PR #566, opened 2019-06-06
Dont set --allow-privileged on kubernetes > 1.14 as it got removed

alvaroaleman:

Fixes #

```release-note
none
```

/assign @nikhita

### DIFF
--- a/pkg/userdata/coreos/provider_test.go
+++ b/pkg/userdata/coreos/provider_test.go
@@ -264,6 +264,29 @@ func TestUserDataGeneration(t *testing.T) {
 				DisableAutoUpdate: true,
 			},
 		},
+		{
+			name: "v1.15.0-vsphere",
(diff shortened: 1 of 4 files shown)


pkg/userdata/coreos/provider_test.go
```diff
@@ -264,6 +264,29 @@ func TestUserDataGeneration(t *testing.T) {
 				DisableAutoUpdate: true,
 			},
 		},
+		{
+			name: "v1.15.0-vsphere",
+			providerSpec: &providerconfig.Config{
+				CloudProvider: "vsphere",
```
nikhita: Why aren't SSH public keys needed here? :thinking:

alvaroaleman: Well, they are not a mandatory property and I thought the fixture could be without a SSH key :)

PR #574, opened 2019-06-17
Add docker registry mirror

### LINKED ISSUE #570: Add support for docker registry-mirrors
Some environments have dedicated docker mirrors to speed up the image pull process.

Docker supports image mirrors via a Flag: `--registry-mirror` or `daemon.json` option:
```json
{
  "registry-mirrors": ["https://registry.docker-cn.com"]
}
```

This will make docker pull all images using that registry, instead of the one which gets specified via the Kubernetes manifest. Which makes this option perfect for transparent image caching!

To pass through that option to the docker daemon, the machine-controller needs a new flag, like [--node-insecure-registries](https://github.com/kubermatic/machine-controller/blob/master/cmd/controller/main.go#L204), which then gets passed through to the nodes.

themue:

Fixes #570 

```release-note
- Adds flag -node-registry-mirrors to set mirrors for Docker image registries
```

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -181,7 +181,8 @@ type controllerRunOptions struct {
 	node machinecontroller.NodeSettings
 }
 
-func main() {
+// parseFlags encapsulates the parsing of all flags.
+func parseFlags() {
 	// This is also being registered in kubevirt.io/kubevirt/pkg/kubecli/kubecli.go so
 	// we have to guard it
 	//TODO: Evaluate alternatives to importing the CLI. Generate our own client? Use a dynamic client?
@@ -210,6 +211,10 @@ func main() {
 	flag.Parse()
 	kubeconfig = flag.Lookup("kubeconfig").Value.(flag.Getter).Get().(string)
 	masterURL = flag.Lookup("master").Value.(flag.Getter).Get().(string)
+}
+
+func main() {
+	parseFlags()
 
 	clusterDNSIPs, err := parseClusterDNSIPs(clusterDNSIPs)
 	if err != nil {
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -338,15 +338,13 @@ func main() {
 	}
 
 	for _, registry := range strings.Split(nodeInsecureRegistries, ",") {
-		trimmedRegistry := strings.TrimSpace(registry)
-		if trimmedRegistry != "" {
+		if trimmedRegistry := strings.TrimSpace(registry); trimmedRegistry != "" {
 			runOptions.node.InsecureRegistries = append(runOptions.node.InsecureRegistries, trimmedRegistry)
 		}
 	}
 
 	for _, mirror := range strings.Split(nodeRegistryMirrors, ",") {
-		trimmedMirror := strings.TrimSpace(mirror)
-		if trimmedMirror != "" {
+		if trimmedMirror := strings.TrimSpace(mirror); trimmedMirror != "" {
 			runOptions.node.RegistryMirrors = append(runOptions.node.RegistryMirrors, trimmedMirror)
 		}
 	}
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -181,8 +181,7 @@ type controllerRunOptions struct {
 	node machinecontroller.NodeSettings
 }
 
-// parseFlags encapsulates the parsing of all flags.
-func parseFlags() {
+func main() {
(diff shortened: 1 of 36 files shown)


pkg/controller/machine/machine.go
```diff
@@ -107,17 +107,19 @@ type Controller struct {
...
-	// Translates to --cluster-dns on the kubelet
+	// Translates to --cluster-dns on the kubelet.
 	ClusterDNSIPs []net.IP
 	// If set, this proxy will be configured on all nodes.
 	HTTPProxy string
-	// If set this will be set as NO_PROXY on the node
+	// If set this will be set as NO_PROXY on the node.
 	NoProxy string
-	// If set, this image registry will be used for pulling all required images on the node
+	// If set, this image registry will be used for pulling all required images on the node.
```
mrIncompetent: Could you fix this comment? I screwed that up in the initial PR :(

`If set, those registries will be configured as insecure on the container runtime.`

themue: Done. Will rush in with next push.

pkg/userdata/centos/provider.go
```diff
@@ -268,11 +268,16 @@ write_files:
...
 {{ containerRuntimeHealthCheckSystemdUnit | indent 4 }}
 
-{{- if .InsecureRegistries }}
+{{- if or .InsecureRegistries .RegistryMirrors }}
 - path: /run/containers/registries.conf
   permissions: "0644"
   content: |
-    INSECURE_REGISTRY="--insecure-registry {{ .InsecureRegistries | join " --insecure-registry "}}"
+    {{- if .InsecureRegistries}}
+    INSECURE_REGISTRY="{{range .InsecureRegistries}}--insecure-registry {{.}} {{end}}"
```
alvaroaleman: Why did you change this expression? The prior one did not leave a trailing space.

cmd/controller/main.go
```diff
@@ -331,8 +338,16 @@ func main() {
 	}
 
 	for _, registry := range strings.Split(nodeInsecureRegistries, ",") {
-		if strings.TrimSpace(registry) != "" {
-			runOptions.node.InsecureRegistries = append(runOptions.node.InsecureRegistries)
+		trimmedRegistry := strings.TrimSpace(registry)
```
alvaroaleman: Nit: Mind using here and below the shorted ` if trimmedRegistry := strings.TrimSpace(registry); trimmgedRegeistry != "" {`

cmd/controller/main.go
```diff
@@ -180,7 +181,8 @@ type controllerRunOptions struct {
 	node machinecontroller.NodeSettings
 }
 
-func main() {
+// parseFlags encapsulates the parsing of all flags.
+func parseFlags() {
```
alvaroaleman: Why?

pkg/userdata/centos/provider.go
themue: Henrik asked me to change it to a loop.

cmd/controller/main.go
themue: Will change.

cmd/controller/main.go
themue: After adding only my one flag the cyclomatic complexity test failed. 🤦‍♂ Moving it out into this function made it pass again. 

IMHO the better solution would be a refactoring into a `Main` type with helping methods and a major `Run()` method called in `main()`. This would also be possible to be unit tested (needing a lot of stubbing). But that would be a different issue.

alvaroaleman: Or just remove the cyclomatic complecity linter? It doesn't care about nesting at all it just counts number of branches within a function which IMHO is a pretty useless metric.

themue: Would have no problem with it. Got some similar messages in my private code too where IMHO the splitting with the needed data structures and/or arguments would lead to an even more unmaintainable code.

But we need a general agreement then.

alvaroaleman: Can you just add an exception for this specific case so we can get this pull merged?

themue: Done.

pkg/userdata/centos/provider.go
mrIncompetent: Yes, that `strings.Join` was rather counter intuitive & meant we always set `--insecure-registry`.
That worked well as long as we only had one variable - Since we now have 2 which can be set individually that would not have worked

PR #575, opened 2019-06-17
Enable fine-grained control over coreos update.

ainmosni:
For certain setups, mainly the ones using the Container Linux Update Operator,
it's required to have the CoreOS update-engine enabled while disabling
locksmithd. Previously, you could only enable/disable these together.
Thos patch adds two options so that they can be individually tweaked
while also retaining the old behaviour.

This patch also fixes the test data, as my patch introduced some
whitespace differences.

Fixes #

```release-note
Add Container Linux specific config options: `disableLocksmithD` and `disableUpdateEngine`, to allow for more fine-grained control over the Container Linux update process.
```

### DIFF
--- a/pkg/userdata/coreos/coreos.go
+++ b/pkg/userdata/coreos/coreos.go
@@ -24,7 +24,9 @@ import (
 
 // Config contains specific configuration for CoreOS.
 type Config struct {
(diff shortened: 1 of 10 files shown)


alvaroaleman:
Thanks!

/lgtm
/approve

PR #580, opened 2019-07-01
Add Openstack tenant id

### LINKED ISSUE #568: Add support for OpenStack TENANT_ID
In some environments the tenant name is not unique and will lead to errors.
To avoid this, the tenant ID must be used.

**Acceptance criteria**:
- [ ] The machine-controller can create instances when using the tenant id instead of the name

mrIncompetent:
This PR adds support to set the TenantID for OpenStack machines.

Fixes #568

```release-note
Add support to set the TenantID for OpenStack machines.
```

/hold 
needs at least a manual test run

### DIFF
--- a/pkg/providerconfig/types.go
+++ b/pkg/providerconfig/types.go
@@ -297,10 +297,7 @@ func (configVarResolver *ConfigVarResolver) GetConfigVarStringValueOrEnv(configV
 		return cfgVar, err
 	}
 
-	envVal, envValFound := os.LookupEnv(envVarName)
-	if !envValFound {
-		return "", fmt.Errorf("all mechanisms(value, secret, configMap) of getting the value failed, including reading from environment variable = %s which was not set", envVarName)
-	}
+	envVal, _ := os.LookupEnv(envVarName)
 	return envVal, nil
 }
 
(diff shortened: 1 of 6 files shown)


alvaroaleman:
Should we add a check if the name of the tenant passed via `tenantID` matches the `tenantName`, if set?

mrIncompetent:
@alvaroaleman 

> Should we add a check if the name of the tenant passed via tenantID matches the tenantName, if set?

Those fields are being used during authentication - So i don't think its necessary to check after successful authentication if the supplied tenantName matches the tenantID.
If both fields are set, they will be used to authenticate - The OpenStack API will either accept it or not - We should not invalidate an successful authentication.

mrIncompetent:
@alvaroaleman PTAL

pkg/providerconfig/types.go
```diff
@@ -297,10 +297,7 @@ func (configVarResolver *ConfigVarResolver) GetConfigVarStringValueOrEnv(configV
 		return cfgVar, err
 	}
 
-	envVal, envValFound := os.LookupEnv(envVarName)
-	if !envValFound {
```
alvaroaleman: Why did you remove this?

mrIncompetent: Not property which gets loaded via this func is mandatory.
If we throw an error here, it is essentially mandatory.

Instead on relying on this implicit handling, we must put the validation of those fields into the corresponding "Validate" function.

Issue #545, opened 2019-05-27
Migrate machine-controller to use controller-runtime


alvaroaleman:
To align with other controllers, reduce duplicate caches, simplify the code and make testing easier, we should migrate the machine-controller to use controller-runtime.

AC:

* The  machine-controller uses controller-runtime

/kind cleanup

PR #582, opened 2019-07-01
Use dynamic client everywhere

alvaroaleman:

Changes the machine-controller to use controller-runtime everywhere, except in the node evictions, as evictions are a subresource the dynamic client cant handle.

The goal of this is to simplify the machine-controller, working towards #545 .

Fixes #

Sorry for the size of the PR :grimacing: 

/assign @mrIncompetent 

```release-note
none
```


alvaroaleman:
/hold

We should do at least one further machine-controller release before we merge this

cmd/controller/main.go
```diff
@@ -501,14 +488,12 @@ func startControllerViaLeaderElection(runOptions controllerRunOptions) error {
...
-			runOptions.nodeLister,
-			runOptions.machineInformer,
-			runOptions.machineLister,
-			runOptions.secretSystemNsLister,
+			mgr.GetClient(),
+			mgr.GetRecorder("machine-controller"),
 			runOptions.metrics,
 			runOptions.prometheusRegisterer,
+			runOptions.machineInformer,
+			runOptions.nodeInformer,
```
mrIncompetent: Why do we still need those informers?

pkg/controller/machine/machine.go
```diff
@@ -84,12 +82,9 @@ const (
 
 // Controller is the controller implementation for machine resources
 type Controller struct {
-	kubeClient    kubernetes.Interface
-	machineClient clusterv1alpha1clientset.Interface
-
-	nodesLister          listerscorev1.NodeLister
-	machinesLister       clusterlistersv1alpha1.MachineLister
-	secretSystemNsLister listerscorev1.SecretLister
+	ctx        context.Context
```
mrIncompetent: What is this context for?

cmd/controller/main.go
alvaroaleman: For triggering the reconciliation. Currently, this controller is not a controller-runtime controller, the only thing the PR changes is that the dynamic client is used everywhere (except on the eviction where that is not possible).

I was thinking this makes the reviewing easier, but I can also extend this PR to make it a controller-runtime controller. WDYT?

pkg/controller/machine/machine.go
alvaroaleman: Its used in the api calls and will get canceled when the leader lock was lost

cmd/controller/main.go
mrIncompetent: I would prefer keeping the PR size as is :grimacing:

PR #583, opened 2019-07-02
Always add instance delete finalizer initially

### LINKED ISSUE #581: Potential of orphaning instances
Because we add the `delete-instalce` finalizer after the `provider.Create` call finished, there's a chance that we might orphan instances in case the machine gets deleted while the `provider.Create` call is still processing.

Example:
- We create a Digitalocean Machine
- The controller calls `provider.Create` (This usually takes a few minutes)
- The machine object gets deleted. Since it has no finalizers it will instantly disappear
- `provider.Create` finishes
- The controller tries to add the finalizer - Which fails because the machine is already gone

Potential solution:
- Add the finalizer before calling `provider.Create`
- Remove the finalizer in case `provider.Create` fails

alvaroaleman:

To be sure we don't orphan instances

Fixes #581 

```release-note
none
```

/assign @nikhita

### DIFF
--- a/pkg/controller/machine/machine.go
+++ b/pkg/controller/machine/machine.go
@@ -351,13 +351,26 @@ func (c *Controller) updateMachineErrorIfTerminalError(machine *clusterv1alpha1.
 }
 
 func (c *Controller) createProviderInstance(prov cloudprovidertypes.Provider, machine *clusterv1alpha1.Machine, userdata string) (instance.Instance, error) {
+	// Ensure finalizer is there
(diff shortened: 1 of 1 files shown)


nikhita:
```
        helper.go:151: verify failed due to error=failed to create machine migrateuid-gce-1145995029450330113: An error of type = InvalidConfiguration, with message = Failed to insert instance: googleapi: Error 403: Quota 'IN_USE_ADDRESSES' exceeded. Limit: 30.0 in region europe-west3., quotaExceeded occurred
````

:eyes: :thinking:

PR #584, opened 2019-07-02
Make Openstack NodeVolumeAttachLimit configurable

multi-io:

This makes the Openstack `node-volume-attach-limit` cloud-config setting configurable in the openstack machine settings. If set, the value will be written to cloud-config when the node is created.

The kubelet will write this to the node object (status.allocatable.attachable-volumes-cinder), where the scheduler will use it if configured to do so ("MaxCinderVolumeCount" scheduler predicate; not enabled by default).

```release-note
Openstack node-volume-attach-limit setting configurable per machine
```


alvaroaleman:
@multi-io Can you sign your commit?

multi-io:
Updated the PR so we no longer write a default node-attachment-limit of 256 into the cloud-config if the value isn't set in the machine.

alvaroaleman:
/override pull-machine-controller-e2e-kubevirt

We are currently working on that, see #587 
/lgtm
/approve

PR #587, opened 2019-07-05
Update kubevirt to 0.18

moadqassem:
**Goal:** 
The main goal of the PR was to update the vendors of the Kubevirt package and reflect the changes on the Kubevirt provider in the machine controller. However, there were some changes some types in Kubevirt(such as the Virtual Machine Instance) and those changes need to upgrade our Kubevirt deployment files into the new Kubevirt deployment precesses(v0.17.4), which by using Kubevirt operator. 

kubermatic/kubermatic#3397

1. The `VirtualMachineInstance(vmi)` type has been changed.
2. The new Kubevirt operator and custom definition have been deployed in the e2e env.
3. ITS IMPORTANT to roll out the Kubevirt operator in production in order to enable the machine controller to create vmi otherwise the machine deployment will fail, due to the absence of the vmi `v1alpha3` which comes with the new operator the current vmi is `v1alpha2`. 
4. In order to roll out the new Kubevirt operator, the old Kubevirt installation has to be removed completely with all the `webhooks` and `apiservices`.

```release-note
https://github.com/kubevirt/kubevirt/releases/tag/v0.17.4
```


pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -83,7 +84,9 @@ func (k *kubeVirtServer) ID() string {
 func (k *kubeVirtServer) Addresses() []string {
 	var addresses []string
 	for _, kvInterface := range k.vmi.Status.Interfaces {
-		addresses = append(addresses, kvInterface.IP)
+		if address := strings.Split(kvInterface.IP, "/")[0]; address != "" {
```
alvaroaleman: This will panic if `kvInterface.IP` is an empty string or not?

alvaroaleman:
Can you extend the `.yamllint.conf` file to ignore the vendor folder by adding

```
ignore: |
    vendor/
```

It seems there are symlinks in the vendor that godep doesn't clean up and that point nowhere and yamllint is trying to follow them to non-existing target causing the failure we see

pkg/cloudprovider/provider/kubevirt/provider.go
moadqassem: The `Split` function would return a lengthed-one slice which is an empty string :)

PR #588, opened 2019-07-05
Run Docker with systemd cgroup driver

xmudrii:

This PR changes the Docker daemon manifest (`/etc/docker/daemon.json`) to run Docker with the `systemd` cgroup driver. On some platforms, such as Ubuntu and CoreOS, the default driver is `cgroupfs`. However, as those systems use `systemd`, combining `systemd` and `cgroupfs` driver can make a node unstable and eventually crash it. The reason for that is that `systemd` and `cgroupfs` can see resources differently.

More about this can be found in the [Kubernetes documentation](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#cgroup-drivers).

```release-note
Run Docker with systemd cgroup driver
```

/assign @alvaroaleman

### DIFF
--- a/pkg/userdata/centos/provider.go
+++ b/pkg/userdata/centos/provider.go
@@ -220,11 +220,6 @@ write_files:
   content: |
 {{ kubeletSystemdUnit .KubeletVersion .CloudProviderName .MachineSpec.Name .DNSIPs .ExternalCloudProvider .PauseImage | indent 4 }}
 
-- path: "/etc/systemd/system/kubelet.service.d/extras.conf"
-  content: |
-    [Service]
-    Environment="KUBELET_EXTRA_ARGS=--cgroup-driver=systemd"
-
 - path: "/etc/kubernetes/cloud-config"
   content: |
 {{ .CloudConfig | indent 4 }}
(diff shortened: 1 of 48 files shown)


xmudrii:
/test pull-machine-controller-e2e-aws
I'm not sure is this flake or not, so I'll try running this one again.

alvaroaleman:
@xmudrii CentOS passed everywhre, CoreOS and Ubuntu failed everywhere. I doubt that has something to do with our CI :)

xmudrii:
@alvaroaleman I doubt as well, but just wanted to be sure. I'll do some debugging on Monday. :)

alvaroaleman:
I think the culprit is a mising `--cgroup-driver=systemd` arg on the kubelet. We already set it for CentOS, because there we use the docker version that is part of systemd and that has the systemd cgroup driver preconfigured.

Please just remove the systemd drop-in on centos and instead set it as kubelet flag for all distributions

alvaroaleman:
/override pull-machine-controller-e2e-kubevirt 
 because WIP

/lgtm
/approve

PR #594, opened 2019-07-17
Aws optional public ip

mrIncompetent:
This PR makes the public IP assignment for AWS instances optional.

```release-note
NONE
```

### DIFF
--- a/examples/aws-machinedeployment.yaml
+++ b/examples/aws-machinedeployment.yaml
@@ -60,6 +60,8 @@ spec:
             diskType: "gp2"
             ## Only application if diskType = io1
             diskIops: 500
+            # Assign a public IP to this instance. Default: true
+            assignPublicIP: false
             tags:
               "KubernetesCluster": "6qsm86c2d"
           # Can be 'ubuntu', 'coreos' or 'centos'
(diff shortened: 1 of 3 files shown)


mrIncompetent:
@alvaroaleman 
I had to remove the `assignPublicIP` from the tests again as this only works in subnets which use a NAT gateway.
I've manually tested this in a AWS tenant which had this, but was not able to create an working environment in our AWS tenant :(

alvaroaleman:
> I had to remove the assignPublicIP from the tests again as this only works in subnets which use a NAT gateway.

Can you add that info to the example?

kron4eg:
wait a sec, isn't it enough to just drop instance into "private" subnet and that's end of the story?

mrIncompetent:
> wait a sec, isn't it enough to just drop instance into "private" subnet and that's end of the story?

Also requires a NAT gateway to have internet access. I haven't figured out to correctly configure it & did not wanted to spent much time on it.

kron4eg:
we have example config in terraform which can be translated to go calls directly :) https://github.com/kubermatic/kubeone/blob/master/examples/terraform/aws-private/main.tf

alvaroaleman:
> wait a sec, isn't it enough to just drop instance into "private" subnet and that's end of the story?

Thats an ortohonal thing to do or not? Instances can always have a public IP associated, it just wont work in some cases.

alvaroaleman:
/lgtm
/hold

In case you want to continue discussing pre-merge

PR #618, opened 2019-08-21
vSphere: Reconfigure VM's based on the clone's devices, not the template

mrIncompetent:
This ensures that we used the devices of the cloned VM to create our change specifications.
Otherwise the change specifications contain references to the template VM.

```release-note
NONE
```


mrIncompetent:
@alvaroaleman PTAL

PR #621, opened 2019-08-23
Fix kernel module loading part 2

mrIncompetent:
```release-note
NONE
```

/assign @alvaroaleman

### DIFF
--- a/pkg/userdata/centos/testdata/kubelet-v1.11-aws.yaml
+++ b/pkg/userdata/centos/testdata/kubelet-v1.11-aws.yaml
@@ -17,19 +17,15 @@ write_files:
     #!/usr/bin/env bash
     set -euo pipefail
 
-    kernel=$(uname -r)
-    kernel_main_version=${kernel:0:1}
-
     modprobe ip_vs
     modprobe ip_vs_rr
     modprobe ip_vs_wrr
     modprobe ip_vs_sh
 
-    # nf_conntrack_ipv4 got removed in the kernel version 4.19
-    if [[ "${kernel_main_version}" -ge "5" ]] || [[ "${kernel}" == 4.19* ]] || [[ "${kernel}" == 4.20* ]]; then
-      modprobe nf_conntrack
-    else
+    if modinfo nf_conntrack_ipv4 &> /dev/null; then
       modprobe nf_conntrack_ipv4
+    else
+      modprobe nf_conntrack
     fi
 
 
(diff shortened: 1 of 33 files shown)


pkg/userdata/centos/testdata/kubelet-v1.11-aws.yaml
```diff
@@ -25,8 +25,10 @@ write_files:
     modprobe ip_vs_wrr
     modprobe ip_vs_sh
 
-    # nf_conntrack_ipv4 got removed in the kernel version 5.0
-    if [ "${kernel_main_version}" -le "4" ]; then
+    # nf_conntrack_ipv4 got removed in the kernel version 4.19
+    if [[ "${kernel_main_version}" -ge "5" ]] || [[ "${kernel}" == 4.19* ]] || [[ "${kernel}" == 4.20* ]]; then
```
alvaroaleman: `s/[[/{/g` Posix compliance _and_smalller userdata :)

mrIncompetent: Then the substring comparison won't work (At least i cant get them working without adding way more characters) :grimacing: 

```sh
set -euo pipefail

kernel="4.19.8"
kernel_main_version=${kernel:0:1}

echo $kernel
echo $kernel_main_version

# nf_conntrack_ipv4 got removed in the kernel version 4.19
if [ $kernel_main_version -ge "5" ] || [ $kernel == 4.19* ] || [ $kernel == 4.20* ]; then
  echo "New Kernel"
else
  echo "Old Kernel"
fi
```

```
4.19.8
4
Old Kernel
```

alvaroaleman: (still busy discussing the matter with the expert, will give feedback once the correct POSIX way is found)

thz: `[ -z "${kernel##4.19*}" -o -z "${kernel##5*}" ]`

thz: But how about a completely different way:
```
if modinfo nf_conntrack_ipv4 2& > /dev/null; then
  modprobe nf_conntrack_ipv4
else
  modprobe nf_conntrack
fi
```

alvaroaleman:
@thz can you review and approve?

PR #624, opened 2019-08-27
Set current-context for bootstrap kubeconfig

alvaroaleman:

Fixes #

```release-note
none
```

/assing @thz

### DIFF
--- a/pkg/controller/machine/kubeconfig.go
+++ b/pkg/controller/machine/kubeconfig.go
@@ -66,12 +66,22 @@ func (r *Reconciler) createBootstrapKubeconfig(name string) (*clientcmdapi.Confi
 	}
 
 	outConfig := infoKubeconfig.DeepCopy()
+	clusterContextName := ""
+	for key := range infoKubeconfig.Clusters {
+		clusterContextName = key
+	}
+	cluster := outConfig.Clusters[clusterContextName].DeepCopy()
+	delete(outConfig.Clusters, clusterContextName)
+	outConfig.Clusters["a"] = cluster
 
 	outConfig.AuthInfos = map[string]*clientcmdapi.AuthInfo{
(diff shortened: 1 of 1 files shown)


pkg/controller/machine/kubeconfig.go
```diff
@@ -66,13 +68,23 @@ func (r *Reconciler) createBootstrapKubeconfig(name string) (*clientcmdapi.Confi
 	}
 
 	outConfig := infoKubeconfig.DeepCopy()
+	clusterContextName := ""
+	for key := range infoKubeconfig.Clusters {
+		clusterContextName = key
```
themue: So `clusterContextName` will get the last key as it always will be overwritten inside the loop?

pkg/controller/machine/kubeconfig.go
```diff
@@ -42,6 +42,8 @@ const (
 	tokenSecretKey           string            = "token-secret"
 	expirationKey            string            = "expiration"
 	tokenFormatter           string            = "%s.%s"
+	// Keep this short, userdata is limited
+	contextIdentifier string = "c"
```
themue: Pretty short. ;)

alvaroaleman: yeah, the problem we have is that the userdata size is limited, so I wanted to keep it short :P

pkg/controller/machine/kubeconfig.go
alvaroaleman: Yes, its because we have some validation down the line that fails if this has more than one key. Lemme add a comment for that.

alvaroaleman:
@themue I've added some comments explaining the rationale for this, I hope they make it easier to understand and also prevents ppl from removing the code in the future because they don't understand the purpose

PR #627, opened 2019-08-28
Add support for Hetzner Private Networks

### LINKED ISSUE #599: Add support for hetzner private network
Hetzner recently added support for private networking, would be good to extend machine-controller to support this.

This is how it looks like with new version of terraform hetzner provider: https://github.com/kubermatic/kubeone/pull/596/files#diff-8aec760bfdbb019e856b85fc10e43205

xmudrii:

This PR implements support for Hetzner Private Networks. A new optional field is added to the Hetzner spec called `network` that consumes a network ID or name. It is expected that the network already exists, i.e. that it's created by the user.

To make this work I had to vendor the latest version of the `hcloud-go` library which includes support for networks.

Fixes #599 

```release-note
Add support for Hetzner Private Networks
```

### DIFF
--- a/examples/hetzner-machinedeployment.yaml
+++ b/examples/hetzner-machinedeployment.yaml
@@ -49,7 +49,7 @@ spec:
             location: "fsn1"
             # Optional: network IDs or names
             networks:
-              - ""
+              - "<< YOUR_NETWORK >>"
             # Optional
             labels:
               my: label
--- a/examples/hetzner-machinedeployment.yaml
+++ b/examples/hetzner-machinedeployment.yaml
@@ -47,6 +47,9 @@ spec:
             # Optional
             datacenter: ""
             location: "fsn1"
+            # Optional: network IDs or names
+            networks:
+              - ""
             # Optional
             labels:
               my: label
--- a/pkg/cloudprovider/provider/hetzner/provider.go
+++ b/pkg/cloudprovider/provider/hetzner/provider.go
@@ -164,7 +164,7 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 	if len(c.Networks) != 0 {
 		for _, network := range c.Networks {
 			if _, _, err = client.Network.Get(ctx, network); err != nil {
-				return fmt.Errorf("failed to get network: %v", err)
+				return fmt.Errorf("failed to get network %q: %v", network, err)
 			}
 		}
 	}
--- a/pkg/cloudprovider/provider/hetzner/provider.go
+++ b/pkg/cloudprovider/provider/hetzner/provider.go
@@ -414,6 +414,9 @@ func (s *hetznerServer) Addresses() []string {
 	for _, fips := range s.server.PublicNet.FloatingIPs {
 		addresses = append(addresses, fips.IP.String())
 	}
+	for _, privateNetwork := range s.server.PrivateNet {
+		addresses = append(addresses, privateNetwork.IP.String())
+	}
 
 	return append(addresses, s.server.PublicNet.IPv4.IP.String(), s.server.PublicNet.IPv6.IP.String())
 }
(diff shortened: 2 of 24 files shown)


xmudrii:
@alvaroaleman When you get some time please take a look at questions I left in the PR description. It's nothing urgent, so no hurry. :slightly_smiling_face:

alvaroaleman:
> Running machine-controller binary on the local machine complains about plugins.

That is supposed to work :thinking: 

> Should I add some E2E (or other) tests for this feature?

Hmm, I think this is a bit tricky, because something has to create the network we use there beforehand

> Should we support just a single network or multiple networks?

I guess there is no harm in supporting more? I don't really see the usecase, but if we don't, someone will complain sooner or later I imagine

> How machine-controller knows which network interface to use by default,

I don't understand the question, where does the machine-controller care about the network interface?

xmudrii:
@alvaroaleman 
> I don't understand the question, where does the machine-controller care about the network interface?

That's what I'm asking. :slightly_smiling_face: Does the machine-controller care about the network interfaces anywhere? As far as I get all the existing code, it's just enough to attach the machine to the network and that should be it.

Also, have I broke something? I guess this is not really a flake. :sweat_smile:

alvaroaleman:
> Also, have I broke something? I guess this is not really a flake. ￼

not sure what was going on with the tests yesterday, probably hetzner running out of capacity.

>  Does the machine-controller care about the network interfaces anywhere? 

Not really, it uses the IP to match the machine to its node on platforms that do not have a cloudprovider like Hetzner but thats about it

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -386,6 +414,9 @@ func (s *hetznerServer) Addresses() []string {
 	for _, fips := range s.server.PublicNet.FloatingIPs {
 		addresses = append(addresses, fips.IP.String())
 	}
+	for _, privateNetwork := range s.server.PrivateNet {
+		addresses = append(addresses, privateNetwork.IP.String())
```
xmudrii: Is there a correct order here, e.g. should private addresses go before or after other addresses?

xmudrii:
> Not really, it uses the IP to match the machine to its node on platforms that do not have a cloudprovider like Hetzner but thats about it

Okay, that makes sense. I think this PR has everything covered now and should be ready for review. :slightly_smiling_face:

pkg/cloudprovider/provider/hetzner/provider.go
alvaroaleman: Nom, doesn't matter

examples/hetzner-machinedeployment.yaml
```diff
@@ -47,6 +47,9 @@ spec:
             # Optional
             datacenter: ""
             location: "fsn1"
+            # Optional: network IDs or names
+            networks:
+              - ""
```
alvaroaleman: can we put something like `<<my-network>>` between the braces?

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -152,6 +161,14 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 		}
 	}
 
+	if len(c.Networks) != 0 {
+		for _, network := range c.Networks {
+			if _, _, err = client.Network.Get(ctx, network); err != nil {
+				return fmt.Errorf("failed to get network: %v", err)
```
alvaroaleman: nit: Can we put the name of the network we couldnt get into the error?

examples/hetzner-machinedeployment.yaml
xmudrii: Added `<< YOUR_NETWORK >>`.

pkg/cloudprovider/provider/hetzner/provider.go
xmudrii: Fixed

xmudrii:
@alvaroaleman Fixed, PTAL

PR #638, opened 2019-10-15
define slices for all OS/providers

xrstf:
This adds two new slices and sorts the cloud providers alphabetically except for the fake provider, which is last.

```release-note
NONE
```


xrstf:
@alvaroaleman What can I do to fix the kittyflood?

PR #643, opened 2019-10-24
Add alibaba cloud provider to machine controller

moadqassem:
Adding a cloud provider for AlibabaCloud in the machine-controller
 
Fixes #4356

```release-note
NONE
```


alvaroaleman:
@moadqassem You need to rebase it onto master first, it become incompatible

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,413 @@
...
+			if err != nil {
+				return nil, err
+			}
+		case runningStatus:
+			ipAddress := ecs.CreateAllocatePublicIpAddressRequest()
+			ipAddress.InstanceId = foundInstance.InstanceId
+
+			_, err = client.AllocatePublicIpAddress(ipAddress)
+			if err != nil {
+				panic(err)
```
alvaroaleman: Don't panic

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,413 @@
...
+	createInstanceRequest.InstanceName = machine.Name
+	createInstanceRequest.InstanceType = c.InstanceType
+	createInstanceRequest.VSwitchId = c.VSwitchID
+	createInstanceRequest.InternetMaxBandwidthOut = requests.Integer(c.InternetMaxBandwidthOut)
+	encodedUserData := base64.StdEncoding.EncodeToString([]byte(userdata))
+	createInstanceRequest.UserData = encodedUserData
+	createInstanceRequest.SystemDiskCategory = "cloud_efficiency"
+	createInstanceRequest.ZoneId = c.ZoneID
+
+	res, err := client.CreateInstance(createInstanceRequest)
```
alvaroaleman: If the creation is a multi-step process, the provider needs to add a distinct finalizer for each step, to make sure we clean up even if the creation failed halfways. See the azure provider as an example

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,413 @@
...
+		case stoppedStatus:
+			startRequest := ecs.CreateStartInstanceRequest()
+			startRequest.InstanceId = res.InstanceId
+
+			_, err = client.StartInstance(startRequest)
+			if err != nil {
+				return nil, err
+			}
+		case runningStatus:
+			ipAddress := ecs.CreateAllocatePublicIpAddressRequest()
```
alvaroaleman: Does this require the instance to be running? What cleans up the public IP?

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,413 @@
...
+		foundInstance, err = getInstance(client, machine.Name)
+		if err != nil {
+			return nil, err
+		}
+
+		if tries >= maxRetries {
+			return nil, fmt.Errorf("instance %v isn't in a ready state", foundInstance.InstanceId)
+		}
+
+		switch instanceStatus(foundInstance.Status) {
```
alvaroaleman: Blocking for a prolongued time is a bit problematic, can it be avoided, e.G. by starting the instance within the `Get` func?

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,413 @@
...
+		return nil, errors.ErrInstanceNotFound
+	}
+
+	return &response.Instances.Instance[0], nil
+}
+
+func getImageIDForOS(os providerconfigtypes.OperatingSystem) (string, error) {
+	switch os {
+	case providerconfigtypes.OperatingSystemUbuntu:
+		return "ubuntu_18_04_64_20G_alibase_20190624.vhd", nil
```
alvaroaleman: This will get outdated quickly, can we do one of:
* Look this up dynamically 
* Allow users to set it
* Allow users to set it and fall back to looking it up dynamicall (e.G. AWS does this)

?

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -243,6 +244,30 @@ func TestPacketProvisioningE2E(t *testing.T) {
...
+	if len(accessKeyID) == 0 {
+		t.Fatal("unable to run the test suite, ALIBABA_ACCESS_KEY_ID environment variable cannot be empty")
+	}
+
+	accessKeySecret := os.Getenv("ALIBABA_ACCESS_KEY_SECRET")
+	if len(accessKeySecret) == 0 {
+		t.Fatal("unable to run the test suite, ALIBABA_ACCESS_KEY_SECRET environment variable cannot be empty")
+	}
+
+	excludeSelector := &scenarioSelector{osName: []string{"coreos"}}
```
alvaroaleman: If we don't support this, we must add it to the validation

pkg/cloudprovider/provider/alibaba/provider.go
moadqassem: Yes, otherwise it will return an error, that the machine is not running thus cannot allocate an IP address. About cleaning up the public IP, once the machine/instance is deleted, the public IP address will be unbound and available again.

pkg/cloudprovider/provider/alibaba/provider.go
moadqassem: The only thing that we allocate is a public IP and it gets deleted once the instance is deleted by default, thus I am not quite sure that it is necessary. especially that we don't invoke any deletion for the IP addresses in the cleanup method since its an implicit part of the instance cleanup.

pkg/cloudprovider/provider/alibaba/provider.go
alvaroaleman: Yeah but it shouldn't be available, it should be deallocated, just like we allocated it before. It doesn't really matter if the thing is in use, as long as its allocated, you have to pay

moadqassem: It will be deallocated once the instance which uses it has been deleted/released. Per VPC we have 2000 IP addresses that we can use and once an IP address is allocated they will be 1999 and once the instance which uses this IP is deleted the available IP addresses will be 2000 again which means it gets unbound and deallocated.

test/e2e/provisioning/all_e2e_test.go
moadqassem: Addressed here:
https://github.com/kubermatic/machine-controller/pull/643/files?file-filters%5B%5D=.go#diff-e96854fadd3b15014f29f87ac940a01bR421

pkg/cloudprovider/provider/alibaba/provider.go
alvaroaleman: It definitely is necesary because if we successfully created the instance but got an error trying to allocate the IP, `Create` will return an error which results in the cleanup finalizer not being added which results in us orphaning the instance.

moadqassem: I have added the finalizer, small note though, the allocating of the public IP address happens in the `Get` method, which means it will try to allocate the IP address multiple times.

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,439 @@
...
+	createInstanceRequest.VSwitchId = c.VSwitchID
+	createInstanceRequest.InternetMaxBandwidthOut = requests.Integer(c.InternetMaxBandwidthOut)
+	encodedUserData := base64.StdEncoding.EncodeToString([]byte(userdata))
+	createInstanceRequest.UserData = encodedUserData
+	createInstanceRequest.SystemDiskCategory = "cloud_efficiency"
+	createInstanceRequest.ZoneId = c.ZoneID
+
+	_, err = client.CreateInstance(createInstanceRequest)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create instance at Alibaba cloud: %v", err)
```
alvaroaleman: The finalizer is needed for the instance, not for the IP and must be added right here, once the createInstance request was done without an error. You do the `getIntance` call afterwards and that one may fail. We must make sure we add a finalizer as soon as we created something.

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,439 @@
...
+		}
+		return nil, fmt.Errorf("instance %v is in a stopped state", foundInstance.InstanceId)
+	case runningStatus:
+		if len(foundInstance.PublicIpAddress.IpAddress) == 0 {
+			ipAddress := ecs.CreateAllocatePublicIpAddressRequest()
+			ipAddress.InstanceId = foundInstance.InstanceId
+
+			_, err = client.AllocatePublicIpAddress(ipAddress)
+			if err != nil {
+				return nil, err
```
alvaroaleman: Please wrap this and all other errors so its possible to figure out where exactly a given error comes from

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,439 @@
...
+			}
+			if err = data.Update(machine, func(updatedMachine *v1alpha1.Machine) {
+				if !kuberneteshelper.HasFinalizer(updatedMachine, finalizerPublicIP) {
```
alvaroaleman: If the public IP doesn't need a cleanup we don't need a finalizer for it, we just need a finalizer that gets added as soon as the instance was  created. The main `Reconcile` only does this after the whole `Create` went throught successfully and it currently does another API call after the creation

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,439 @@
...
+	client, err := getClient(c.RegionID, c.AccessKeyID, c.AccessKeySecret)
+	if err != nil {
+		return nil, err
+	}
+
+	if c.Labels == nil {
+		c.Labels = map[string]string{}
+	}
+
+	c.Labels[machineUIDTag] = string(machine.UID)
```
alvaroaleman: We never set the labels on the instance

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,439 @@
...
+	client, err := ecs.NewClientWithAccessKey(regionID, accessKeyID, accessKeySecret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get Alibaba cloud client: %v", err)
+	}
+	return client, nil
+}
+
+func getInstance(client *ecs.Client, instanceName string) (*ecs.Instance, error) {
+	describeInstanceRequest := ecs.CreateDescribeInstancesRequest()
+	describeInstanceRequest.InstanceName = instanceName
```
alvaroaleman: Please use the UID label to identify your instance instead of the name

moadqassem: addressed here: https://github.com/kubermatic/machine-controller/pull/643/files#diff-e96854fadd3b15014f29f87ac940a01bR234-R247

pkg/cloudprovider/provider/alibaba/provider.go
moadqassem: addressed here: https://github.com/kubermatic/machine-controller/pull/643/files#diff-e96854fadd3b15014f29f87ac940a01bR234-R247

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,450 @@
...
+		}
+	}
+
+	client, err := getClient(c.RegionID, c.AccessKeyID, c.AccessKeySecret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get alibaba client: %v", err)
+	}
+
+	createInstanceRequest := ecs.CreateCreateInstanceRequest()
+	createInstanceRequest.ImageId, _ = p.getImageIDForOS(machine.Spec, pc.OperatingSystem)
```
alvaroaleman: Nit: Can we handle the error here?

pkg/cloudprovider/provider/alibaba/provider.go
alvaroaleman: I don't follow, how does that code address anything? The code here still filters the retrieved instance only based on the name and not on the tags

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,450 @@
...
+	}); err != nil {
+		return nil, fmt.Errorf("failed updating machine %v finzaliers: %v", machine.Name, err)
+	}
+
+	foundInstance, err := getInstance(client, machine.Name)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get alibaba instance %v due to %v", machine.Name, err)
+	}
+
+	tag := ecs.AddTagsTag{
```
alvaroaleman: Please configure the tags as part of the creation by setting it on the `createInstanceRequest` to make this atomic. Otherwise if the creation succeeds at the labeling does not, we won't find our instance in the `Get` func and will orphan it

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -0,0 +1,450 @@
...
+	foundInstance, err := getInstance(client, machine.Name)
+	if err != nil {
+		return false, fmt.Errorf("failed to get alibaba instance %v due to %v", machine.Name, err)
+	}
+
+	deleteInstancesRequest := ecs.CreateDeleteInstancesRequest()
+	deleteInstancesRequest.InstanceId = &[]string{foundInstance.InstanceId}
+
+	deleteInstancesRequest.Force = requests.Boolean("True")
+	if _, err = client.DeleteInstances(deleteInstancesRequest); err != nil {
```
alvaroaleman: Please update this code to ignore 404, i.E. if someone deleted the instance out of bounds, else we are stuck here because we won't remove the finalizer but the deletion doesn't succeed here

moadqassem: That's a good catch, however, I will add the check before the deletion. Here: https://github.com/kubermatic/machine-controller/pull/643/files/cc2554f14d210965e8c11b7945929038359302e7#diff-e96854fadd3b15014f29f87ac940a01bR272

moadqassem:
@alvaroaleman PTAL

pkg/cloudprovider/provider/alibaba/provider.go
alvaroaleman: I see. But then you need to move the finalizer cleanup after the `if err == cloudprovidererrors.ErrInstanceNotFound`

PR #654, opened 2019-11-19
Fix the cleanup condition in azure cloud provider 

moadqassem:
This PR fixes the finalizers updating when machines are being cleaned up.

Fixes #4637

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/provider/alibaba/provider.go
+++ b/pkg/cloudprovider/provider/alibaba/provider.go
@@ -278,7 +278,13 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
 		return false, fmt.Errorf("failed to delete instance with instanceID %s, due to %v", c.InstanceID, err)
 	}
 
-	return false, nil
+	if err := data.Update(machine, func(updatedMachine *v1alpha1.Machine) {
+		updatedMachine.Finalizers = kuberneteshelper.RemoveFinalizer(updatedMachine.Finalizers, finalizerInstance)
+	}); err != nil {
+		return false, fmt.Errorf("failed updating machine %v finzaliers: %v", machine.Name, err)
+	}
+
+	return true, nil
 }
 
 func (p *provider) MachineMetricsLabels(machine *v1alpha1.Machine) (map[string]string, error) {
--- a/pkg/cloudprovider/provider/alibaba/provider.go
+++ b/pkg/cloudprovider/provider/alibaba/provider.go
@@ -278,13 +278,7 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
 		return false, fmt.Errorf("failed to delete instance with instanceID %s, due to %v", c.InstanceID, err)
 	}
 
-	if err := data.Update(machine, func(updatedMachine *v1alpha1.Machine) {
-		updatedMachine.Finalizers = kuberneteshelper.RemoveFinalizer(updatedMachine.Finalizers, finalizerInstance)
-	}); err != nil {
-		return false, fmt.Errorf("failed updating machine %v finzaliers: %v", machine.Name, err)
-	}
-
-	return true, nil
+	return false, nil
 }
 
 func (p *provider) MachineMetricsLabels(machine *v1alpha1.Machine) (map[string]string, error) {
--- a/pkg/cloudprovider/provider/alibaba/provider.go
+++ b/pkg/cloudprovider/provider/alibaba/provider.go
@@ -32,6 +32,7 @@ import (
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider/instance"
 	alibabatypes "github.com/kubermatic/machine-controller/pkg/cloudprovider/provider/alibaba/types"
 	cloudprovidertypes "github.com/kubermatic/machine-controller/pkg/cloudprovider/types"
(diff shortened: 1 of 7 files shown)


test/e2e/provisioning/helper.go
```diff
@@ -51,6 +51,8 @@ var (
 	}
 )
 
+const defaultNamespace = "kube-system"
```
moelsayed: Nit: maybe use `metav1.NamespaceSystem` instead?

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -250,7 +250,7 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
 			if err := data.Update(machine, func(updatedMachine *v1alpha1.Machine) {
 				updatedMachine.Finalizers = kuberneteshelper.RemoveFinalizer(updatedMachine.Finalizers, finalizerInstance)
 			}); err != nil {
-				return false, fmt.Errorf("failed updating machine %v finzaliers: %v", machine.Name, err)
+				return false, err
```
kron4eg: why removing error wrapping?

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -278,7 +278,13 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
 		return false, fmt.Errorf("failed to delete instance with instanceID %s, due to %v", c.InstanceID, err)
 	}
 
-	return false, nil
+	if err := data.Update(machine, func(updatedMachine *v1alpha1.Machine) {
+		updatedMachine.Finalizers = kuberneteshelper.RemoveFinalizer(updatedMachine.Finalizers, finalizerInstance)
```
kron4eg: I'm not entirely sure, but probably we can omit second RemoveFinalizer() call (and logic), and simply return the error, to let this method be called once again.

pkg/cloudprovider/provider/alibaba/provider.go
```diff
@@ -278,7 +278,13 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
...
+	}); err != nil {
+		return false, err
+	}
+
+	return true, nil
```
kron4eg: are you sure it should be true here (was false)?

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -489,11 +489,22 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
 	_, err = p.get(machine)
 	// If a defunct VM got created, the `Get` call returns an error - But not because the request
 	// failed but because the VM has an invalid config hence always delete except on err == cloudprovidererrors.ErrInstanceNotFound
-	if err != cloudprovidererrors.ErrInstanceNotFound {
-		klog.Infof("deleting VM %q", machine.Name)
-		if err = deleteVMsByMachineUID(context.TODO(), config, machine.UID); err != nil {
-			return false, fmt.Errorf("failed to delete instance for  machine %q: %v", machine.Name, err)
+	if err != nil {
+		if err == cloudprovidererrors.ErrInstanceNotFound {
```
kron4eg: this "if err cloudprovidererrors.ErrInstanceNotFound then remove finalizer" is getting repetitive, should this be extracted somehow?

pkg/cloudprovider/provider/alibaba/provider.go
moadqassem: Yeah this was false before by mistake

pkg/cloudprovider/provider/alibaba/provider.go
moadqassem: Yeah actually this is how it was. But when I changed it to true I duplicated it. Maybe its better to revert to my old implementation, even though alicoud delete instances sync.

pkg/cloudprovider/provider/alibaba/provider.go
moadqassem: Yeah because alicloud deletes the instances in sync(at least that's what the docs says) but for extra gaurentee I will revert it back

pkg/cloudprovider/provider/alibaba/provider.go
moadqassem: I added the wrapping by myself but when I took another look on all the update functions, not of them wrap it. so I removed it for consistency

pkg/cloudprovider/provider/azure/provider.go
moadqassem: maybe we can add util functionality to check for that instead of keep writing all the time. WDYT ?

kron4eg: yeah, something like `DeleteFinalizerOnInstanceNotFound(error, *v1alpha1.Machine) (bool, error) {}`

moadqassem: Sounds like a good idea. Plus we need a `data *github.com/kubermatic/machine-controller/pkg/cloudprovider/types.ProviderData`  as a third parameter to execute the update. 

Since this is out of the scope for this PR I would like to create a ticket for it and do it separately, because all the current provider is doing the same currently and I don't want to touch all of them in this PR

PR #663, opened 2019-12-16
Add support for EBS volume encryption

moadqassem:
This indicates whether the encryption state of an EBS volume is changed while being restored from a backing snapshot. 

Fixes #

```release-note
None
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -1,0 +1,26 @@
+presubmits:
+  #########################################################
+  # unit tests
+  #########################################################
+
+  - name: pull-machine-controller-e2e-aws-ebs-encryption-enabled
+    always_run: true
+    decorate: true
+    error_on_eviction: true
+    clone_uri: 'ssh://git@github.com/kubermatic/machine-controller.git'
+    labels:
+      preset-aws: "true"
+      preset-hetzner: "true"
+      preset-e2e-ssh: "true"
+    spec:
+      containers:
+        # Uses go1.11.1
+        - image: quay.io/kubermatic/dep:0.5.4-2
+          command:
+            - "./hack/ci-e2e-test.sh"
+          args:
+            - "TestAWSProvisioningE2EWithEbsEncryptionEnabled"
+          resources:
+            requests:
+              memory: 1Gi
+              cpu: 500m
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -1,7 +1,4 @@
 presubmits:
-  #########################################################
-  # unit tests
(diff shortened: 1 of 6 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -509,6 +514,7 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloudprovidertypes.Pr
 					DeleteOnTermination: aws.Bool(true),
 					VolumeType:          aws.String(config.DiskType),
 					Iops:                config.DiskIops,
+					Encrypted:           pointer.BoolPtr(config.EBSVolumeEncrypted),
```
alvaroaleman: Can we add an e2e test where this is enabled?

test/e2e/provisioning/testdata/machinedeployment-aws.yaml
```diff
@@ -33,6 +33,7 @@ spec:
             instanceProfile: "kubernetes-v1"
             diskSize: 50
             diskType: "gp2"
+            ebsVolumeEncrypted: false
```
alvaroaleman: Please also add this to the sample manifest in `examples/`

pkg/cloudprovider/provider/aws/provider.go
moadqassem: I ran the current e2e tests with this sat to `true` and the tests are passed( I am not sure though if this change has any influence on the machine, e.g: the machine will not be provisioned). We can add another test with this enabled. WDYT?

alvaroaleman: sgtm, but I don't think we need to test this on all kube versions and all distributions, one version on one distribution should be enough

moadqassem:
@alvaroaleman PTAL

alvaroaleman:
> @alvaroaleman PTAL

So where is the green e2e run as requested?

moadqassem:
> > @alvaroaleman PTAL
> 
> So where is the green e2e run as requested?

https://public-prow.loodse.com/view/gcs/prow-public-data/pr-logs/pull/kubermatic_machine-controller/663/pull-machine-controller-e2e-aws-ebs-encryption-enabled/1207254364134051840

.prow.yaml
```diff
@@ -0,0 +1,26 @@
+presubmits:
+  #########################################################
+  # unit tests
```
alvaroaleman: Also no unittest. Other than that, looks good.

PR #664, opened 2020-01-06
kubelet sets intial machine taints via --register-with-taints

multi-io:
Signed-off-by: Olaf Klischat <o.klischat@syseleven.de>

If the machine spec of a machine being created contains taints, this adds a corresponding `--register-with-taints` option to the machine's kubelet invocation, which means the taints will be set on the node resource from the start. Without this patch, the taints would only be added asynchronously during the first reconcile loop after the node has come up. At that point pods that don't tolerate the taints may already have been scheduled, and they won't be removed again (unless the taint's effect is NoExecute), leading to a permanently misscheduled pod. This can be easily reproduced with DaemonSet pods.

```release-note
kubelet sets intial machine taints via --register-with-taints
```

### DIFF
--- a/pkg/userdata/helper/kubelet.go
+++ b/pkg/userdata/helper/kubelet.go
@@ -20,7 +20,10 @@ import (
 	"bytes"
 	"fmt"
 	"net"
+	"strings"
 	"text/template"
+
+	corev1 "k8s.io/api/core/v1"
 )
 
 const (
@@ -58,6 +61,9 @@ const (
 {{- if .PauseImage }}
 --pod-infra-container-image={{ .PauseImage }} \
 {{- end }}
+{{- if .InitialTaints }}
+--register-with-taints={{- .InitialTaints }} \
+{{- end }}
 --kube-reserved=cpu=100m,memory=100Mi,ephemeral-storage=1Gi \
 --system-reserved=cpu=100m,memory=100Mi,ephemeral-storage=1Gi \
 --cgroup-driver=systemd`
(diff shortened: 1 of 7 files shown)


pkg/userdata/helper/kubelet.go
```diff
@@ -19,6 +19,7 @@ package helper
 import (
 	"bytes"
 	"fmt"
+	v1 "k8s.io/api/core/v1"
```
kron4eg: Please move this after "text/template" with a blank line and preferably with import alias corev1

```
import (
	"bytes"
	"fmt"
	"net"
	"text/template"

	corev1 "k8s.io/api/core/v1"
)
```

pkg/userdata/helper/kubelet.go
```diff
@@ -58,6 +59,9 @@ const (
 {{- if .PauseImage }}
 --pod-infra-container-image={{ .PauseImage }} \
 {{- end }}
+{{- if .InitialTaints }}
+--register-with-taints={{- range $i, $t := .InitialTaints }}{{ $t.Key }}={{ $t.Value }}:{{ $t.Effect }}{{ if (lt $i (sub (len $.InitialTaints) 1)) }},{{ end }}{{ end }} \
```
kron4eg: Probably this logic can be moved to `KubeletFlags()` and be pre-computed before template render. Would be much easier to reason about it.

pkg/userdata/helper/kubelet_test.go
```diff
@@ -18,6 +18,7 @@ package helper
 
 import (
 	"fmt"
+	v1 "k8s.io/api/core/v1"
```
kron4eg: Same here as with import in pkg/userdata/helper/kubelet.go

multi-io:
I'll address the comments tomorrow.

multi-io:
comments addressed

xrstf:
/approve

@kron4eg Please :lgtm: when you're happy with it.

kron4eg:
/assign @moadqassem

@moadqassem PTAL

PR #666, opened 2020-01-09
Enable swap accounting for ubuntu deployments

### LINKED ISSUE #597: Enable cgroups memory and swap accounting  for Ubuntu
Docker recommends enabling cgroups memory and swap accounting in the kernel to allow Docker to fully manage memory constraints. This is not the case on Ubuntu by default.

These settings are required by some Kubernetes-based platforms, for example SUSE Cloud Application Platform (CAP).

see also https://github.com/Azure/acs-engine/issues/3107

moelsayed:
Enable swap accounting for ubuntu deployments
Fixes #597

```release-note
NONE
```

### DIFF
--- a/pkg/userdata/ubuntu/provider.go
+++ b/pkg/userdata/ubuntu/provider.go
@@ -144,6 +144,12 @@ write_files:
   content: |
 {{ kernelSettings | indent 4 }}
 
+- path: "/etc/default/grub.d/60-swap-accounting.cfg"
+  content: |
+    # Added by kubermatic machine-controller
+    # Enable cgroups memory and swap accounting
+    GRUB_CMDLINE_LINUX="cgroup_enable=memory swapaccount=1"
+
 - path: "/etc/apt/sources.list.d/docker.list"
   permissions: "0644"
   content: deb [arch=amd64] https://download.docker.com/linux/ubuntu bionic stable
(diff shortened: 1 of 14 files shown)


kron4eg:
there is cloud-init module to reboot machine: https://cloudinit.readthedocs.io/en/latest/topics/modules.html#power-state-change

moelsayed:
@kron4eg I know, but that would _always_ reboot the nodes after deployment. I am not sure we'd want that.

kron4eg:
@moelsayed you could exploit `power_state.condition`.

pkg/userdata/ubuntu/provider.go
```diff
@@ -265,6 +271,13 @@ write_files:
...
     DEBIAN_FRONTEND=noninteractive apt-get -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" dist-upgrade -y
     {{- end }}
+
+    # Update grub to include kernel command options to enable swap accounting.
+    if grep -v -q swapaccount=1 /proc/cmdline
+    then
+      update-grub
+      touch /var/run/reboot-required
+    fi
     if [[ -e /var/run/reboot-required ]]; then
```
kron4eg: this section need to be deleted

moelsayed: Removing this and using the power_state module instead is causing a race condition in one of the test cases. Basically power_state triggers a reboot _during_ the OS update process, rendering apt in a broken state that requires a manual fix and causes the node to be stuck in a loop. 

I removed the powerstate module entry and restored this to avoid this effect.

moelsayed:
@kron4eg could you please re- /lgtm /approve?

PR #669, opened 2020-01-24
Support RedHat Enterprise Linux os in the machine controller

moadqassem:
Adding the support for RHEL os in the machine controller.

Fixes kubermatic/kubermatic#4965

```release-note
NONE
```

### DIFF
--- a/pkg/userdata/rhel/provider.go
+++ b/pkg/userdata/rhel/provider.go
@@ -1,0 +1,320 @@
+/*
+Copyright 2019 The Machine Controller Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//
+// UserData plugin for RHEL.
+//
+
+package rhel
+
+import (
+	"bytes"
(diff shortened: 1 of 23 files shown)


pkg/userdata/rhel/provider.go
```diff
@@ -0,0 +1,320 @@
...
+		req.CloudConfig = *pconfig.OverwriteCloudConfig
+	}
+
+	if pconfig.Network != nil {
+		return "", errors.New("static IP config is not supported with RHEL")
+	}
+
+	rhelConfig, err := LoadConfig(pconfig.OperatingSystemSpec)
+	if err != nil {
+		return "", fmt.Errorf("failed to parse OperatingSystemSpec: '%v'", err)
```
xrstf: Why do we quote the error? (`'%v'`)

pkg/userdata/rhel/provider_test.go
```diff
@@ -0,0 +1,258 @@
...
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+var (
+	update = flag.Bool("update", false, "update testdata files")
+
+	pemCertificate = `-----BEGIN CERTIFICATE-----
```
xrstf: Is it obvious what this certificate is or should we leave a comment?

moadqassem: This is a certificate which actually used in all our userdata providers testing

pkg/userdata/rhel/provider.go
moadqassem: This is actually a copy and past mistake from this code here: 
https://github.com/kubermatic/machine-controller/blob/master/pkg/userdata/centos/provider.go#L48

Will fix it in both centos and rhel

PR #671, opened 2020-01-29
Support datastore cluster

### LINKED ISSUE #653: Support datastore clusters on vSphere
It should be possible to specify a datastore cluster in a Machine/MachineDeployment instead of specific datastore on vSphere. Specifying a datastore and a datastore cluster should be mutually exclusive.

irozzo-1A:
This PR provides initial datastore clusters support for the vSphere provider.

Fixes #653 

```release-note
vsphere: datastore clusters can now be specified for VMs instead of singular datastores
```


pkg/cloudprovider/provider/vsphere/types/types.go
```diff
@@ -20,18 +20,23 @@ import (
...
 	Cluster        providerconfigtypes.ConfigVarString `json:"cluster"`
-	Folder         providerconfigtypes.ConfigVarString `json:"folder,omitempty"`
-	Datastore      providerconfigtypes.ConfigVarString `json:"datastore"`
-	CPUs           int32                               `json:"cpus"`
-	MemoryMB       int64                               `json:"memoryMB"`
-	DiskSizeGB     *int64                              `json:"diskSizeGB,omitempty"`
-	AllowInsecure  providerconfigtypes.ConfigVarBool   `json:"allowInsecure"`
+	Folder         providerconfigtypes.ConfigVarString `json:"folder"`
+
+	// Either Datastore of DatastoreCluster have to be provided.
```
moadqassem: I think there is a  type here, of should be or right?

irozzo-1A: Indeed, good catch

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -362,46 +363,48 @@ func TestVsphereProvisioningE2E(t *testing.T) {
 		fmt.Sprintf("<< VSPHERE_CLUSTER >>=%s", vsCluster),
 	}
 	runScenarios(t, excludeSelector, params, VSPhereManifest, fmt.Sprintf("vs-%s", *testRunIdentifier))
+	runScenarios(t, excludeSelector, params, VSPhereDatastoreClusterManifest, fmt.Sprintf("vs-dsc-%s", *testRunIdentifier))
```
moadqassem: Why are you running the test explicitly by calling the test runner? since you have already created a test inside `all_e2e_test.go` file, it would be better to create a job(in prow) which runs the test( please take a look here:  https://github.com/kubermatic/machine-controller/blob/master/.prow.yaml#L25-L45 ).

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -69,15 +69,27 @@ func createClonedVM(ctx context.Context, vmName string, config *Config, session
 		targetVMFolder = datacenterFolders.VmFolder
 	}
 
-	datastore, err := session.Finder.Datastore(ctx, config.Datastore)
+	relocateSpec := types.VirtualMachineRelocateSpec{
+		//Datastore:    types.NewReference(datastore.Reference()),
```
moadqassem: Please get rid of this comment

test/e2e/provisioning/all_e2e_test.go
irozzo-1A: I did not create a new test, the one you're probably referring to was already there and commented out

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -362,46 +363,48 @@ func TestVsphereProvisioningE2E(t *testing.T) {
...
-//		name:              "Coreos Docker Kubernetes v1.11.0",
-//		osName:            "coreos",
-//		containerRuntime:  "docker",
-//		kubernetesVersion: "1.11.0",
-//		executor:          verifyCreateAndDelete,
-//	}
-//
-//	testScenario(t, scenario, fmt.Sprintf("vs-staticip-%s", *testRunIdentifier), params, vssip_manifest, false)
-//}
+/*func TestVsphereStaticIPProvisioningE2E(t *testing.T) {
```
irozzo-1A: I will revert this it's just a diff for nothing

test/e2e/provisioning/all_e2e_test.go
moadqassem: Ok maybe I didn't explain it well. The new manifest that you have created to test the datastore cluster can be run as a separate test. Take a look at the aws cloud provider's tests. We have one test for default behavior and another test for `ebs` encryption. so instead of calling the runner n times the runner checks what to test and it will test it.

irozzo-1A: ok, you' re proposing to have two distinct tests. I initially did that, then I grouped them to avoid some repetitions. It's true that having having dedicated tests bring some benefits, I can do that.

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -45,31 +45,46 @@ import (
...
 
+type configGetter func(v1alpha1.ProviderSpec) (*Config, *providerconfigtypes.Config, *vspheretypes.RawConfig, error)
+
+// Ensures that getConfig function is of type configGetter
+var _ configGetter = (&provider{}).getConfig
+
 type provider struct {
 	configVarResolver *providerconfig.ConfigVarResolver
+	// This is to ease unit tests
+	getConfigFunc configGetter
```
moadqassem: Why this is needed? there is already a method which get the configurations of the provider for you.

moadqassem: Instead of defining a new unneeded field to the provider to test something, you can use the getConfig in your tests and pass your custom configurations there instead of finding a way to override those values.

irozzo-1A: It's to ease unit tests, I'm mocking this function in the unit test:
https://github.com/irozzo-1A/machine-controller/blob/support-datastore-cluster/pkg/cloudprovider/provider/vsphere/provider_test.go#L136

moadqassem: Yeah as I said this you can simply use `v1alpha1.ProviderSpec` object to customize your test. I am not sure it is a good idea to add code that belongs to the tests in the production code that code should be built in the matter that its testable and it is in this case.

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -185,8 +208,17 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 	}
 	defer session.Logout()
 
-	if _, err := session.Finder.Datastore(ctx, config.Datastore); err != nil {
-		return fmt.Errorf("failed to get datastore %s: %v", config.Datastore, err)
+	// If the datastore cluster is present it takes precedence over datastore.
+	if config.DatastoreCluster != "" && config.Datastore == "" {
```
moadqassem: Since you point that out explicitly that datasotre cluster has a higher priority here, this code can be simpler by: 
```go
if config.DatastoreCluster == "" && config.Datastore == "" {
		return fmt.Errorf("at least one of datastore and datastore cluster should be specified: %v", err)
	}

if config.DatastoreCluster != "" {
	if _, err := session.Finder.DatastoreCluster(ctx, config.DatastoreCluster); err != nil {
		return fmt.Errorf("failed to get datastore cluster %s: %v", config.DatastoreCluster, err)
	}
} else {
	if _, err := session.Finder.Datastore(ctx, config.Datastore); err != nil {
		return fmt.Errorf("failed to get datastore %s: %v", config.Datastore, err)
	}
}
```
This way you have an early exist and you are not building condition trees.

irozzo-1A: @moadqassem Actually the comment is outdated, finally I decided to have the two properties mutually exclusive, because it's misleading if both are specified. It's better to fail in such scenario. I'll update the comment.

pkg/cloudprovider/provider/vsphere/provider.go
irozzo-1A: > Instead of defining a new unneeded field to the provider to test something, you can use the getConfig in your tests and pass your custom configurations there instead of finding a way to override those values.

I know that I can but it's not the `getConfig` method that I want to test here. I don't want to open a philosophical debate whether changing the code to ease testing is a good practice or not, but I normally prefer to reduce the scope of unit test as much as I can. I found this pattern somewhere in the k8s code base quite handy. What's your concern about adding an unexported field?

moadqassem: Well, this field is simply not needed and the logic behind can be substituted by using a method that already exists. I know that you are not testing the `getConfig` method but you want to test something that the `getConfig` assigns in the runtime code. All our cloud providers are following the very same pattern.

irozzo-1A: to be fair the other providers have no unit tests, then the problem does not arise :-)
There are many ways of doing things, in absolute terms it's not required, but it's functional to its purpose. It simplifies a bit the test at the cost of complexifying a bit the code. I think it's largely matters of taste. We can discuss pros and cons offline if you wish, that said I think it's not worthy spending much time on this point, if it's a blocker for you I can invest some time to change it.

moadqassem: (IMHO) Adding code in order to facilitate your test is a bad design. The reason for that is, most of the time when code is added in production only to facilitate/enable testing show some design flaws, which is funny in your case because you can test the code without adding anything new or adding any wrapping. You can easily test your code without adding anything, however, you simply chose to:

- Add something in the production code which has no value 
- Ignore what's already there and can be used and reflects the applications' workflow in production.
- Breaking the code consistency between this particular cloud provider and the others without realistic and valuable reasons.
 
If none of those points are sufficient we can take this discussion as you suggested offline.

irozzo-1A: @moadqassem thanks for clarifying your view, I honestly think that your remark is more aesthetic than substantial, but as you dislike it so much and removing it does not cost me much effort I'll remove it ;-)

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -376,7 +408,19 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
 			}
 		}
 	}
-
+	var props mo.VirtualMachine
+	// Obtain VM properties
+	if err := virtualMachine.Properties(ctx, virtualMachine.Reference(), nil, &props); err != nil {
```
moadqassem: Can you please explain why this is needed?

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -376,7 +408,19 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
...
+		return false, fmt.Errorf("error getting VM properties: %v", err)
+	}
+	datastorePath, err := getDatastorePathObjFromVMDiskPath(props.Summary.Config.VmPathName)
+	if err != nil {
+		return false, err
```
moadqassem: Please add error wrapping for this error here.

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -376,7 +408,19 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
...
+	}
+	datastore, err := session.Finder.Datastore(ctx, datastorePath.Datastore)
+	if err != nil {
+		return false, err
```
moadqassem: Please add error wrapping for this error here

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -218,13 +211,80 @@ func createClonedVM(ctx context.Context, vmName string, config *Config, session
...
+		// Get the storage placement result
+		storageResourceManager := object.NewStorageResourceManager(session.Client.Client)
+		result, err := storageResourceManager.RecommendDatastores(ctx, storagePlacementSpec)
+		if err != nil {
+			return nil, fmt.Errorf("error occurred while getting storage placement recommendation: %v", err)
+		}
+
+		// Get the recommendations
+		recommendations := result.Recommendations
+		if len(recommendations) == 0 {
```
moadqassem: If there was no recommended datastore in this cluster, can't we just pick a random one and report a warning error, instead of completely aborting the task?

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -218,13 +211,80 @@ func createClonedVM(ctx context.Context, vmName string, config *Config, session
 	return virtualMachine, nil
 }
 
-func uploadAndAttachISO(ctx context.Context, session *Session, vmRef *object.VirtualMachine, localIsoFilePath, datastoreName string) error {
-	datastore, err := session.Finder.Datastore(ctx, datastoreName)
-	if err != nil {
-		return fmt.Errorf("failed to get datastore: %v", err)
+func resolveDatastoreRef(ctx context.Context, config *Config, session *Session, vm *object.VirtualMachine, folder *object.Folder, cloneSpec *types.VirtualMachineCloneSpec) (*types.ManagedObjectReference, error) {
+	// Based on https://github.com/vmware/govmomi/blob/v0.22.1/govc/vm/clone.go#L358
+	if config.DatastoreCluster != "" && config.Datastore == "" {
```
moadqassem: I don't think you need to keep checking whether the value of `DatastoreCluster` and the `Datastore` repeatedly. You do validation on the machine creation and I think that's fairly enough.

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -218,13 +211,80 @@ func createClonedVM(ctx context.Context, vmName string, config *Config, session
...
+		return nil, fmt.Errorf("please provide either a datastore or a datastore cluster")
 	}
+}
+
+func uploadAndAttachISO(ctx context.Context, session *Session, vmRef *object.VirtualMachine, localIsoFilePath string) error {
 	p := soap.DefaultUpload
 	remoteIsoFilePath := fmt.Sprintf("%s/%s", vmRef.Name(), "cloud-init.iso")
+	var props mo.VirtualMachine
+	// Obtain VM properties
+	if err := vmRef.Properties(ctx, vmRef.Reference(), nil, &props); err != nil {
```
moadqassem: There was a parameter for this function where the datastore name is sent via the machine config, thus this is not needed. Is there a reason why removed that and fetched the properties of the vm to get the ds path, while you had before hand?

moadqassem:
@irozzo-1A I would suggest having an interface that might be called DatastoreManager added to the vSphere provider, with two types(e.g: `DatastoreProvider` and `DatastoreClusterProvider`) implementing that interface. This way the code can be a bit cleaner and without the need of always checking whether I use a datastore or a datastore cluster.

pkg/cloudprovider/provider/vsphere/provider.go
irozzo-1A: This is to get the datastore where the VM is located, before we had the datastore in the config now we are not sure to have it (i.e. if the datastore cluster is specified instead) then I added this code to get the actual VM datastore location.

pkg/cloudprovider/provider/vsphere/helper.go
irozzo-1A: I think it's better to fail in such circumstances, because it's an unexpected condition. Note that this code is inspired by the[ vmware command line tool](https://github.com/vmware/govmomi/blob/e179aa206e05606aeb5b1fde9c79a6613a0e048d/govc/vm/clone.go#L385) , as I'm not a vSphere expert this was my reference.

pkg/cloudprovider/provider/vsphere/helper.go
irozzo-1A: IMHO this is simple enough to stay as it is, I don't see any substantial gain in doing something like:

```
if config.DatastoreCluster != nil {
   ...
   return ...
}
if config.Datastore != nil {
   ...
   return ...
}
```
I agree that the validation should guarantee that just one of the two is present.

pkg/cloudprovider/provider/vsphere/helper.go
irozzo-1A: I could pass the datastore explicitly to save an API call as a small optimisation, but as it is a property of the VM there is no real need of passing it explicitly.

irozzo-1A:
> @irozzo-1A I would suggest having an interface that might be called DatastoreManager added to the vSphere provider, with two types(e.g: `DatastoreProvider` and `DatastoreClusterProvider`) implementing that interface. This way the code can be a bit cleaner and without the need of always checking whether I use a datastore or a datastore cluster.

I'm not sure to clearly understand the the design you have in mind, but personally I don't think that we need more abstractions to handle this simple use case, at the end we are just doing the check during the validation and the machine creation.

pkg/cloudprovider/provider/vsphere/helper.go
moadqassem: I am not saying to do it as you suggested, you should do it as: 
```
if config.DatastoreCluster != "" {
 .....
} else {
// Since its gonna be either Datastore Cluster or Datastore, then this is a Datastore and the validation method can guarantee that!!
}
```

Please adjust accordingly as we should never cover something that will never happen!?

pkg/cloudprovider/provider/vsphere/helper.go
moadqassem: Please check with the customer or with kamil about this!

pkg/cloudprovider/provider/vsphere/helper.go
moadqassem: You are adding a boilerplate to achieve something that you already have. Please retain the code as it was because there is no real value in executing a request to vsphere to request for something you have beforehand!

pkg/cloudprovider/provider/vsphere/provider.go
moadqassem: As I said in another comment, you have already the value of the resource that you are trying reach and in order to make sure what you have is a datastore or a datastore cluster just check which one is set. This is a canonical workflow where the source of truth is coming from the configurations(especially when it comes to storage!)

pkg/cloudprovider/provider/vsphere/helper.go
irozzo-1A: it's more self contained now, the information about the datastore is already associated to the VM. Passing it explicitly is more effective because we save an API call, but is redundant. On the other hand if we want to optimise we can directly pass the reference to the datastore object, it's even better than what we had before. I don't think that this kind of optimisation should be the main driver at this stage.

pkg/cloudprovider/provider/vsphere/helper.go
irozzo-1A: Again, I tend to think that this is kind of cosmetic change. Your proposal is less redundant, mine is more defensive.

moadqassem: Yours is simply checking for something that the added workflow doesn't satisfy. The `config.Datastore == ""` is always true  if the `config.DatastoreCluster != ""` and same for ` config.DatastoreCluster == ""` in the scond condition, it is always true if `config.Datastore != ""`. thus those checks are useless not defensive.

pkg/cloudprovider/provider/vsphere/helper.go
moadqassem: The information of the vm should come from the machine configuration and the machine configuration is coming from the MachineDeployment configurations and those configurations are the source of truth which the vm should be bounded with!!

pkg/cloudprovider/provider/vsphere/helper.go
irozzo-1A: If the validation makes his job indeed there is no reason why both are populated. I try to explain you what I meant with defensive. Consider those two scenarios:

1. The mutatingwebhookconfigurations for machine-controller gets deleted for some reason
2. In a further development we break accidentally the logic of the validation

Maybe it's a bit paranoid :-) but it was just to explain myself

pkg/cloudprovider/provider/vsphere/helper.go
irozzo-1A: the `MachineDeployment` after my development may not contain the datastore

pkg/cloudprovider/provider/vsphere/helper.go
kdomanski: It seems that failing to get a recommendation is an error condition and vsphere. Let's treat it as such unless we find a compelling reason to ignore the failure.

pkg/cloudprovider/provider/vsphere/helper.go
kdomanski: An implicit additional check costs nothing and adds more readability to what is expected than implicitly relying on the condition being checked by the validation code.

pkg/cloudprovider/provider/vsphere/provider.go
irozzo-1A: done

pkg/cloudprovider/provider/vsphere/provider.go
irozzo-1A: refer to the reply I thin in the other comment

irozzo-1A:
\retest

test/e2e/provisioning/testdata/machinedeployment-vsphere-datastore-cluster.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: "cluster.k8s.io/v1alpha1"
```
moadqassem: Please create a similar file like this on in the example folder.

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -360,3 +412,18 @@ func validateDiskResizing(disks []*types.VirtualDisk, requestedSize int64) error
 	}
 	return nil
 }
+
+//getDatastoreFromVM gets the datastore where the VM files are located.
+func getDatastoreFromVM(ctx context.Context, session *Session, vmRef *object.VirtualMachine) (*object.Datastore, error) {
```
moadqassem: You create a vm based on specific configurations, such as whether this vm is using a datastore cluster or datastore. And based on your implementation you made it distinct by using different names in the template: `datastore` or `datastoreCluster`. Those two properties are mutually exclusive which means they will never be set for the same machines set. And since you have all this information in the config a simple if statement can decide what is the right path for this machine:

```
if config.DatastoreCluster != "" && config.Datastore == "" {
  return session.Finder.Datastore(ctx, config.DatastoreCluster)
} else {
  return session.Finder.Datastore(ctx, config.Datastore)
}
```
The places where this function is needed have already access to that information.

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -69,15 +69,26 @@ func createClonedVM(ctx context.Context, vmName string, config *Config, session
 		targetVMFolder = datacenterFolders.VmFolder
 	}
 
-	datastore, err := session.Finder.Datastore(ctx, config.Datastore)
+	relocateSpec := types.VirtualMachineRelocateSpec{
```
moadqassem: As the comment below suggests, we split the cloning mechanism from reconfiguring the newly spawned vm due to some permissions validating. I think this is a good approach to have better insights into what went wrong.

pkg/cloudprovider/provider/vsphere/helper.go
irozzo-1A: this does not work: `session.Finder.Datastore(ctx, config.DatastoreCluster)`.  It would mean trying to lookup a datastore with the datastore cluster name we had in input.

kdomanski: `*object.Datastore` and `*object.StoragePod` are different, so I think it's not so straightforward. Even `resolveDatastoreRef` returns a specific datastore based on a cluster, except it is much more complicated than `getDatastoreFromVM`, so maybe just looking up the specific datastore where the VM already is, is not such a bad idea.

pkg/cloudprovider/provider/vsphere/helper.go
irozzo-1A: Personally I'm not convinced that splitting clone, relocate, reconfigure in three distinct steps is a good idea. We could do this with one single API call and we would reduce the probability of failing in the middle of the machine preparation. In addition if the issue is that the error message is not explicit enough we should address it elsewhere. That said I did that because in order to have a placement recommendation I had to provide the `CloneSpec` that include that include the `VirtualMachineRelocateSpec`. At this point cloning the VM and relocate it after that seemed a nonsense to me.

kdomanski: I agree with Iacopo. The previous authors never properly determined and documented the necessary permissions, so "tell which specific action failed due to lacking permissions" was used as a crutch for trial-and-error.

PR #676, opened 2020-02-06
Using systemd instead of cgroupfs in rhel

moadqassem:
Using `systemd` as  a cgroup driver for both kubelet and docker instead of cgroupfs.  

Fixes #

```release-note
None
```


xmudrii:
Seems like a flake
/retest

PR #678, opened 2020-02-10
add redhat subscription manager

moadqassem:
At the moment all the machines which are created by the machine controller in the cloud provider that supports rhel create redhat subscriptions. However, when those machines are deleted the subscriptions are not deleted automatically. This PR adds the functionality to delete those subscriptions(optionally) whenever those machines are deleted.
 
Fixes #

At the moment this functionality is only supported in the Kubevirt and Vsphere cloud providers since they are the only cloud providers that supports rhel for the time being. However, rhel will be integrated with all cloud providers in the next steps.

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/rhsm/subscription_manager.go
+++ b/pkg/cloudprovider/rhsm/subscription_manager.go
@@ -1,0 +1,228 @@
+package rhsm
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"k8s.io/klog"
+	"net/http"
+	"net/url"
+	"strings"
+	"time"
+)
+
+// RedHatSubscriptionManager is responsible for removing redhat subscriptions.
+type RedHatSubscriptionManager interface {
+	UnregisterInstance(machineName string) error
+}
+
+type pagination struct {
+	Offset int `json:"offset"`
+	Limit  int `json:"limit"`
+	Count  int `json:"count"`
+}
+
+type body struct {
+	Name string `json:"name"`
+	UUID string `json:"uuid"`
+}
+
+type systemsResponse struct {
+	Pagination pagination `json:"pagination"`
+	Body       []*body    `json:"body"`
+}
+
+type credentials struct {
+	AccessToken string `json:"access_token"`
+	ExpiresIn   int    `json:"expires_in"`
+}
+
+type defaultRedHatSubscriptionManager struct {
+	offlineToken string
+	authUrl      string
+	apiUrl       string
+	client       *http.Client
+	credentials  *credentials
+}
+
+var unauthenticatedRequestError = errors.New("unauthenticated")
+
+func NewRedHatSubscriptionManager(offlineToken string) (RedHatSubscriptionManager, error) {
+	if offlineToken == "" {
+		return nil, errors.New("offline token, authURL, or apiPath cannot be empty")
+	}
+
+	token, err := base64.StdEncoding.DecodeString(offlineToken)
+	if err != nil {
+		return nil, fmt.Errorf("failed decoding offline token: %v", err)
+	}
+
+	return &defaultRedHatSubscriptionManager{
+		client:       &http.Client{},
+		authUrl:      "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
+		apiUrl:       "https://api.access.redhat.com/management/v1/systems",
+		offlineToken: string(token),
+	}, nil
+}
+
+func (d *defaultRedHatSubscriptionManager) UnregisterInstance(machineName string) error {
+	if d.credentials == nil {
+		klog.Info("access token has been expired, refreshing token")
+		if err := d.refreshToken(); err != nil {
+			return fmt.Errorf("failed to refresh offline token")
+		}
+	}
+
+	var (
+		retries    = 0
+		maxRetries = 15
+	)
+
+	for retries < maxRetries {
+		machineUUID, err := d.findSystemsProfile(machineName)
+		if err != nil {
+			if err == unauthenticatedRequestError {
+				if err := d.refreshToken(); err != nil {
+					klog.Errorf("failed to refresh offline token: %v", err)
+					continue
+				}
+			}
+			return fmt.Errorf("failed to find system profile: %v", err)
+		}
+
+		if machineUUID == "" {
+			klog.Infof("machine uuid %s is not found", machineUUID)
+			return nil
+		}
+
+		err = d.deleteSubscription(machineUUID)
+		if err == nil {
+			klog.Infof("subscription for vm %v has been deleted successfully", machineUUID)
+			return nil
+		}
+
+		if err == unauthenticatedRequestError {
+			if err := d.refreshToken(); err != nil {
+				klog.Errorf("failed to refresh offline token: %v", err)
+				continue
+			}
+		}
+		klog.Errorf("failed to delete subscription for system: %s due to: %v", machineUUID, err)
+		time.Sleep(2 * time.Second)
+		retries++
+	}
+
+	return errors.New("failed to delete system profile after max retires number has been reached")
+}
+
+func (d *defaultRedHatSubscriptionManager) refreshToken() error {
+	payload := url.Values{}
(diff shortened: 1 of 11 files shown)


pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+type credentials struct {
+	AccessToken string `json:"access_token"`
+	ExpiresIn   int    `json:"expires_in"`
+}
+
+type defaultRedHatSubscriptionManager struct {
+	offlineToken string
+	authUrl      string
```
golangcibot: struct field `authUrl` should be `authURL` (from `golint`)

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,228 @@
...
+	apiUrl       string
```
golangcibot: struct field `apiUrl` should be `apiURL` (from `golint`)

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,228 @@
...
+	client       *http.Client
+	credentials  *credentials
+}
+
+var unauthenticatedRequestError = errors.New("unauthenticated")
```
golangcibot: error var unauthenticatedRequestError should have name of the form errFoo (from `golint`)

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,238 @@
...
+func NewRedHatSubscriptionManager(offlineToken string) (RedHatSubscriptionManager, error) {
+	if offlineToken == "" {
+		return nil, errors.New("offline token, authURL, or apiPath cannot be empty")
+	}
+
+	return &defaultRedHatSubscriptionManager{
+		client:       &http.Client{},
+		authURL:      "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
+		apiURL:       "https://api.access.redhat.com/management/v1/systems",
+		offlineToken: string(offlineToken),
```
golangcibot: unnecessary conversion (from `unconvert`)

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,238 @@
...
+			return fmt.Errorf("failed to refresh offline token")
+		}
+	}
+
+	var (
+		retries    = 0
+		maxRetries = 15
+	)
+
+	for retries < maxRetries {
```
kron4eg: this loop needs backoff logic, and probably backoff should be reported to calling functions somehow (I mean the API surface of this package).

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,238 @@
...
+		}
+
+		err = d.deleteSubscription(machineUUID)
+		if err == nil {
+			klog.Infof("subscription for vm %v has been deleted successfully", machineUUID)
+			return nil
+		}
+
+		if err == errUnauthenticatedRequest {
+			if err := d.refreshToken(); err != nil {
```
kron4eg: what if we expose `RefreshToken()` and let callers decide on this?

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+type body struct {
+	Name string `json:"name"`
+	UUID string `json:"uuid"`
+}
+
+type systemsResponse struct {
+	Pagination pagination `json:"pagination"`
+	Body       []*body    `json:"body"`
```
kron4eg: Is it possible that upstream API will return `[{<object>}, nil, {<object>}]`?

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,238 @@
...
+	apiURL       string
+	client       *http.Client
+	credentials  *credentials
+}
+
+var errUnauthenticatedRequest = errors.New("unauthenticated")
+
+func NewRedHatSubscriptionManager(offlineToken string) (RedHatSubscriptionManager, error) {
+	if offlineToken == "" {
+		return nil, errors.New("offline token, authURL, or apiPath cannot be empty")
```
kron4eg:
```suggestion
		return nil, errors.New("RedHatSubscriptionManager offline token cannot be empty")
```

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	return &defaultRedHatSubscriptionManager{
+		client:       &http.Client{},
```
kron4eg: oh, this is really really dangerous, PTAL https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,238 @@
...
+*/
+
+package rhsm
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"k8s.io/klog"
```
kron4eg: should be in separate block

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -0,0 +1,238 @@
...
+}
+
+type systemsResponse struct {
+	Pagination pagination `json:"pagination"`
+	Body       []*body    `json:"body"`
+}
+
+type credentials struct {
+	AccessToken string `json:"access_token"`
+	ExpiresIn   int    `json:"expires_in"`
```
kron4eg: Since we have this timestamp, we can take it into account before other then `auth` API calls, since we can calculate before the call if our current access token have expired.

moadqassem: Actually this is not needed and I actually added it so I do this calculation but after some scenarios testing, it simply doesn't worth to add this logic. The reason for that is, the token is renewed every 5 minutes. Thus, we try to execute a request if it passes good, if not(because the token has expired, the API will return 401 and we refresh the token.

pkg/cloudprovider/rhsm/subscription_manager.go
moadqassem: For the used rout no. it returns this body: 
```
{"pagination": {"offset": 0, "limit": 100,"count": 1}, 
  "body": [
	"{"name": "test-machine-mqasim", "uuid": machineUUID"}
  ]
}

File: pkg/cloudprovider/rhsm/subscription_manager.go (hunk 6 above)
Comment (moadqassem) [reply pkg/cloudprovider/rhsm/subscription_manager.go]: Why? this logic should be encapsulated within the SubscriptionManager because it is part of the api calls. The caller should only be interested if the Unsubscribe was successful or not. That's why I abstracted into an interface and had the unsubscribe there.

File: pkg/cloudprovider/rhsm/subscription_manager.go (hunk 5 above)
Comment (moadqassem) [reply pkg/cloudprovider/rhsm/subscription_manager.go]: I am not quite sure that back off mechanism is really necessary here, the reason for that is we try to execute this loop 15 times every two seconds, not 100 times each 500 milliseconds, and if this fails forever( > 15) the reconcile loop will trigger this again on the next reconcile which has a back-off mechanism( which has a back-off mechanism )

File: pkg/cloudprovider/rhsm/subscription_manager.go (hunk 7 above)
Comment (kron4eg) [reply pkg/cloudprovider/rhsm/subscription_manager.go]: Then I'd suggest
```suggestion
	Body       []body    `json:"body"`
```
slice of body pointers is kinda useless (and dangerous, if case if there will be for real a nil, because there is no check for nil of body == NPE).

moadqassem: How is that dangerous? the rest api will return an empty body(`"body": []`), which will be a slice of length zero and the loop won't be processed. However, having a slice of pointers is not benefiting me since I am not changing any values in the slice and I am only reading it. will change it slice of structs instead of pointers.

PR #679, opened 2020-02-12
Support rhel in Openstack cloud provider

moadqassem:
Adding support for RHEL os in openstack cloud provider

Fixes #5009

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -38,6 +38,7 @@ import (
 	cloudprovidererrors "github.com/kubermatic/machine-controller/pkg/cloudprovider/errors"
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider/instance"
 	openstacktypes "github.com/kubermatic/machine-controller/pkg/cloudprovider/provider/openstack/types"
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider/rhsm"
 	cloudprovidertypes "github.com/kubermatic/machine-controller/pkg/cloudprovider/types"
 	"github.com/kubermatic/machine-controller/pkg/providerconfig"
 	providerconfigtypes "github.com/kubermatic/machine-controller/pkg/providerconfig/types"
@@ -86,7 +87,8 @@ type Config struct {
 	RootDiskSizeGB        *int
 	NodeVolumeAttachLimit *uint
 
-	Tags map[string]string
+	Tags    map[string]string
+	manager rhsm.RedHatSubscriptionManager
 }
 
 const (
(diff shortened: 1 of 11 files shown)


pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -853,6 +862,11 @@ func (p *provider) cleanupFloatingIP(machine *v1alpha1.Machine, updater cloudpro
 		return fmt.Errorf("failed to delete %s finalizer from Machine: %v", floatingIPReleaseFinalizer, err)
 	}
 
+	if pc.OperatingSystem == providerconfigtypes.OperatingSystemRHEL && c.manager != nil {
+		if err := c.manager.UnregisterInstance(machine.Name); err != nil {
+			return fmt.Errorf("failed delete machine %s subscription: %v", machine.Name, err)
```
xrstf: failed *to* delete

test/e2e/provisioning/helper.go
```diff
@@ -125,6 +126,20 @@ func testScenario(t *testing.T, testCase scenario, cloudProvider string, testPar
 	scenarioParams = append(scenarioParams, fmt.Sprintf("<< KUBERNETES_VERSION >>=%s", testCase.kubernetesVersion))
 	scenarioParams = append(scenarioParams, fmt.Sprintf("<< YOUR_PUBLIC_KEY >>=%s", os.Getenv("E2E_SSH_PUBKEY")))
 
+	rhelSubscriptionManagerUser := os.Getenv("RHEL_SUBSCRIPTION_MANAGER_USER")
+	rhelSubscriptionManagerPassword := os.Getenv("RHEL_SUBSCRIPTION_MANAGER_PASSWORD")
+	rhsmOfflineToken := os.Getenv("REDHAT_SUBSCRIPTIONS_OFFLINE_TOKEN")
```
xrstf: These can be moved inside the if statement.

pkg/cloudprovider/provider/openstack/provider.go
moadqassem: actually the place of subscription deletion is wrong, it should be at line 599

test/e2e/provisioning/helper.go
```diff
@@ -125,6 +126,20 @@ func testScenario(t *testing.T, testCase scenario, cloudProvider string, testPar
...
+
+	if testCase.osName == string(providerconfigtypes.OperatingSystemRHEL) {
+		if rhelSubscriptionManagerUser == "" || rhelSubscriptionManagerPassword == "" || rhsmOfflineToken == "" {
+			t.Errorf("Unable to run kubevirt tests, KUBEVIRT_E2E_TESTS_KUBECONFIG, RHEL_SUBSCRIPTION_MANAGER_USER, " +
```
xrstf: In all other error messages about missing env vars, we only mention the relevant ones for the specific check. I would therefore not mention `KUBEVIRT_E2E_TESTS_KUBECONFIG` here again, its existence has already been checked. Maybe phrase it `Unable to run kubevirt tests with RHEL, .... must be set.`.

moadqassem: Yeah, that's a copy and paste mistake ;). Kubevirt shouldn't be mentioned here because this message now is generic for all cloud providers thus should be something like: 
`Unable to run e2e tests, RHEL_SUBSCRIPTION_MANAGER_USER, RHEL_SUBSCRIPTION_MANAGER_PASSWORD, and REDHAT_SUBSCRIPTIONS_OFFLINE_TOKEN must be set when rhel is used as an os`.

PR #680, opened 2020-02-14
Migrate to go modules

irozzo-1A:
This PR is to migrate from dep dependency manager to modules.

Fixes #

Take particular attention to [Makefile](https://github.com/kubermatic/machine-controller/pull/680/files#diff-b67911656ef5d18c4ae36cb6741b7965) changes.

```release-note
Migrated to gomods
```


go.mod
```diff
@@ -0,0 +1,78 @@
...
+	github.com/vmware/govmomi v0.22.2
+	golang.org/x/crypto v0.0.0-20191206172530-e9b2fee46413
+	golang.org/x/oauth2 v0.0.0-20190402181905-9f3314589c9a
+	golang.org/x/xerrors v0.0.0-20191011141410-1b5146add898 // indirect
+	gomodules.xyz/jsonpatch v2.0.1+incompatible // indirect
+	google.golang.org/api v0.3.1
+	gopkg.in/gcfg.v1 v1.2.3
+	gopkg.in/ini.v1 v1.46.0 // indirect
+	gopkg.in/warnings.v0 v0.1.2 // indirect
+	k8s.io/api v0.0.0-20191004120003-3a12735a829a
```
kron4eg: There is support go mods from upstream: 
https://pkg.go.dev/mod/k8s.io/api@v0.17.3?tab=versions (the earliest version is v0.15.7, for all k8s.io dependencies). This however could potentially conflict with sigs.k8s.io/controller-runtime v0.3.0 go.mod file definition.

irozzo-1A: Ok, I'll give it a try

grpc.log
```diff
@@ -0,0 +1,71 @@
+github.com/kubermatic/machine-controller google.golang.org/grpc@v1.27.1
```
xmudrii: What is the purpose of this file? I've never seen it before.

irozzo-1A: This is something I used to investigate a conflict on grpc, I committed this by mistake. Thx for spotting it ;-)

irozzo-1A:
> What is the reason for keeping the `vendor` directory committed? With Go modules, that's not required anymore, so we can eventually get rid of that one.

I kept it mainly because we're still supporting go 1.12, and the default behaviour for 1.12 when the repository is located in GOPATH is not to use modules. Keeping the vendor directory makes the build work even without modules support. But it's a good point, we could definitely get rid of 70M of stuff.

kron4eg:
@irozzo-1A BTW, can you please not delete Gopkg.toml and Gopkg.lock for now. We have other `dep` based projects that currently importing machine-controller and I'm most certain their dep resolution will start misbehaving.

irozzo-1A:
> @irozzo-1A BTW, can you please not delete Gopkg.toml and Gopkg.lock for now. We have other `dep` based projects that currently importing machine-controller and I'm most certain their dep resolution will start misbehaving.

@kron4eg I don't know `dep` in depth enough to understand the exact implication of removing those files for a project that has `machine-controller` as a dependency if any, but I did a test on one of our projects and the migration was smooth. I would rather prefer removing them to avoid double maintenance.

Makefile
```diff
@@ -12,72 +12,62 @@
...
-		-ldflags '-s -w' \
-		-o machine-controller-userdata-rhel \
-		github.com/kubermatic/machine-controller/cmd/userdata/rhel
-
-webhook: $(shell find cmd pkg -name '*.go') vendor
-	go build -v \
-		-ldflags '-s -w' \
-		-o webhook \
-		github.com/kubermatic/machine-controller/cmd/webhook
+			make machine-controller CGO_ENABLED="$(CGO_ENABLED)" MODFLAG="$(MODFLAG)" E2E_SSH_PUBKEY=
```
xmudrii: I'm not a makefile expert, but I'm wondering what is the reason for overriding `E2E_SSH_PUBKEY` with empty value? Or is something else happening here?

irozzo-1A: good point, let me explain that because it isn't obvious. We are defaulting the [E2E_SSH_PUBKEY](https://github.com/kubermatic/machine-controller/pull/680/files#diff-b67911656ef5d18c4ae36cb6741b7965L21) variable by running this shell command `cat ~/.ssh/id_rsa.pub`, unless a value is provided, and it can be empty. In targets `machine-controller-docker` and  `test-unit` we're invoking make inside a container and I noted that we were not forwarding the values that could have been provided in input for `CGO_ENABLED` and `E2E_SSH_PUBKEY`. This was a bug and I fixed it. Note that the `E2E_SSH_PUBKEY` is not required for both targets, the only reason why I specified an empty value is to avoid evaluating the variable and polluting the output with:
```
cat: /root/.ssh/id_rsa.pub: No such file or directory
```
That said I later added [this](https://github.com/kubermatic/machine-controller/pull/680/files#diff-b67911656ef5d18c4ae36cb6741b7965R19) test, that makes this fix useless, so I'll remove it.
I also spotted on the same line that the target that should be run is now `all` and not `machine-controller`.

xmudrii: Makes sense, thank you for such a detailed explanation!

kron4eg:
/hold cancel

as tested by @irozzo-1A, kubermatic works even with removed Gopkg files

kron4eg:
//approve

Makefile
```diff
@@ -12,72 +12,62 @@
...
 machine-controller-docker:
 	@docker run --rm \
 		-v $$PWD:/go/src/github.com/kubermatic/machine-controller \
 		-v $$PWD/.buildcache:/cache \
 		-e GOCACHE=/cache \
 		-w /go/src/github.com/kubermatic/machine-controller \
 		golang:$(GO_VERSION) \
-			make machine-controller
-
-.PHONY: machine-controller
```
moadqassem: I think after removing this, the `run-machine-controller.sh` in the hack folder will fail or?

irozzo-1A: Good catch, `machine-controller` target is now covered by [this](https://v1-15.docs.kubernetes.io/docs/setup/release/notes/#unchanged) pattern rule .The problem is that it does not include the build of the userdata binaries that are required for the script to run. I have either to use `all` target or a newly introduced target that includes only `machine-controller` and userdata.

PR #687, opened 2020-02-19
AWS: Use security group on instance creation.

thz:
Previously an instance on AWS was always created without specifying
a security group immediately followed by an API call to attach
the security group. This will not work in certain vpc-sharing
scenarios as there is no access to the default security group.
Also: this reduces AWS API usage.

```release-note
NONE
```


thz:
/hold cancel
@moadqassem rebased!

PR #689, opened 2020-02-20
Improve openstack default subnet selection

thz:
Previously no openstack subnet would be defaulted when more than one is available.

Details of this change on openstack default subnet selection:
* pick the first subnet if more than one is available (instead of returning nil)
* error out when no subnet is available (after subnet api request)
* continue to skip subnet api request when network has exactly one subnet already

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/helper.go
+++ b/pkg/cloudprovider/provider/openstack/helper.go
@@ -414,18 +414,15 @@ NetworkLoop:
 }
 
 func getDefaultSubnet(client *gophercloud.ProviderClient, network *osnetworks.Network, region string) (*string, error) {
-	if len(network.Subnets) == 0 {
-		return nil, nil
-	} else if len(network.Subnets) == 1 {
+	if len(network.Subnets) == 1 {
 		return &network.Subnets[0], nil
-	} else {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/openstack/helper.go
```diff
@@ -414,18 +414,15 @@ NetworkLoop:
 }
 
 func getDefaultSubnet(client *gophercloud.ProviderClient, network *osnetworks.Network, region string) (*string, error) {
-	if len(network.Subnets) == 0 {
-		return nil, nil
-	} else if len(network.Subnets) == 1 {
+	if len(network.Subnets) == 1 {
```
moadqassem: What about: 
```
if len(network.Subnets) > 0 {
  return &network.Subnets[0], nil
} else {
.......
}
```

thz: I explicitly did not want to rigorously change the semantics for the case that `network.Subnets` has exactly one element already.
But perhaps we should clarify first if API-requesting sub-networks brings more value (more recent data or a different set of sub-networks than available in the networks structs).

PR #690, opened 2020-02-20
AWS: improve AMI-ID error message

thz:
The error message when searching for the AMI-ID fails did not indicate this problem.

```release-note
NONE
```


kron4eg:
This has nothing to do with DO. So let's skip it.
/override pull-machine-controller-e2e-digitalocean

moadqassem:
@kron4eg Well, as long as we can't create machines in DO then this will fail anyway :)

PR #692, opened 2020-02-21
Use pagination when listing Digitalocean droplets

xrstf:
The DO provider used a nonsensical upper limit (1000), which is waaay above the actual DO hard limit. And the code would break if someone really had more than 1000 droplets in their account.

This PR implements very basic paging support for listing droplets.

```release-note
Fix Digitalocean if there are more than 200 droplets in the same account.
```

### DIFF
--- a/pkg/cloudprovider/provider/digitalocean/provider.go
+++ b/pkg/cloudprovider/provider/digitalocean/provider.go
@@ -378,12 +378,9 @@ func (p *provider) get(machine *v1alpha1.Machine) (*doInstance, error) {
 		}
 	}
 
-	ctx := context.TODO()
-	client := getClient(c.Token)
-	droplets, rsp, err := client.Droplets.List(ctx, &godo.ListOptions{PerPage: 1000})
-
+	droplets, err := p.listDroplets(c.Token)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -395,6 +392,35 @@ func (p *provider) get(machine *v1alpha1.Machine) (*doInstance, error) {
 	return nil, cloudprovidererrors.ErrInstanceNotFound
 }
 
+func (p *provider) listDroplets(token string) ([]godo.Droplet, error) {
+	ctx := context.TODO()
+	client := getClient(token)
+	result := make([]godo.Droplet, 0)
+	page := 1
```
thz: https://github.com/digitalocean/godo
has an slightly simpler example for this paging. it does not start counting at 1 but 0.

xrstf: Yeah I'll make it prettier, I was in a haste when I pushed this.

xrstf: It's now a hypertext enabled HATEOAS client.

PR #706, opened 2020-03-11
Enable redhat subscriptions pagination

moadqassem:
Support getting red hat subscriptions paginations for rhsm. 
  
Fixes kubermatic/kubermatic#5200

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/rhsm/subscription_manager.go
+++ b/pkg/cloudprovider/rhsm/subscription_manager.go
@@ -65,7 +65,7 @@ type defaultRedHatSubscriptionManager struct {
 
 var errUnauthenticatedRequest = errors.New("unauthenticated")
 
-func NewRedHatSubscriptionManager(offlineToken string, requestLimit int) (RedHatSubscriptionManager, error) {
+func NewRedHatSubscriptionManager(offlineToken string) (RedHatSubscriptionManager, error) {
 	if offlineToken == "" {
 		return nil, errors.New("RedHatSubscriptionManager offline token cannot be empty")
 	}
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -238,3 +230,47 @@ func (d *defaultRedHatSubscriptionManager) deleteSubscription(uuid string) error
...
+	data, err := ioutil.ReadAll(res.Body)
+	if err != nil {
+		return nil, fmt.Errorf("failed while reading response: %v", err)
+	}
+
+	if res.StatusCode != http.StatusOK {
+		if res.StatusCode == http.StatusUnauthorized {
+			return nil, errUnauthenticatedRequest
+		}
+		return nil, fmt.Errorf("error while exeucting request with status code: %v and message: %s", res.StatusCode, string(data))
```
xrstf: Can you fix this typo (`exeucting`) throughout this file?

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -168,41 +170,31 @@ func (d *defaultRedHatSubscriptionManager) refreshToken() error {
...
 		}
-		return "", fmt.Errorf("error while exeucting request with status code: %v and message: %s", res.StatusCode, string(data))
 	}
 
-	var fetchedSystems = &systemsResponse{}
-	if err := json.Unmarshal(data, fetchedSystems); err != nil {
-		return "", fmt.Errorf("failed while unmarshalling data: %v", err)
-	}
+	var counter = 1
+	for counter <= (numberOfRequest(systemsInfo.Pagination.Count, d.requestsLimiter) - 1) {
```
xrstf: I'm not 100% a friend of this logic. While it does work, it feels complicated to calculate the number of pages instead of just querying the API and ending the loop when `numberOfFetchedSystems < limit`.

Ooooh is that `systemsInfo.Pagination.Count` thing the reason why we made one initial request outside of this loop? If so I am even more in favor of changing the loop logic, unless I am overlooking someting?

PR #708, opened 2020-03-11
Use OAuth2 client instead of implementing access token retrieval and refresh

irozzo-1A:

This PR uses the oauth2 client from go subrepository in order to avoid implementing the logic for obtaining and refreshing the tokens when talking to RHSM API

Fixes #

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/rhsm/subscription_manager.go
+++ b/pkg/cloudprovider/rhsm/subscription_manager.go
@@ -30,6 +30,8 @@ import (
 	"k8s.io/klog"
 )
 
+const defaultTimeout = 10 * time.Second
+
 // RedHatSubscriptionManager is responsible for removing redhat subscriptions.
 type RedHatSubscriptionManager interface {
 	//TODO(irozzo) add context in input to give more control to the caller
@@ -65,15 +67,18 @@ func NewRedHatSubscriptionManager(offlineToken string) (RedHatSubscriptionManage
 		return nil, errors.New("RedHatSubscriptionManager offline token cannot be empty")
 	}
 	return &defaultRedHatSubscriptionManager{
-		apiURL: "https://api.access.redhat.com/management/v1/systems",
-		client: newOAuthClientWithRefreshToken(offlineToken, "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token"),
+		apiURL:          "https://api.access.redhat.com/management/v1/systems",
+		client:          newOAuthClientWithRefreshToken(offlineToken, "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token"),
+		requestsLimiter: 100,
 	}, nil
 }
 
 func newOAuthClientWithRefreshToken(refreshToken string, tokenURL string) *http.Client {
 	ctx := context.Background()
-	// Use the custom HTTP client when requesting a token.
-	httpClient := &http.Client{Timeout: 5 * time.Second}
+	// Use the custom HTTP client when requesting an access token in order to
+	// set a timeout value.
+	// See: https://github.com/golang/oauth2/blob/c85d3e98c914e3a33234ad863dcbff5dbc425bb8/internal/token.go#L232
+	httpClient := &http.Client{Timeout: defaultTimeout}
 	ctx = context.WithValue(ctx, oauth2.HTTPClient, httpClient)
 	conf := &oauth2.Config{
 		ClientID: "rhsm-api",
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -69,40 +64,40 @@ func NewRedHatSubscriptionManager(offlineToken string) (RedHatSubscriptionManage
...
 	}
-
 	return &defaultRedHatSubscriptionManager{
-		client: &http.Client{
-			Timeout: 10 * time.Second,
-		},
-		authURL:         "https://sso.redhat.com/auth/realms/redhat-external/protocol/openid-connect/token",
-		apiURL:          "https://api.access.redhat.com/management/v1/systems",
-		offlineToken:    offlineToken,
-		requestsLimiter: 100,
```
moadqassem: Please revert back this line as this break the current logic and run the rhsm into an infinite loop which is why the e2e test will fail probably

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -69,40 +64,40 @@ func NewRedHatSubscriptionManager(offlineToken string) (RedHatSubscriptionManage
...
-func (d *defaultRedHatSubscriptionManager) UnregisterInstance(machineName string) error {
-	if d.credentials == nil {
-		klog.Info("access token has been expired or not initialized, refreshing token")
-		if err := d.refreshToken(); err != nil {
-			return fmt.Errorf("failed to refresh offline token: %v", err)
-		}
+func newOAuthClientWithRefreshToken(refreshToken string, tokenURL string) *http.Client {
+	ctx := context.Background()
+	// Use the custom HTTP client when requesting a token.
+	httpClient := &http.Client{Timeout: 5 * time.Second}
```
irozzo-1A: By better looking into it this timeout seem to apply for the token request only, I should add a timeout for the API calls

pkg/cloudprovider/rhsm/subscription_manager.go
irozzo-1A: ok, I removed this by mistake

pkg/cloudprovider/rhsm/subscription_manager.go
moadqassem: Since you are only adding a timeout to the client nothing else, you can set the timeout explicitly on the client instead of passing it by the context which might be overwritten by the new context.

moadqassem:
```go
conf := &oauth2.Config{
		ClientID: "rhsm-api",
		Endpoint: oauth2.Endpoint{
			TokenURL: tokenURL,
		},
	}
client := conf.Client(ctx, tok)
client.Timeout =  5 * time.Second
```

moadqassem: btw, why did you change the timeout from 10 to 5?

pkg/cloudprovider/rhsm/subscription_manager.go
```diff
@@ -17,20 +17,22 @@ limitations under the License.
...
 	"time"
 
+	"golang.org/x/oauth2"
+
 	"k8s.io/klog"
 )
 
 // RedHatSubscriptionManager is responsible for removing redhat subscriptions.
 type RedHatSubscriptionManager interface {
+	//TODO(irozzo) add context in input to give more control to the caller
```
moadqassem: Well adding this might interfere with the logic of the default `UnregisterInstance` method implementation, since everything in the rhsm is abstracted away from the caller(cloud provider) and we are not exposing anything.

pkg/cloudprovider/rhsm/subscription_manager.go
irozzo-1A: > Since you are only adding a timeout to the client nothing else, you can set the timeout explicitly on the client instead of passing it by the context which might be overwritten by the new context.

The client that I'm passing through the context is used by the retrieve token logic (see https://godoc.org/golang.org/x/oauth2#example-Config--CustomHTTP), if I don't provide it the default client is used instead. 
So it's ok to pass it but it's not enough. I should set the default timeout to the returned client as well as you suggested.

pkg/cloudprovider/rhsm/subscription_manager.go
irozzo-1A: What I meant here is that it would be nice to provide the context explicitly by the caller of `UnregisterInstance`, so that it can set timeout and cancellation for instance.

PR #712, opened 2020-03-18
Converge docker.json config usage across operating systems

kron4eg:

* consistently use `dockerConfig` across OS in userdata
* install the same docker version where possible
* use `"max-size":"100m"` as docker log limit everywhere

```release-note
NONE
```

### DIFF
--- a/pkg/userdata/centos/provider.go
+++ b/pkg/userdata/centos/provider.go
@@ -184,7 +184,10 @@ write_files:
     hostnamectl set-hostname {{ .MachineSpec.Name }}
     {{ end }}
 
-    yum install -y docker-1.13.1 \
+    yum install -y yum-utils
+    yum-config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
+
+    yum install -y docker-ce-18.09.9-3.el7 \
       ebtables \
       ethtool \
       nfs-utils \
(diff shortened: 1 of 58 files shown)


pkg/userdata/helper/helper.go
```diff
@@ -102,17 +102,22 @@ SystemMaxUse=5G
 }
 
 type dockerConfig struct {
-	ExecOpts           []string `json:"exec-opts"`
-	StorageDriver      string   `json:"storage-driver"`
-	InsecureRegistries []string `json:"insecure-registries"`
-	RegistryMirrors    []string `json:"registry-mirrors"`
+	ExecOpts           []string          `json:"exec-opts,omitempty"`
+	StorageDriver      string            `json:"storage-driver,omitempty"`
+	StorageOpts        []string          `json:"storage-opts,omitempty"`
```
moadqassem: Where this field is used?

kron4eg: it was intended for RHEL, but since it works without it it's not used currently

pkg/userdata/centos/provider.go
```diff
@@ -202,7 +205,6 @@ write_files:
     systemctl enable --now vmtoolsd.service
     {{ end -}}
 {{- /* Without this, the conformance tests fail with differing tests causing it, the common denominator: They look for some string in container logs and get an empty log */ -}}
```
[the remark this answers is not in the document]
kron4eg: yeah %) removed

kron4eg:
alibaba fails with

```
admission webhook "machinedeployments.machine-controller.kubermatic.io" denied the request: defaulting or validation failed: validation failed: invalid/not supported operating system specified "centos": os not supported 
```

moadqassem:
> edeployments.machine-controller.kubermati

Yeah that's ok, we can forget about this error for now Since it is not relevant to the PR :)

PR #714, opened 2020-03-19
create KubeletConfiguration using actual type

kron4eg:
To avoid mistakes in KubeletConfiguration, use static type to generate YAML 

**Optional Release Note**:
```release-note
NONE
```

### DIFF
--- a/pkg/userdata/centos/provider.go
+++ b/pkg/userdata/centos/provider.go
@@ -233,31 +233,7 @@ write_files:
 
 - path: "/etc/kubernetes/kubelet.conf"
   content: |
-    kind: KubeletConfiguration
-    apiVersion: kubelet.config.k8s.io/v1beta1
-    cgroupDriver: systemd
-    clusterDomain: cluster.local
-    clusterDNS:
-    {{- range .DNSIPs }}
-      - "{{ . }}"
-    {{- end }}
-    rotateCertificates: true
-    podManifestPath: /etc/kubernetes/manifests
-    readOnlyPort: 0
-    featureGates:
-      RotateKubeletServerCertificate: true
-    serverTLSBootstrap: true
-    rotateCertificates: true
-    authorization:
-      mode: Webhook
-    authentication:
-      x509:
-        clientCAFile: /etc/kubernetes/pki/ca.crt
-      webhook:
-        enabled: true
-      anonymous:
-        enabled: false
-    protectKernelDefaults: true
+{{ kubeletConfiguration "cluster.local" .DNSIPs | indent 4 }}
 
 - path: "/etc/kubernetes/pki/ca.crt"
   content: |
--- a/pkg/userdata/helper/kubelet.go
+++ b/pkg/userdata/helper/kubelet.go
@@ -26,6 +26,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeletv1b1 "k8s.io/kubelet/config/v1beta1"
+	"k8s.io/utils/pointer"
 	kyaml "sigs.k8s.io/yaml"
 )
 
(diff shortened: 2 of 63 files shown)


pkg/userdata/centos/provider.go
```diff
@@ -233,31 +233,7 @@ write_files:
...
-    kind: KubeletConfiguration
-    apiVersion: kubelet.config.k8s.io/v1beta1
-    cgroupDriver: systemd
-    clusterDomain: cluster.local
-    clusterDNS:
-    {{- range .DNSIPs }}
-      - "{{ . }}"
-    {{- end }}
-    rotateCertificates: true
-    podManifestPath: /etc/kubernetes/manifests
```
kron4eg: `podManifestPath: /etc/kubernetes/manifests`

in type definition actually was
```
	StaticPodPath string `json:"staticPodPath,omitempty"`
```

pkg/userdata/centos/provider.go
```diff
@@ -233,31 +233,7 @@ write_files:
...
-    readOnlyPort: 0
-    featureGates:
-      RotateKubeletServerCertificate: true
-    serverTLSBootstrap: true
-    rotateCertificates: true
```
kron4eg: second time of `rotateCertificates: true`

pkg/userdata/helper/kubelet.go
```diff
@@ -131,6 +134,51 @@ func KubeletSystemdUnit(kubeletVersion, cloudProvider, hostname string, dnsIPs [
 	return b.String(), nil
 }
 
+// kubeletConfiguration returns marshaled kubelet.config.k8s.io/v1beta1 KubeletConfiguration
+func kubeletConfiguration(clusterDomain string, clusterDNS []net.IP) (string, error) {
+	var (
```
moadqassem: Maybe you can use "k8s.io/utils/pointer" instead of having literal vars declaration :)

pkg/userdata/helper/kubelet.go
```diff
@@ -131,6 +134,51 @@ func KubeletSystemdUnit(kubeletVersion, cloudProvider, hostname string, dnsIPs [
...
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "KubeletConfiguration",
+			APIVersion: "kubelet.config.k8s.io/v1beta1",
+		},
+		Authentication: kubeletv1b1.KubeletAuthentication{
+			X509: kubeletv1b1.KubeletX509Authentication{
+				ClientCAFile: "/etc/kubernetes/pki/ca.crt",
+			},
+			Webhook: kubeletv1b1.KubeletWebhookAuthentication{
+				Enabled: &True,
```
moadqassem:
```go
Enabled: pointer.BoolPtr(true),
```

pkg/userdata/helper/kubelet.go
```diff
@@ -131,6 +134,51 @@ func KubeletSystemdUnit(kubeletVersion, cloudProvider, hostname string, dnsIPs [
...
+			},
+			Anonymous: kubeletv1b1.KubeletAnonymousAuthentication{
+				Enabled: &False,
```
moadqassem:
```go
Enabled: pointer.BoolPtr(false),
```

pkg/userdata/helper/kubelet.go
```diff
@@ -131,6 +134,51 @@ func KubeletSystemdUnit(kubeletVersion, cloudProvider, hostname string, dnsIPs [
...
+		},
+		Authorization: kubeletv1b1.KubeletAuthorization{
+			Mode: kubeletv1b1.KubeletAuthorizationModeWebhook,
+		},
+		CgroupDriver:          "systemd",
+		ClusterDNS:            clusterDNSstr,
+		ClusterDomain:         clusterDomain,
+		FeatureGates:          map[string]bool{"RotateKubeletServerCertificate": true},
+		ProtectKernelDefaults: true,
+		ReadOnlyPort:          0,
```
moadqassem: This field here is going to be omitted which means it won't exist in the `/etc/kubernetes/kubelet.conf`, thus I think we can remove it. In addition, we are not passing any values from the template to it which means it will always be 0.

kron4eg: It's more for being explicit / for readability.

pkg/userdata/helper/kubelet.go
kron4eg: done 👍

PR #722, opened 2020-03-20
Add finalizer for rhsm subscription

moadqassem:
Up until now, we never added finalizers on the machine object when RedHat Subscription Manager is enabled for machines running rhel, this PR changes that.

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -578,10 +578,16 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloudprovidertypes.Pr
 		return nil, awsErrorToTerminalError(err, "failed create instance at aws")
 	}
 
+	if pc.OperatingSystem == providerconfigtypes.OperatingSystemRHEL && config.manager != nil {
+		if err := rhsm.AddRHELSubscriptionFinalizer(machine, data.Update); err != nil {
+			return nil, fmt.Errorf("failed adding finlizer: %v", err)
+		}
+	}
+
 	return &awsInstance{instance: runOut.Instances[0]}, nil
 }
 
-func (p *provider) Cleanup(machine *v1alpha1.Machine, _ *cloudprovidertypes.ProviderData) (bool, error) {
+func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.ProviderData) (bool, error) {
 	instance, err := p.get(machine)
 	if err != nil {
 		if err == cloudprovidererrors.ErrInstanceNotFound {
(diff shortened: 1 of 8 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -578,10 +578,16 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloudprovidertypes.Pr
 		return nil, awsErrorToTerminalError(err, "failed create instance at aws")
 	}
 
+	if pc.OperatingSystem == providerconfigtypes.OperatingSystemRHEL && config.manager != nil {
+		if err := rhsm.AddRHELSubscriptionFinalizer(machine, data.Update); err != nil {
```
irozzo-1A: @moadqassem Why not using an admission controller to add the finalizer? I think it is cleaner and if my assumption that this is not provider specific holds it could avoid repetitions.

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -618,6 +624,10 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, _ *cloudprovidertypes.Prov
 		if err := config.manager.UnregisterInstance(*instance.instance.PrivateDnsName); err != nil {
 			return false, fmt.Errorf("failed delete machine %s subscription: %v", machine.Name, err)
 		}
+
+		if err := rhsm.RemoveRHELSubscriptionFinalizer(machine, data.Update); err != nil {
```
irozzo-1A: Same here, cannot we move this in a centralised place to avoid repeating for different providers?

pkg/cloudprovider/provider/aws/provider.go
moadqassem: Regarding this, I totally agree on adding this somewhere centralized where we can easily maintain it and avoid repetition, however, and what I saw as a stopper for this is: 

-  The RHSM is actually an optional manager so users can activate by providing an offline_token and if no offline_token is present then the manager will be off.
- Not all cloud providers support RHEL as an os take for e.g: Hetzner, DO and Packet. 
- If I start doing some filtration based on the cloud provider and act accordingly, I think this is a bad design because we are exposing a cloud provider-specific logic on a more generalized level which is the abstract machine level.

If those factors were not present I would have done it the way you suggested and having them centralized.

pkg/cloudprovider/provider/aws/provider.go
moadqassem: Please refer to the comment about :) 
https://github.com/kubermatic/machine-controller/pull/722#discussion_r395774608

pkg/cloudprovider/provider/aws/provider.go
irozzo-1A: See my comments below.

> Regarding this, I totally agree on adding this somewhere centralized where we can easily maintain it and avoid repetition, however, and what I saw as a stopper for this is:
> 
> * The RHSM is actually an optional manager so users can activate by providing an offline_token and if no offline_token is present then the manager will be off.
The fact of being optional does not impact its location.
> * Not all cloud providers support RHEL as an os take for e.g: Hetzner, DO and Packet.
This logic will apply to RHEL OS only, it's not a property of the cloud provider but of the OS. If a provider does not support this OS yet than the logic won't apply, I don't see problems here.
> * If I start doing some filtration based on the cloud provider and act accordingly, I think this is a bad design because we are exposing a cloud provider-specific logic on a more generalized level which is the abstract machine level.
Well, I think we both agree that conceptually this feature is OS specific and not provider specific. If we couple this logic to the provider because we think one day we could have a provider specific exception I don't think we're doing the best design ever.
> 
> If those factors were not present I would have done it the way you suggested and having them centralized.

moadqassem: Well, I have been looking into it, over the weekend and done some changes regarding where those finalizers should be created and removed :).

moadqassem:
This PR has been deprecated by this PR https://github.com/kubermatic/machine-controller/pull/724

PR #726, opened 2020-03-30
Add vSphere ResourcePool support

### LINKED ISSUE #596: vSphere: Ability to use resource pool
**What feature would you like to be added?**
When deploying to vSphere I would like to define the target resource pool so all the created virtual machines/resources are within this pool. So the VMs can be isolated from the other virtual machines.

xref https://github.com/kubermatic/kubeone/issues/594

### LINKED ISSUE #670: vSphere: Support VM templates
When creating a MachineDeployment for vSphere specifying VM template instead of a VM e.g.

```
          cloudProvider: vsphere
          cloudProviderSpec:
            allowInsecure: false
            cluster: cl-1
            cpus: 1
            datacenter: dc-1
            datastore: ds-1
            folder: /dc-1/vm/
            memoryMB: 2000
            password: ""
            templateVMName: /dc-1/vm/Ubuntu1804Template # Name of a VM template
            username: ""
            vmNetName: ""
            vsphereURL: ""
          operatingSystem: ubuntu
```

 the following error occurs:

`E0128 21:38:57.878868       1 machine.go:361] Failed to reconcile machine "ubuntu-test-64957bbf86-sncrf": An error of type = InvalidConfiguration, with message = failed to create cloned vm: 'error when waiting for result of clone task: A specified parameter was not correct: spec.location.pool' occurred`

The problem is that when cloning a VM template the resource pool needs to be specified, while when cloning a VM it defaults to the storage pool of the VM itself.

https://code.vmware.com/apis/358/vsphere/doc/vim.vm.RelocateSpec.html

maxilampert:
Adds support for resource pools and coincidently adds [support for VM templates](#670 ) . 

Fixes #596

Introduces resourcePool as required field for the machinedeployments

**Optional Release Note**:

```release-note
Add vSphere ResourcePool support
```

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/helper.go
+++ b/pkg/cloudprovider/provider/vsphere/helper.go
@@ -441,7 +441,6 @@ func resolveResourcePoolRef(ctx context.Context, config *Config, session *Sessio
 			return nil, fmt.Errorf("failed to get target resourcepool: %v", err)
 		}
 		return types.NewReference(targetResourcePool.Reference()), nil
-	} else {
-		return nil, nil
 	}
+	return nil, nil
 }
--- a/pkg/cloudprovider/provider/vsphere/helper.go
+++ b/pkg/cloudprovider/provider/vsphere/helper.go
@@ -435,28 +435,13 @@ func getDatastoreFromVM(ctx context.Context, session *Session, vmRef *object.Vir
 }
 
 func resolveResourcePoolRef(ctx context.Context, config *Config, session *Session, vm *object.VirtualMachine) (*types.ManagedObjectReference, error) {
-	var targetResourcePool *object.ResourcePool
-	var isTemplate, err = vm.IsTemplate(ctx)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get vm template property: %v", err)
-	}
 	if config.ResourcePool != "" {
-		targetResourcePool, err = session.Finder.ResourcePool(ctx, config.ResourcePool)
+		targetResourcePool, err := session.Finder.ResourcePool(ctx, config.ResourcePool)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get target resourcepool: %v", err)
 		}
-	} else if isTemplate {
-		var targetCluster *object.ClusterComputeResource
-		targetCluster, err = session.Finder.ClusterComputeResource(ctx, config.Cluster)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get target cluster: %v", err)
-		}
-		targetResourcePool, err = targetCluster.ResourcePool(ctx)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get target resourcepool from cluster: %v", err)
-		}
+		return types.NewReference(targetResourcePool.Reference()), nil
 	} else {
 		return nil, nil
 	}
-	return types.NewReference(targetResourcePool.Reference()), nil
 }
--- a/pkg/cloudprovider/provider/vsphere/helper.go
+++ b/pkg/cloudprovider/provider/vsphere/helper.go
@@ -84,7 +84,13 @@ func createClonedVM(ctx context.Context, vmName string, config *Config, session 
 		return nil, fmt.Errorf("failed to resolve datastore: %v", err)
 	}
 
+	resourcepoolref, err := resolveResourcePoolRef(ctx, config, session, tpl)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve resourcePool: %v", err)
+	}
+
 	cloneSpec.Location.Datastore = datastoreref
+	cloneSpec.Location.Pool = resourcepoolref
 	// Create a cloned VM from the template VM's snapshot.
 	// We split the cloning from the reconfiguring as those actions differ on the permission side.
 	// It's nicer to tell which specific action failed due to lacking permissions.
@@ -426,4 +432,31 @@ func getDatastoreFromVM(ctx context.Context, session *Session, vmRef *object.Vir
 		return nil, fmt.Errorf("Failed to parse volPath: %s", props.Summary.Config.VmPathName)
 	}
 	return session.Finder.Datastore(ctx, datastorePathObj.Datastore)
+}
+
+func resolveResourcePoolRef(ctx context.Context, config *Config, session *Session, vm *object.VirtualMachine) (*types.ManagedObjectReference, error) {
+	var targetResourcePool *object.ResourcePool
+	var isTemplate, err = vm.IsTemplate(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get vm template property: %v", err)
+	}
+	if config.ResourcePool != "" {
+		targetResourcePool, err = session.Finder.ResourcePool(ctx, config.ResourcePool)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get target resourcepool: %v", err)
+		}
+	} else if isTemplate {
+		var targetCluster *object.ClusterComputeResource
+		targetCluster, err = session.Finder.ClusterComputeResource(ctx, config.Cluster)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get target cluster: %v", err)
+		}
+		targetResourcePool, err = targetCluster.ResourcePool(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get target resourcepool from cluster: %v", err)
+		}
+	} else {
+		return nil, nil
+	}
+	return types.NewReference(targetResourcePool.Reference()), nil
 }
--- a/pkg/cloudprovider/provider/vsphere/provider.go
+++ b/pkg/cloudprovider/provider/vsphere/provider.go
@@ -65,6 +65,7 @@ type Config struct {
 	Datacenter       string
 	Cluster          string
 	Folder           string
+	ResourcePool     string
 	Datastore        string
 	DatastoreCluster string
 	AllowInsecure    bool
@@ -164,6 +165,11 @@ func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *providerconfigt
 		return nil, nil, nil, err
 	}
 
+	c.ResourcePool, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.ResourcePool)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
 	c.Datastore, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Datastore)
 	if err != nil {
 		return nil, nil, nil, err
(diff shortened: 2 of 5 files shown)


pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -226,6 +232,10 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 		return fmt.Errorf("a vm %s/%s already exists", config.Folder, spec.Name)
 	}
 
+	if _, err := session.Finder.ResourcePool(ctx, config.ResourcePool); err != nil {
```
irozzo-1A: We should validate that the resource pool exists only when it is specified (i.e. `if config.ResourcePool !=  ""` to be clearer) not to break backward compatibility.

moadqassem:
@maxilampert So the added logic needs to be added in the tests too, this field `resourcePool` should be added in the test templates: 
https://github.com/maxilampert/machine-controller/blob/add-vsphere-resourcepool/test/e2e/provisioning/testdata/machinedeployment-vsphere-datastore-cluster.yaml#L27

https://github.com/maxilampert/machine-controller/blob/add-vsphere-resourcepool/test/e2e/provisioning/testdata/machinedeployment-vsphere.yaml#L27

Also please validate if there is a value in there or not

pkg/cloudprovider/provider/vsphere/provider.go
maxilampert: okay makes sense and is changed

maxilampert:
@irozzo-1A  I changed the logic a bit so that if no resourcePool variable is specified that the resource pool of the cluster will be used, to still allow VM templates to work (they require a resource pool) and the cluster is a mandatory field so I guessed that it would make sense that way. I tested it with VM templates, regular VMs, with and without resourcePool variable and it worked in all cases correctly.

@moadqassem Will have a look at the tests soon because I never used tests before :) 
> Also please validate if there is a value in there or not, if there is no value log it and disable your logic.

Not sure I understand the request, please see my above explanation about why we need the resource pool in general to support both regular VM and template clones

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -69,10 +69,29 @@ func createClonedVM(ctx context.Context, vmName string, config *Config, session
 		targetVMFolder = datacenterFolders.VmFolder
 	}
 
+	var targetResourcePool *object.ResourcePool
+	if config.ResourcePool != "" {
+		targetResourcePool, err = session.Finder.ResourcePool(ctx, config.ResourcePool)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get target resourcepool: %v", err)
+		}
+	} else {
```
irozzo-1A, line 78: @maxilampert I don't disagree on the principle of defaulting to the root pool associated to the  `ClusterComputeResource`, but this still changes a bit the behaviour. With a regular VM the same pool is used if not otherwise specified. Don't know if someone is relying on this but it is better to stay on the safe side.

What I propose is to use this logic only in case of templates. The code snippet below could be used to obtain the info:

```
	isTemplate, err = vm.IsTemplate(ctx)
	if err != nil {
		return nil, fmt.Errorf("failed to get vm template property: %v", err)
	}

```

maxilampert: @irozzo-1A I restored the old default logic for cloned VMs and moved them into a func to clean up the  `createCloneVM` func. All of the values can be overwritten by the `resourcePool` field, for templates it defaults to the cluster default pool. I think that restores the default behaviour and allows the most possible flexibility for the other options.

In regards to testing I wonder how the infrastructure look like for the E2E tests what ResourcePools to use and what use cases should be verified?

maxilampert:
In regards to testing I wonder how the infrastructure look like for the E2E tests what resource pools to use and what use cases should be verified?

moadqassem:
> In regards to testing I wonder how the infrastructure look like for the E2E tests what resource pools to use and what use cases should be verified?

What I would suggest is, for the resource pool field you add a tag just take a look at this field in the `MachineDeployment` manifests(https://github.com/kubermatic/machine-controller/blob/master/test/e2e/provisioning/testdata/machinedeployment-vsphere.yaml#L38)https://github.com/kubermatic/machine-controller/blob/master/test/e2e/provisioning/testdata/machinedeployment-vsphere.yaml#L38 as. As you can see this field has a tag that will be replaced by a value in the `helper.go`. Or simply add the field over there in the templates and leave it empty. I would try to come up of a scenario in our infra to make this testable.

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -427,3 +433,33 @@ func getDatastoreFromVM(ctx context.Context, session *Session, vmRef *object.Vir
...
+		var targetCluster *object.ClusterComputeResource
+		targetCluster, err = session.Finder.ClusterComputeResource(ctx, config.Cluster)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get target cluster: %v", err)
+		}
+		targetResourcePool, err = targetCluster.ResourcePool(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get target resourcepool from cluster: %v", err)
+		}
+	} else {
```
irozzo-1A, line 458: this can be avoided, we could just not set  the resourcePool and leave it handle the defaulting as we had before

irozzo-1A:
> > In regards to testing I wonder how the infrastructure look like for the E2E tests what resource pools to use and what use cases should be verified?
> 
> What I would suggest is, for the resource pool field you add a tag just take a look at this field in the `MachineDeployment` manifests(https://github.com/kubermatic/machine-controller/blob/master/test/e2e/provisioning/testdata/machinedeployment-vsphere.yaml#L38)https://github.com/kubermatic/machine-controller/blob/master/test/e2e/provisioning/testdata/machinedeployment-vsphere.yaml#L38 as. As you can see this field has a tag that will be replaced by a value in the `helper.go`. Or simply add the field over there in the templates and leave it empty. I would try to come up of a scenario in our infra to make this testable.

I think it would be complicated for him to add and e2e test as he has no access to out infrastructure and cannot know which pool to use, IMHO we could just run the tests we have and verify that there are no regressions. We can add a dedicated e2e in a follow up PR.

pkg/cloudprovider/provider/vsphere/helper.go
maxilampert: okay I changed that now to return nil not sure if that is the correct way of doing it but it correctly worked in my testing and is using the pool of the source VM accordingly

irozzo-1A:
@maxilampert Could you please rebase and update the [documentation](https://github.com/kubermatic/machine-controller/blob/master/docs/vsphere.md#provider-configuration)? After this for me it's ready to go ;-).

P.S. I updated the description of the PR not to close #670, because it's true that technically this enable the support of VM Templates, but we still need to invest some time in documentation and automated tests before advertise the feature.

moadqassem:
@maxilampert atm the kubevirt tests for your PR will keep failing due some issues on our side. That's why I stopped the testing on your PR ;)

maxilampert:
@irozzo-1A I added the information to the docs. If I should extend that more let me know. While writing the docs I noticed something else. 

The `cluster` parameter has no technical use currently(besides in my code now that if the VM is a template then select the cluster pool which could also refer to host if it is no cluster). I can specify a different cluster but the VM will still be cloned to the location of the templateVM and I wonder what is the idea behind it that the parameter is mandatory and exists in general? 

The only reason I could think of is that it checks in `validate` the existence of the cluster. What that causes it that it blocks deployment of machines to standalone hosts that don't belong to a cluster, which may or may not be intentional. 

The reason I'm asking is that the resourcePool code uses the given resourcePool path and if that is in a cluster then it will use the cluster if not then it won't but it never considers the `cluster` parameter in finding the pools that are exclusive to the given cluster. 

I could workaround the cluster limitation in my tests by creating an empty cluster to satisfy the validate but then specify the `resourcePool` to the default pool of a standalone host and now machines are created successfully on that standalone host.

So I wonder what your opinion is on that?

irozzo-1A:
@maxilampert you are raising a good point. I don't have a definitive answer, but find below my opinion.

> The `cluster` parameter has no technical use currently(besides in my code now that if the VM is a template then select the cluster pool which could also refer to host if it is no cluster). I can specify a different cluster but the VM will still be cloned to the location of the templateVM and I wonder what is the idea behind it that the parameter is mandatory and exists in general?
>

The only usage I see right now is to add a label for prometheus (see https://github.com/kubermatic/machine-controller/blame/master/pkg/cloudprovider/provider/vsphere/provider.go#L566), but when the support for vSphere was originally introduced the cluster was used to determine the pool similarly to what you did in your PR (see https://github.com/kubermatic/machine-controller/commit/e1e3a668fda5715ad84fd76210939ca33c93204f#diff-9c179e2498b1c9b419b34e11831d3997R69). I cannot tell why the resourcePool was not given directly in input instead.
 
> The only reason I could think of is that it checks in `validate` the existence of the cluster. What that causes it that it blocks deployment of machines to standalone hosts that don't belong to a cluster, which may or may not be intentional.
> 
> The reason I'm asking is that the resourcePool code uses the given resourcePool path and if that is in a cluster then it will use the cluster if not then it won't but it never considers the `cluster` parameter in finding the pools that are exclusive to the given cluster.
> 
> I could workaround the cluster limitation in my tests by creating an empty cluster to satisfy the validate but then specify the `resourcePool` to the default pool of a standalone host and now machines are created successfully on that standalone host.
>
> So I wonder whats your opinion is on that?

As I mentioned above the reason is mainly historical, I think that the best approach would be to deprecate the `Cluster` parameter, that should not have a big impact as it's mostly not used apart for Prometheus labels, and remove it or make it optional in future releases.  I have still to discuss this with the team though.

Regarding your PR I think that in light of the findings it would be better to avoid this defaulting mechanism based on the cluster as it would make more difficult to get rid of it or having it optional. I would rather return an error in case a VM Template is used and no `resourcePool` is specified. WDYT?

Thanks again for pointing out this ;-)

maxilampert:
@irozzo-1A I agree on the point to deprecate `cluster` as there is no need for it besides the prometheus values, but these could be still filled by querying the vCenter API and put in either the cluster or eventually the host value where the VM is running on.

Regarding this PR now I think there are 3 ways to fix it:

1. If VM templates should be supported in the long-term I think that we should check in the validate  func if the VM is a template. If that is true `resourcePool` becomes a mandatory value if not it stays optional and the pool will be taken from the source VM, and can be overwritten by specifying `resourcePool`.
This will basically move the template check from the helper func to the validate phase so it fails before even trying to make changes which I think is a good idea. Also we don't have a breaking change in behaviour to the current functionality because we don't introduce a new mandatory value by default.

Another PR then should be made to set `cluster` as optional to allow above functionality to work without having to fill in a dummy cluster value or remove it altogether, but to maybe still fill the prometheus label in another way or remove it as well. I can't tell if anyone is using that value for anything. 

2. As suggested by you I can just adjust the helper func to return an error if the templateVM is a VM template.

3. Making `resourcePool` a mandatory variable in the future and always force users to provide the pool to use to remove all the extra logic required in option 1 but I breaks current behaviour for non templates and adds another mandatory variable.

I personally would go for option one because I would like to see both template and resource pool support going forward and I think that would be good solution to solve it without introducing major code changes nor adding any mandatory values by default that could be problematic.  Let me know what you think :)

irozzo-1A:
@maxilampert  I agree that makes more sense to do the check in the validation, but the support for VM templates is not a high priority for us for the moment.

I think that makes more sense to reduce the scope in order to integrate your PR quickly as the resource pool support is something very valuable.

My recommendation is to completely remove the check on the VM template property. I will do a PR to make explicit that VM templates are not supported for the moment and we will cover it with a dedicated PR including the necessary documentation and testing (contributions are welcome of course ;-)).

maxilampert:
@irozzo-1A okay so I removed the template support now

PR #733, opened 2020-04-02
Update the examples folder

moadqassem:
Updating the example folder

```release-note
None
```


moadqassem:
@youssefazrak Here, I have updated the exmples folder as you requests ;)

PR #736, opened 2020-04-03
Improve vsphere doc

irozzo-1A:

Fixes #

Check that links are not broken and how it's rendered [here](https://github.com/irozzo-1A/machine-controller/blob/improve-vsphere-doc/docs/vsphere.md)

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/docs/vsphere.md
+++ b/docs/vsphere.md
@@ -1,23 +1,267 @@
 # VMware vSphere
 
+## Supported versions
+
+* 6.5
+* 6.7
+
+## Template VMs preparation
+
 To use the machine-controller to create machines on VMWare vsphere, you must first
-create a template.
+create a VM to be used as a template.
 
-Ubuntu & CoreOS:
+*Note that:* 
+`template VMs` in this document refers to regular VMs and not
+[VM Templates][vm_templates] according to vSphere terminology.
+The difference is quite subtle, but VM Templates are not supported yet by
+`machine controller`.
 
-1. Go into the VSphere WebUI, select your datacenter, right click onto it and choose "Deploy OVF Template"
-2. Fill in the "URL" field with the appropriate url:
-    * Ubuntu: `https://cloud-images.ubuntu.com/releases/18.04/release/ubuntu-18.04-server-cloudimg-amd64.ova`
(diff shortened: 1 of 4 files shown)


irozzo-1A:
> I think there should be some consistency with code blocks. Ether indentation or backticks should be used to indicate codeblock, not mix of them

@kron4eg I'm consistently using back-ticks for codeblocks, the indentation you mentioned is because it's inside a numbered list

docs/vsphere.md
```diff
@@ -1,23 +1,267 @@
...
+
+CentOS 7 image can be found at the following link: <https://cloud.centos.org/centos/7/images/CentOS-7-x86_64-GenericCloud.qcow2r>.
+
+Follow [qcow2](#create-template-vm-from-qcow2) template VM creation guide.
+
+## Provider configuration
+
+VSphere provider accepts the following configuration parameters:
+
+```yaml
```
kron4eg, line 222: * backticks
* syntax hint
* no indent

docs/vsphere.md
```diff
@@ -1,23 +1,267 @@
...
+Prerequisites:
+
+* [GOVC](https://github.com/vmware/govmomi/tree/master/govc): tested on version 0.22.1
+* [jq](https://stedolan.github.io/jq/)
+
+Procedure:
+
+1. Download the `OVA` for the targeted OS.
+
+    ```
```
kron4eg, line 44: * backticks
* no syntax hint
* indent

plus bunch of same blocks

docs/vsphere.md
irozzo-1A: no indent because we are not inside a numbered list

docs/vsphere.md
irozzo-1A: > * backticks
this is consistently used in the whole document
> * no syntax hint
no synthax hint because here we have some commands, I could put `bash` but does not provide much value as it is not something structured like `json`, `yaml`, etc.
> * indent
because we are inside a numbered list
> 
> plus bunch of same blocks
The same apply for all other blocks

Issue #727, opened 2020-03-31
Upgrade KubeVirt to the latest version


mfranczy:
Since we want to sell KubeVirt to potential customers it would be good to update the KubeVirt version to the latest stable https://github.com/kubevirt/kubevirt/releases/tag/v0.26.4

moadqassem:
The kubevirt client-go has been updated to 0.28.0 and however the operator and the api are not upgraded due to this issue https://github.com/kubevirt/kubevirt/issues/3009 . once this issue is solved we can upgrade kubevirt to the same version as the client.

mfranczy:
@moadqassem the fix PR has been merged, could you upgrade kubevirt when you have some free time? Thanks.

PR #739, opened 2020-04-08
adding kubevirt-operator-0.26.4.yaml

moadqassem:
Upgrading kubevirt to v0.26.4
Fixes kubermatic/machine-controller#727

**Optional Release Note**:

```release-note
None
```


kron4eg:
I think it’s safe to override failing job since it’s not connected to the change.

moadqassem:
@kron4eg Yeah this PR has nothing to do with the change, however, I have a feeling why coreos is failing and I think that's related to the ignition feature gate in the kubevirt cluster

PR #741, opened 2020-04-15
Upgrade kubevirt go-client

moadqassem:
Upgrading k8s version to 1.16.4 and controller runtime to 0.4.0. In addition to upgrading kubevirt go client to the latest stable release.

At the moment there are some technical issues with kubevirt running coreos vmi, where the ignition support is somehow compromised. See this issue https://github.com/kubevirt/kubevirt/issues/3009. Once this problem is fixed we can safely upgrade the operator to the latest. 
 
**Optional Release Note**:
```release-note
None
```


pkg/admission/machines.go
```diff
@@ -63,7 +63,7 @@ func (ad *admissionData) mutateMachines(ar admissionv1beta1.AdmissionReview) (*a
 		// * oldMachine has Initializers on it
```
xrstf, line 63: This comment isn't true anymore.

Also, can you describe why you removed the finalizer handling here?

pkg/apis/cluster/v1alpha1/migrations/migrations.go
```diff
@@ -141,26 +142,34 @@ func MigrateMachinesv1Alpha1MachineToClusterv1Alpha1MachineIfNecessary(
 	kubeClient kubernetes.Interface,
 	providerData *cloudprovidertypes.ProviderData) error {
 
-	err := client.Get(ctx, types.NamespacedName{Name: machines.CRDName}, &apiextensionsv1beta1.CustomResourceDefinition{})
-	if err != nil {
-		if kerrors.IsNotFound(err) {
-			klog.Infof("CRD %s not present, no migration needed", machines.CRDName)
-			return nil
+	for try := 0; try < 15; try++ {
```
xrstf, line 145: There *has* to be a better way to wait. I'm somewhat sure that there are helpers for waiting and exponential backoffs and all that stuff.

pkg/admission/machines.go
kron4eg: @xrstf not finalizer but initializer. Initializers has been removed from upstream k8s.io/api.

xrstf: Oh mea culpa. Still, the comment is now wrong, no?

moadqassem: Yeah that's actually my nad. the commit message is wrong :/

PR #744, opened 2020-04-20
Use addresses instead of node ref for rhel aws instances

moadqassem:
Attaching the Node Address Type to the created machines, and cleanup rhel subscription based on the private dns for aws cloud provider. 

Fixes kubermatic/kubermatic#5307

**Optional Release Note**:

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/provider/azure/provider.go
+++ b/pkg/cloudprovider/provider/azure/provider.go
@@ -40,6 +40,7 @@ import (
 	"github.com/kubermatic/machine-controller/pkg/providerconfig"
 	providerconfigtypes "github.com/kubermatic/machine-controller/pkg/providerconfig/types"
 
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog"
 	"k8s.io/utils/pointer"
@@ -91,11 +92,11 @@ type config struct {
 
 type azureVM struct {
 	vm          *compute.VirtualMachine
-	ipAddresses []string
+	ipAddresses map[string]v1.NodeAddressType
 	status      instance.Status
 }
 
-func (vm *azureVM) Addresses() []string {
+func (vm *azureVM) Addresses() map[string]v1.NodeAddressType {
 	return vm.ipAddresses
 }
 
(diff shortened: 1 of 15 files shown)


moadqassem:
cc @xmudrii: here I have made this PR to attach the Node Address Type for the addresses ;)

xmudrii:
/hold
review in progress

pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -320,15 +323,18 @@ func getNICIPAddresses(ctx context.Context, c *config, ifaceName string) ([]stri
...
 			}
 
 			internalIPs, err := getInternalIPAddresses(ctx, c, ifaceName, name)
 			if err != nil {
 				return nil, fmt.Errorf("failed to retrieve internal IP string for IP %q: %v", name, err)
 			}
-
-			ipAddresses = append(ipAddresses, internalIPs...)
+			for _, ip := range internalIPs {
+				ipAddresses[ip] = v1.NodeExternalIP
```
xmudrii, line 336: Should we use `v1.NodeInternalIP` since we are talking about internal IPs? Or it works a bit differently for Azure?

pkg/cloudprovider/provider/gce/instance.go
```diff
@@ -56,10 +58,10 @@ func (gi *googleInstance) ID() string {
 }
 
 // Addresses implements instance.Instance.
-func (gi *googleInstance) Addresses() []string {
-	var addrs []string
+func (gi *googleInstance) Addresses() map[string]v1.NodeAddressType {
+	addrs := map[string]v1.NodeAddressType{}
 	for _, ifc := range gi.ci.NetworkInterfaces {
-		addrs = append(addrs, ifc.NetworkIP)
+		addrs[ifc.NetworkIP] = v1.NodeInternalIP
```
xmudrii, line 64: Doesn't the address type depend on if we assign a public IP address to the instance?

pkg/cloudprovider/provider/azure/provider.go
moadqassem: You are right, copy/paste mistake :)

pkg/cloudprovider/provider/gce/instance.go
moadqassem: I am not quite sure about this since the documentation for this field states clearly that, `NetworkIP` in the `NetworkInterface` is the internal ip Address of the instance. 
https://godoc.org/google.golang.org/api/compute/v1#NetworkInterface

PR #746, opened 2020-04-27
Add --volume-plugin-dir kubelet flag

kron4eg:
Default plugin dir "/usr/libexec/kubernetes/kubelet-plugins/volume/exec/"
is not writable on coreos/flatcat. So we use other well known location
`/var/lib/kubelet/volumeplugins`

Fixes #

**Optional Release Note**:

```release-note
always set --volume-plugin-dir=/var/lib/kubelet/volumeplugins
```


pkg/userdata/userdata.go
```diff
@@ -0,0 +1 @@
+package userdata
```
xmudrii, line 1: Why do we need this empty package?

kron4eg: sorry, it was a forgotten leftover

xrstf:
I disagree with overriding, but because our vSphere is currently down and people cannot wait for this and I am super super sure that someone(tm) will later test that this does not break vSphere support

PR #748, opened 2020-04-29
adding support for flatcar in azure

moadqassem:
Adding support for Flatcar linux to Azure cloud provider.

Fixes kubermatic/machine-controller#717

**Optional Release Note**:

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/provider/azure/provider.go
+++ b/pkg/cloudprovider/provider/azure/provider.go
@@ -489,11 +489,7 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloudprovidertypes.Pr
 	}
 	tags[machineUIDTag] = to.StringPtr(string(machine.UID))
 
-	adminUserName := string(providerCfg.OperatingSystem)
-	if adminUserName == "coreos" || adminUserName == "flatcar" {
-		// CoreOS uses core by default everywhere, so we adhere to that
-		adminUserName = "core"
-	}
+	adminUserName := getOSUsername(providerCfg.OperatingSystem)
 	storageProfile, err := getStorageProfile(config, providerCfg)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get StorageProfile: %v", err)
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -476,7 +490,7 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloudprovidertypes.Pr
 	tags[machineUIDTag] = to.StringPtr(string(machine.UID))
 
 	adminUserName := string(providerCfg.OperatingSystem)
```
kron4eg, line 492: kinda funny Operating System variable name :)

Probably getting default / hardcoded login based on the input OS deserves dedicated public function

moadqassem: I am not sure tbh since my PR has nothing to do with it :). However, why do you think this needs a public function?

kron4eg: even if public function is questionable, the var name `adminUserName` which actually hold name of the OS but not username, needs to be changed (but adminUserName is also needed of course).

xrstf: The variable does not hold the OS name, it holds the user name. The user name is just initialized with the OS name. It makes perfect sense. By your logic you cannot have `a := 3`, but it needs to be `three := 3` ;-)

moadqassem: Yeah that's correct. I didn't like it either tbh but I didn't want to fiddle with something that my task has nothing to do with and it is not failing. However, since someone shed the light on it I will change it into a private function

moadqassem: @xrstf what @kron4eg meant is, having a dedicated function for this would be cleaner. because if you take a look of how this adminUserName is used and validated can be moved into a function called e.g: osUsername(), and there the validation is done too

PR #755, opened 2020-05-14
Fix setup script permissions

kron4eg:
Fixed too wide permissions of the /opt/bin/setup

```release-note
NONE
```


kron4eg:
/hold cancel

3 days have passed, removing hold

PR #758, opened 2020-05-15
Set Kubelet node-ip to the instance default interface IP

### LINKED ISSUE #731: Limit kubelet to first network interface
User plans to attach additional network interfaces afterwards to nodes that have been provisioned by machine-controller. Ensure that the correct network is used for Kubernetes, make sure that the kubelet only uses the first interface that have been attached on node provisioning.

At least the following flags should be set to the first NICs address:
`--address`
`--node-ip`

moelsayed:

Fixes #731

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/pkg/userdata/helper/helper.go
+++ b/pkg/userdata/helper/helper.go
@@ -140,3 +140,42 @@ https_proxy=%s
 NO_PROXY=%s
 no_proxy=%s`, proxy, proxy, proxy, proxy, noProxy, noProxy)
 }
+
+func SetupNodeIPEnvScript() string {
+	return `#!/usr/bin/env bash
+echodate() {
+  echo "[$(date -Is)]" "$@"
+}
+
+
+# try to figure out the default gw interface
+DEFAULT_INTERFACE=$(ip route | grep default | grep -oP "dev \K\S+")
+
+if [ -z ${DEFAULT_INTERFACE} ]
+then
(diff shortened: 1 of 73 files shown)


pkg/userdata/helper/helper.go
```diff
@@ -140,3 +140,32 @@ https_proxy=%s
...
+DEFAULT_IFC_IP=$(ip -o  route get 1 | grep -oP "src \K\S+")
+
+if [ -z "${DEFAULT_IFC_IP}" ]
+then
+	echodate "Failed to get IP address for the default route interface"
+	exit 1
+fi
+
+# write the nodeip_env file
+if grep -q coreos /etc/os-release
```
irozzo-1A, line 160: Minor, I would make the test a bit more robust, e.g.
```
grep -q '^ID=coreos$\|^ID_LIKE=coreos$' /etc/os-release
```
or
```
grep -q '^ID=coreos$\|^ID=flatcar$' /etc/os-release
```

moelsayed: I usually try to simplify the bash code as much as possible. Matching for `coreos` is the simplest way to do this.

irozzo-1A: ok, in any case the risk of false positive is very low and we can change the test later in case `ID_LIKE=coreos` for flatcar will be dropped in future.

PR #759, opened 2020-05-19
Bind docker-ce-cli to the same version as docker-ce

xmudrii:

This PR binds the `docker-ce-cli` package to the same version as the `docker-ce` package for Ubuntu, CentOS, and RHEL. As far as I know, there's nothing to do for CentOS/Flatcar and SLES. This is supposed to fix the incompatibilities between CLI and daemon.

Before this fix, the health-monitor script was logging the following error:

```
health-monitor.sh[23271]: Error response from daemon: client version 1.40 is too new. Maximum supported API version is 1.39
```

Relevant: https://github.com/kubermatic/kubeone/pull/896

**Optional Release Note**:
```release-note
Bind docker-ce-cli to the same version as docker-ce
```

/assign @kron4eg

### DIFF
--- a/pkg/userdata/ubuntu/provider.go
+++ b/pkg/userdata/ubuntu/provider.go
@@ -247,6 +247,7 @@ write_files:
     swapoff -a
 
     export CR_PKG='docker-ce=5:18.09.9~3-0~ubuntu-bionic'
+    export CR_CLI_PKG='docker-ce-cli=5:18.09.9~3-0~ubuntu-bionic'
     export DEBIAN_FRONTEND=noninteractive
 
     apt-get -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install -y \
@@ -263,6 +264,7 @@ write_files:
       socat \
       util-linux \
       ${CR_PKG} \
+      ${CR_CLI_PKG} \
       {{- if eq .CloudProviderName "vsphere" }}
       open-vm-tools \
       {{- end }}
@@ -271,6 +273,7 @@ write_files:
 {{- /* If something failed during package installation but docker got installed, we need to put it on hold */}}
     apt-mark hold docker.io || true
     apt-mark hold docker-ce || true
+    apt-mark hold docker-ce-cli || true
 
     # Update grub to include kernel command options to enable swap accounting.
     # Exclude alibaba cloud until this is fixed https://github.com/kubermatic/machine-controller/issues/682
--- a/pkg/userdata/ubuntu/testdata/dist-upgrade-on-boot.yaml
+++ b/pkg/userdata/ubuntu/testdata/dist-upgrade-on-boot.yaml
@@ -140,6 +140,7 @@ write_files:
     swapoff -a
 
     export CR_PKG='docker-ce=5:18.09.9~3-0~ubuntu-bionic'
+    export CR_CLI_PKG='docker-ce-cli=5:18.09.9~3-0~ubuntu-bionic'
     export DEBIAN_FRONTEND=noninteractive
 
     apt-get -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install -y \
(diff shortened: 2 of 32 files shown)


xmudrii:
/hold
to prevent retests

pkg/userdata/ubuntu/provider.go
```diff
@@ -247,6 +247,7 @@ write_files:
     swapoff -a
 
     export CR_PKG='docker-ce=5:18.09.9~3-0~ubuntu-bionic'
```
irozzo-1A, line 249: For k8s 1.17 should we also pin containerd to version `1.2.10-3` as described [here?](https://v1-17.docs.kubernetes.io/docs/setup/production-environment/container-runtimes/#docker)

irozzo-1A: I think that exporting and env variable here is useless, why don't we put the package directly in the install command?

pkg/userdata/ubuntu/testdata/dist-upgrade-on-boot.yaml
```diff
@@ -156,6 +157,7 @@ write_files:
       socat \
       util-linux \
       ${CR_PKG} \
+      ${CR_CLI_PKG} \
       ipvsadm
     apt-mark hold docker.io || true
```
irozzo-1A, line 162: I think this line is there for historical reasons and could be removed, isn't it?

pkg/userdata/ubuntu/provider.go
xmudrii: > I think that exporting and env variable here is useless, why don't we put the package directly in the install command?

I was just following what has been done for `CR_PKG`. It's the same situation and I'm not really sure why we have that environment variable.

> For k8s 1.17 should we also pin containerd to version 1.2.10-3 as described here?

I agree that we should bind containerd as well, but we need to find the compatibility matrix, which I wasn't able to find.

pkg/userdata/ubuntu/testdata/dist-upgrade-on-boot.yaml
```diff
@@ -156,6 +157,7 @@ write_files:
...
     apt-mark hold docker-ce || true
```
irozzo-1A, line 163: we should add `docker-ce-cli` as well

pkg/userdata/ubuntu/provider.go
irozzo-1A: > I was just following what has been done for CR_PKG. It's the same situation and I'm not really sure why we have that environment variable.

Not sure too, but there's nothing relying on it in the script, I think we can get rid of it safely. If you don't want to do it in this PR I will open a follow-up PR.

> I agree that we should bind containerd as well, but we need to find the compatibility matrix, which I wasn't able to find.

I see your point, I did not find supported versions in k8s release notes too. Let's keep it unpinned for the moment.

pkg/userdata/ubuntu/testdata/dist-upgrade-on-boot.yaml
xmudrii: We have agreed to keep it for now and get rid of it in a follow up.

xmudrii:
/retest
constantly no logs

pkg/userdata/ubuntu/testdata/dist-upgrade-on-boot.yaml
xmudrii: That should be done now.

pkg/userdata/ubuntu/provider.go
```diff
@@ -271,6 +273,7 @@ write_files:
 {{- /* If something failed during package installation but docker got installed, we need to put it on hold */}}
     apt-mark hold docker.io || true
     apt-mark hold docker-ce || true
+    apt-mark hold docker-ce-cli || true
```
irozzo-1A, line 276: This is not a blocker but I would have done it in one single line:

```
apt-mark hold docker-ce docker-ce-cli || true
```

irozzo-1A:
LGTM

pkg/userdata/ubuntu/provider.go
xmudrii: That's a good point, but let's leave it for a follow-up along with `CR_PKG`. Tests are already running and I think we should try to get this merged as soon as possible.

xmudrii:
/test pull-machine-controller-e2e-deployment-upgrade
/test pull-machine-controller-yamllint
/test pull-kubermatic-machine-controller-validate-prow-yaml
re-running the jobs because they never started

PR #760, opened 2020-05-19
Unify the way kubelet works on Flatcar/CoreOS with other OS

kron4eg:
Unify the way kubelet works on Flatcar/CoreOS with other OS, now it uses exactly the same kubelet flags and binary download procedures.

Because of the above, hyperkube image is no longer needed so its handling is removed and CLI flag is DEPRECATED and NOOP.

Part of the https://github.com/kubermatic/kubeone/issues/125

**Optional Release Note**:
```release-note
* `-node-hyperkube-image` flag is deprecated and has no effect
* Flatcar/CoreOS will use "normal" kubelet like all other OS
```

### DIFF
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -163,10 +163,15 @@ func main() {
 	flag.StringVar(&nodeInsecureRegistries, "node-insecure-registries", "", "Comma separated list of registries which should be configured as insecure on the container runtime")
 	flag.StringVar(&nodeRegistryMirrors, "node-registry-mirrors", "", "Comma separated list of Docker image mirrors")
 	flag.StringVar(&nodePauseImage, "node-pause-image", "", "Image for the pause container including tag. If not set, the kubelet default will be used: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/")
-	flag.String("node-hyperkube-image", "k8s.gcr.io/hyperkube-amd64", "Image for the hyperkube container excluding tag. (DEPRECATED, NOOP)")
+	flag.String("node-hyperkube-image", "", "Image for the hyperkube container excluding tag. (DEPRECATED, NOOP)")
 	flag.BoolVar(&nodeCSRApprover, "node-csr-approver", false, "Enable NodeCSRApprover controller to automatically approve node serving certificate requests.")
 
 	flag.Parse()
+
+	if fv := flag.Lookup("node-hyperkube-image"); fv.DefValue != fv.Value.String() {
+		klog.Warning("-node-hyperkube-image flag IS DEPRECATED and has no EFFECT ANYMORE")
+	}
+
 	kubeconfig = flag.Lookup("kubeconfig").Value.(flag.Getter).Get().(string)
 	masterURL = flag.Lookup("master").Value.(flag.Getter).Get().(string)
 
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -38,7 +38,6 @@ import (
 	"k8s.io/client-go/tools/leaderelection"
 	"k8s.io/client-go/tools/leaderelection/resourcelock"
 
-	"github.com/docker/distribution/reference"
 	"github.com/heptiolabs/healthcheck"
 	"github.com/oklog/run"
 	"github.com/prometheus/client_golang/prometheus"
@@ -81,7 +80,6 @@ var (
 	nodeInsecureRegistries string
 	nodeRegistryMirrors    string
 	nodePauseImage         string
-	nodeHyperkubeImage     string
 )
 
 const (
@@ -165,7 +163,7 @@ func main() {
 	flag.StringVar(&nodeInsecureRegistries, "node-insecure-registries", "", "Comma separated list of registries which should be configured as insecure on the container runtime")
 	flag.StringVar(&nodeRegistryMirrors, "node-registry-mirrors", "", "Comma separated list of Docker image mirrors")
 	flag.StringVar(&nodePauseImage, "node-pause-image", "", "Image for the pause container including tag. If not set, the kubelet default will be used: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/")
-	flag.StringVar(&nodeHyperkubeImage, "node-hyperkube-image", "k8s.gcr.io/hyperkube-amd64", "Image for the hyperkube container excluding tag.")
+	flag.String("node-hyperkube-image", "k8s.gcr.io/hyperkube-amd64", "Image for the hyperkube container excluding tag. (DEPRECATED, NOOP)")
 	flag.BoolVar(&nodeCSRApprover, "node-csr-approver", false, "Enable NodeCSRApprover controller to automatically approve node serving certificate requests.")
 
 	flag.Parse()
@@ -197,14 +195,6 @@ func main() {
 	}
 	if err := clusterv1alpha1.AddToScheme(scheme.Scheme); err != nil {
 		klog.Fatalf("failed to add clusterv1alpha1 api to scheme: %v", err)
-	}
-	// Check if the hyperkube image has a tag set
-	hyperkubeImageRef, err := reference.Parse(nodeHyperkubeImage)
-	if err != nil {
-		klog.Fatalf("failed to parse --node-hyperkube-image %s: %v", nodeHyperkubeImage, err)
-	}
-	if _, ok := hyperkubeImageRef.(reference.NamedTagged); ok {
-		klog.Fatalf("--node-hyperkube-image must not contain a tag. The tag will be dynamically set for each Machine.")
 	}
 
 	cfg, err := clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
@@ -246,11 +236,10 @@ func main() {
 		skipEvictionAfter:     skipEvictionAfter,
 		nodeCSRApprover:       nodeCSRApprover,
 		node: machinecontroller.NodeSettings{
-			ClusterDNSIPs:  clusterDNSIPs,
-			HTTPProxy:      nodeHTTPProxy,
-			NoProxy:        nodeNoProxy,
-			HyperkubeImage: nodeHyperkubeImage,
-			PauseImage:     nodePauseImage,
+			ClusterDNSIPs: clusterDNSIPs,
+			HTTPProxy:     nodeHTTPProxy,
+			NoProxy:       nodeNoProxy,
+			PauseImage:    nodePauseImage,
 		},
 	}
 	if parsedJoinClusterTimeout != nil {
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,6 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/digitalocean/godo v1.1.3
 	github.com/dnaeon/go-vcr v1.0.1 // indirect
-	github.com/docker/distribution v2.7.1+incompatible
 	github.com/emicklei/go-restful v2.11.2+incompatible // indirect
 	github.com/ghodss/yaml v1.0.0
 	github.com/go-openapi/jsonreference v0.19.3 // indirect
@@ -44,7 +43,6 @@ require (
 	github.com/mailru/easyjson v0.7.1 // indirect
 	github.com/mattbaird/jsonpatch v0.0.0-20171005235357-81af80346b1a
 	github.com/oklog/run v1.0.0
-	github.com/opencontainers/go-digest v1.0.0-rc1 // indirect
 	github.com/packethost/packngo v0.1.1-0.20190410075950-a02c426e4888
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/pborman/uuid v1.2.0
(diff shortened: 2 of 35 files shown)


irozzo-1A:
@kron4eg I think you should take a look at the last commit I pushed to https://github.com/kubermatic/machine-controller/pull/761

cmd/machine-controller/main.go
```diff
@@ -198,14 +196,6 @@ func main() {
...
 		klog.Fatalf("failed to add clusterv1alpha1 api to scheme: %v", err)
 	}
-	// Check if the hyperkube image has a tag set
-	hyperkubeImageRef, err := reference.Parse(nodeHyperkubeImage)
-	if err != nil {
-		klog.Fatalf("failed to parse --node-hyperkube-image %s: %v", nodeHyperkubeImage, err)
-	}
-	if _, ok := hyperkubeImageRef.(reference.NamedTagged); ok {
-		klog.Fatalf("--node-hyperkube-image must not contain a tag. The tag will be dynamically set for each Machine.")
-	}
```
xmudrii, line 208: We should show a warning if the flag is provided that setting the hyperkube image is deprecated.

kron4eg: 👍

irozzo-1A:
@kron4eg looks like this PR introduces some instabilities with the e2e tests, wouldn't be worthy to understand the cause before integrating it?

irozzo-1A:
After some reflection and reading some documentation of container linux I think I don't fully agree that unifying the way we run the kubelet, at least in this direction, is a good thing.

I think that the philosophy of container linux is that applicative code should run in containers. Why you want to change this? What is the main driver of your change?

> The second building block, Docker (docs), is where your applications and code run. It is installed on each Flatcar Container Linux machine. You should make each of your services (web server, caching, database) into a container and connect them together by reading and writing to etcd. You can quickly try out a minimal busybox container in two different ways:

https://docs.flatcar-linux.org/os/quickstart/#container-management-with-docker

This discussion also contains some interesting elements:

https://groups.google.com/forum/#!topic/coreos-dev/FHBdG9B9SAg

kron4eg:
@irozzo-1A 

* We've been running kubeone coreos control-plane nodes like this since the inception
* hyperkube image is deprecated. xref https://github.com/kubernetes/kubeadm/issues/1889

xmudrii:
@irozzo-1A The first problem we have is that the hyperkube is deprecated by the upstream community. I'd also like to mention the following issue and PR:

* https://github.com/kubernetes/kubernetes/issues/81760
* https://github.com/kubernetes/kubernetes/pull/83454

Hyperkube is not part of the Kubernetes repository anymore, neither of the Kubernetes release cycle. Because of this, I strongly agree that we should get rid of hyperkube as soon as possible.

On the other side, I think it doesn't matter how we run it as long as we know it works. The way that @kron4eg proposed in this PR has been used by KubeOne since the beginning, even in production setups, and we never received any complaints.

To my knowledge, there is no best/recommended way to run Kubelet, so it's up to the operators what they prefer. I don't believe that we can compare Kubelet to other applications since it's a bit specific. Kubelet is often referred to as a daemon/node agent.

If we decide to run it in the container, we would have to pay a lot of attention to the security of that container, does it work properly in all setups, and so on.

On the other side, this approach is well-tested. We have to ensure that the image we use have all dependencies preinstalled or to install what's missing. A notable dependency to mention here might be `conntrack` because we had some problems with that one and KubeOne in the past.

irozzo-1A:
@xmudrii @artiom I see your point, and I agree we need to find a solution. It's not clear to me yet if there will be an official kubelet image, and if maintainers of flatcar linux will propose a new solution themselves for the kubelet wrapper:

https://github.com/flatcar-linux/Flatcar/issues/15 

My main concern is about [dependencies](https://github.com/kubernetes/kubernetes/issues/26093) indeed, and in container linux it's even more problematic as there is not package manager. 

> I don't believe that we can compare Kubelet to other applications since it's a bit specific. Kubelet is often referred to as a daemon/node agent.

The kubelet is an application like another, probably not the easiest one to run in a container (https://github.com/kubernetes/kubernetes/issues/4869),  but this is the common approach used so far for container linux and derived distros.

> If we decide to run it in the container, we would have to pay a lot of attention to the security of that container, does it work properly in all setups, and so on.

Note that I did not say we should run the kubelet as a container for all distros, even if probably it would be more portable approach than ensuring dependencies relying on each distribution package management, and downloading the kubelet binary.
Regarding the security aspects I'm not sure about what you meant exactly.

> On the other side, this approach is well-tested. We have to ensure that the image we use have all dependencies preinstalled or to install what's missing. A notable dependency to mention here might be conntrack because we had some problems with that one and KubeOne in the past.

In container linux kubelet was already running in a container since ages, and in OCP 4 RHCOS kubelet still runs in a container, so it is more proven to work than the approach proposed here I think.

kron4eg:
@irozzo-1A please write a proposal, with solution and we will discuss it. In the meantime, this PR is needed for kubeone integration, for avoiding unmaintained hyperkube image and so on.

Maybe in the future there will be some solution, but it's not yet here and we can't wait.

irozzo-1A:
@kron4eg Maybe I'm missing the central point here. Why kubeone would require this PR to be integrated? 

> @irozzo-1A please write a proposal, with solution and we will discuss it. In the meantime, this PR is needed for kubeone integration, for avoiding unmaintained hyperkube image and so on.

My proposal is quite simple, let's keep the kubelet run in a container.  If I'm not mistaken no official kubelet image is provided at the moment so either we keep using the hyperkube image (community maintained but not gone) or we maintain our own image (e.g. https://github.com/poseidon/kubelet#packages).

> Maybe in the future there will be some solution, but it's not yet here and we can't wait.

This PR does not address the problem of the dependencies. I think that the most of the functionalities should work at the moment, but there will be some regressions e.g. no glusterfs client (BTW we should install it for other distros we support too :-)). Maybe the impact is limited, but it's worthy to take this into account.

xmudrii:
@irozzo-1A 

> Maybe I'm missing the central point here. Why kubeone would require this PR to be integrated?

We want to unify the bootstrap scripts and have them in a single place. The plan is that KubeOne imports scripts and bootstrapping logic from machine-controller. You can find more details in the following issue: https://github.com/kubermatic/kubeone/issues/125

> If I'm not mistaken no official kubelet image is provided at the moment

That's right and to my knowledge there are no plans for the official Kubelet image.

> we keep using the hyperkube image (community maintained but not gone) or we maintain our image (e.g. https://github.com/poseidon/kubelet#packages).

Currently, I'm strongly against using the Hyperkube image because of the reasons stated in the previous comment. About maintaining our image, that's probably not needed at all.

> This PR does not address the problem of the dependencies.

I agree regarding dependencies, but that problem also stands if we decide to use containers.

I'll try to find more references about the Hyperkube project and image status, as well as, what's the recommended way to run Kubelet on the Container Linux distros. In meanwhile, I still believe this approach is okay until we don't have a better one.

xmudrii:
@irozzo-1A @kron4eg 

> I'll try to find more references about the Hyperkube project and image status, as well as, what's the recommended way to run Kubelet on the Container Linux distros.

I've searched the Kubernetes docs for a bit. I found that the recommended way to run Kubelet is to use the plain systemd service file. This can be especially important if Kubeadm is being used.

[The `Installing Kubeadm` document](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl) uses the plain systemd unit file to run Kubelet. The systemd unit file can be found in the [k/release repo](https://github.com/kubernetes/release/blob/v0.2.7/cmd/kubepkg/templates/latest/deb/kubelet/lib/systemd/system/kubelet.service).

Besides that, [the `Configuring each kubelet in your cluster using kubeadm` document](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/) explains the relationship between kubeadm, kubelet, and systemd.

I'd like to highlight [the `The kubelet drop-in file for systemd` part of the document](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/kubelet-integration/#the-kubelet-drop-in-file-for-systemd), as well as, the following paragraph:

> Since the kubelet is a daemon, it needs to be maintained by some kind of an init system or service manager. When the kubelet is installed using DEBs or RPMs, systemd is configured to manage the kubelet. You can use a different service manager instead, but you need to configure it manually.

As we use Kubeadm in KubeOne to provision the clusters, I think going forward with this approach is very important for us.

kron4eg:
@xmudrii I've figured out why `pull-machine-controller-e2e-digitalocean` failing and that's quite interesting story.

So kubelet launched from hyperkube under RKT was able to figure out the correct local IPs somehow. But under "direct binary launch" kubelet is not able to detect correct IP, which leads to absence of correct connection between machine and node objects (the recognition of which node equals to which machine it's based on IPs https://github.com/kubermatic/machine-controller/blob/23656f1f2c518c1e555b12362501b835d36794c3/pkg/controller/machine/machine.go#L879-L885). 

In addition generate kubelet CSR contains incorrect IPs and node-csr-approver rightfully refuses to approve it.

Inside the kubeone cluster this is "band-aid" with DO CCM, but in our e2e tests those machines have no CCM, and their kubelets have no `--cloud-provider=external`.

irozzo-1A:
@xmudrii Thx for giving me more context and for your research. There are still some aspects in your reasoning that seemed to me a bit biased and/or unclear.

> Currently, I'm strongly against using the Hyperkube image because of the reasons stated in the previous comment. About maintaining our image, that's probably not needed at all.

The reasons stated in previous comment are not so clear to me. I can guess that the main argument against the usage of the hyperkube image is that it is community managed it can fall behind or being discontinued. This is a fair point. Regarding the comment about maintaning our own image, well it is needed if we decide to go that way :-).

> I agree regarding dependencies, but that problem also stands if we decide to use containers.

Container Linux does not come with any package manager because it relies on using containers to run applications. If we build an image with a debian base for instance we can install DEB packages, as the hyperkube image, and this basically solves the dependencies issue I was referring to. Of course we can still miss some dependencies but this is another kind of problems, and we can solve it easily.

> I've searched the Kubernetes docs for a bit. I found that the recommended way to run Kubelet is to use the plain systemd service file. This can be especially important if Kubeadm is being used.

I agree that using a systemd service is good to ensure that the kubelet is always running. This is orthogonal to using or not a container. Right now the kubelet is running in a container and by a systemd unit.

> The Installing Kubeadm document uses the plain systemd unit file to run Kubelet. The systemd unit file can be found in the k/release repo.

I saw this, but it does not mean that the best approach. I made the example of glusterfs client but there could be other dependencies that are not satisfied (ceph ?) by default in container linux. To be fair looking at the dependencies of the [RPM](https://github.com/kubernetes/release/blob/master/cmd/kubepkg/templates/latest/rpm/kubelet/kubelet.spec#L10) and Debian packages not all dependencies that are installed in the hyperkube image are listed, and it's possible that the one listed here are satisfied by default in container linux. To be verified.

> I'd like to highlight the The kubelet drop-in file for systemd part of the document, as well as, the following paragraph:

> Since the kubelet is a daemon, it needs to be maintained by some kind of an init system or service manager. When the kubelet is installed using DEBs or RPMs, systemd is configured to manage the kubelet. You can use a different service manager instead, but you need to configure it manually.

Here I don't see any argument in favour or against the usage of containers. We cannot use the DEBs or RPMs of course in container linux, and we don't use them even where we could. We can take into account the conventions used by the kubeadm drop-in file and doing something compatible whatever approach we decide to pursuit.

irozzo-1A:
@kron4eg  @xmudrii Let me recap the main options I think we have:

1. Downloading and running the kubelet binary and run it through systemd unit (solution proposed here).
    _Pros:_
    - Already implemented :)
    - We use the same script for all distros

    _Cons:_
    - We cannot easily manage kubelet dependencies as there is no package manager in container linux.
2. Run kubelet in container using hyperkube image through systemd unit:
    _Pros:_
    - We don't need to maintain any image.
    - All dependencies are supposed to be cleared.

    _Cons:_
    - The image is community managed, we have no guarantees on it.
    -  The image contains several binaries and their dependencies, it's quite heavy.
2. Run kubelet in container using an image managed by us through systemd unit:
    _Pros:_
    - We can easily ensure and we have full control on kubelet dependencies .
    - We can generalize this to other distros we support.

    _Cons:_
    - We need to maintain the image and publish it for each kubelet release we support (not for free).
    - More complicated than just running a binary (see kubelet-wrapper).

Let me know if you think I missed something important.

kron4eg:
@irozzo-1A Yes, you’ve missed the main point why it was started. To integrate with kubeone, to avoid maintaining same stuff in 2 places.
Hyperkube image debacle appeared in the middle and is not affecting targen in anyway.

I’m really getting tired of volume of dead demagoguery. There’s nothing to argue about in here. Let’s wrap it up.

irozzo-1A:
@kron4eg Indeed initially I missed some context, then @xmudrii pointed me to the relevant issues.
I think that the point I raised is worth at least being considered and debated on its merits. There is no benefit in raising the tone.

kron4eg:
@irozzo-1A yeah, the discussion here slipped somewhere else, and I apologize for that.

kron4eg:
Given the magnitude of change we can't "just accept" this PR without broader discussion with dev team (especially since this PR breaks DO, and it's not a flake).

irozzo-1A:
> @irozzo-1A yeah, the discussion here slipped somewhere else, and I apologize for that.

no pb mate it arrives, I was a bit too pushy and verbose with my comments ;-)

PR #761, opened 2020-05-20
Clean-up userdata scripts and testdata

irozzo-1A:
Follow up of PR of https://github.com/kubermatic/machine-controller/pull/759 bringing some clean-up.

Fixes #

**Optional Release Note**:

```release-note
NONE
```


pkg/userdata/rhel/provider.go
```diff
@@ -192,19 +192,23 @@ write_files:
...
       sudo \
       socat \
       wget \
       curl \
+      python3-dnf-plugin-versionlock \
       {{- if eq .CloudProviderName "vsphere" }}
       open-vm-tools \
       {{- end }}
       ipvsadm
+    dnf versionlock add docker-ce
```
kron4eg, line 210:
```suggestion
```
I think this line should be deleted

kron4eg:
@irozzo-1A "Clean-up ubuntu" took you too far away :D 
Please drop parts with coreos/flatcar

README.md
```diff
@@ -22,6 +22,15 @@
...
 - Using Ubuntu, CoreOS/RedHat ContainerLinux or CentOS 7 distributions ([not all distributions work on all providers](/docs/operating-system.md))
 
+## Supported Kubernetes versions
+Machine controller tries to follow as close as possible the Kubernetes version
+[support policy](https://kubernetes.io/docs/setup/release/version-skew-policy/).
+
+Currently supported K8S versions are:
+- 1.17
+- 1.16
+- 1.15
```
xmudrii, line 32: Why is 1.18 not supported? 1.18 came out a few weeks ago and we successfully use it for 1.18 clusters in KubeOne since then.

Also, should we have 1.15 on the list since it's not supported by the upstream anymore?

README.md
```diff
@@ -22,6 +22,15 @@
 - Creation of worker nodes on AWS, Digitalocean, Openstack, Azure, Google Cloud Platform, VMWare Vsphere, Linode, Hetzner cloud and Kubevirt (experimental)
 - Using Ubuntu, CoreOS/RedHat ContainerLinux or CentOS 7 distributions ([not all distributions work on all providers](/docs/operating-system.md))
 
+## Supported Kubernetes versions
+Machine controller tries to follow as close as possible the Kubernetes version
```
xmudrii, line 26:
```suggestion
machine-controller tries to follow as close as possible the Kubernetes version
```
Nit: we call it machine-controller elsewhere.

pkg/userdata/ubuntu/provider.go
```diff
@@ -246,11 +246,11 @@ write_files:
...
     swapoff -a
 
-    export CR_PKG='docker-ce=5:18.09.9~3-0~ubuntu-bionic'
-    export CR_CLI_PKG='docker-ce-cli=5:18.09.9~3-0~ubuntu-bionic'
-    export DEBIAN_FRONTEND=noninteractive
+{{- /* We need to explicitly specify docker-ce and docker-ce-cli to the same version.
+	See: https://github.com/docker/cli/issues/2533 */}}
 
-    apt-get -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" install -y \
+    DOCKER_VERSION='5:18.09.9~3-0~ubuntu-bionic'
```
xmudrii, line 252: What's the reason for having the version string starting with `5:` and ending with `~3-0~ubuntu-bionic`? It can be confusing from where this string comes from. It should also work well if we only provide the version.

README.md
irozzo-1A: Depends if we still support it or not I guess, I'm ok to add 1.18 if we support it.

pkg/userdata/ubuntu/provider.go
irozzo-1A: this is the version of the package, I don't know where this convention comes from by I don't think we should only take part of it.

irozzo-1A:
> @irozzo-1A "Clean-up ubuntu" took you too far away :D
> Please drop parts with coreos/flatcar

I agree that this PR is not very cohesive, but I don't consider dropping this a good option. This part was functional to the cause as it avoided me to apply the exact same fixes twice.

I could extract it to a new PR and merge it before this one, but it would cost me some effort as I have to adjust the testdata.

Could you please clarify what's the problem with this concretely speaking?

PR #762, opened 2020-05-20
[release/v1.10] Bind docker-ce-cli to the same version as docker-ce

xmudrii:

https://github.com/kubermatic/machine-controller/pull/759

Bind `docker-ce-cli` to the same version as `docker-ce`.

CentOS is not affected in 1.10 release because we are using the [Docker v1.13 package](https://github.com/kubermatic/machine-controller/blob/04c3f2ab0fe688858b8ac124ce17cf686cd6d9c7/pkg/userdata/centos/provider.go#L187).

**Optional Release Note**:
```release-note
release/v1.10: Bind docker-ce-cli to the same version as docker-ce
```

/assign @kron4eg

### DIFF
--- a/pkg/userdata/rhel/provider.go
+++ b/pkg/userdata/rhel/provider.go
@@ -191,8 +191,8 @@ write_files:
 
     mkdir /etc/docker
 
-    yum install -y docker-ce-3:18.09.1-3.el7 \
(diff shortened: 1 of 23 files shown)


kron4eg:
@xmudrii CentOS is missing

xmudrii:
@kron4eg That's intentional, check the PR description.

PR #764, opened 2020-05-20
[release/v1.8] Bind docker-ce-cli to the same version as docker-ce

xmudrii:

https://github.com/kubermatic/machine-controller/pull/759

Bind `docker-ce-cli` to the same version as `docker-ce`.

CentOS is not affected in 1.10 release because we are using the [Docker v1.13 package](https://github.com/kubermatic/machine-controller/blob/2b8a521f63efb757c091c9355597acccd5a13bdf/pkg/userdata/centos/provider.go#L187).

**Optional Release Note**:
```release-note
release/v1.8: Bind docker-ce-cli to the same version as docker-ce
```

/assign @kron4eg

### DIFF
--- a/pkg/userdata/ubuntu/provider.go
+++ b/pkg/userdata/ubuntu/provider.go
@@ -236,6 +236,7 @@ write_files:
     swapoff -a
 
     export CR_PKG='docker-ce=5:18.09.2~3-0~ubuntu-bionic'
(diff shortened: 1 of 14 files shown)


kdomanski:
No working tests on branch this old, but it is identical to commits on top of newer branches.
/lgtm
/approve

PR #765, opened 2020-05-21
Mutulize flatcar and coreos container linux userdata code to avoid duplications

irozzo-1A:
Flatcar and CoreOS userdata code is basically the same. This PR removes code and testdata duplications.

Fixes #

 31 additions and 5,691 deletions.
**Optional Release Note**:

```release-note
NONE
```


kron4eg:
/hold

#760 introducing removal of RKT and kubelet-wrapper usage, and this PR directly conflict with it, at least wait until #760 is merged to avoid huge mess in git confict form

irozzo-1A:
> #760 introducing removal of RKT and kubelet-wrapper usage, and this PR directly conflict with it, at least wait until #760 is merged to avoid huge mess in git confict form

@kron4eg I don't see any difficult merge here, but as your PR came first I have no pb of taking care of the rebase. BTW #765 on that respect is worst due to the testdata changes.

kdomanski:
Is the similarity causing any technical issues?

kron4eg:
CoreOS container linux's EOL is tomorrow. After some time we could simply remove userdata/coreos package.

What might be interesting is support for RHCOS / FedoraCOS but that's more question to @kdomanski and I really not sure we need it at all 🤷‍♂️

irozzo-1A:
> Is the similarity causing any technical issues?

well, "similar" in an euphemism in this context :-) they are exactly the same apart from the kubelet wrapper path that I factored out. The main advantage is that we avoid double maintenance where is not needed  and at the same time we can get rid of some test data, as it's quite tedious to update it

irozzo-1A:
> CoreOS container linux's EOL is tomorrow. After some time we could simply remove userdata/coreos package.

If the plan is to get rid completely of coreos support and not just deprecate it I agree, I thought we wanted to have a transition phase.

kron4eg:
> If the plan is to get rid completely of coreos support

Even coreos (RH/IBM) planned and executed just that, we can't do anything above it.

>I thought we wanted to have a transition phase.

Never heard about such plans, did you?

irozzo-1A:
@kron4eg I agree that eventually we will get rid of coreos support as it does not make sense to keep support for something unmaintained.

I was not involved in any discussion about a transition phase, but I'm open to discussion of course. I thought that if we drop it brutally customers that did not migrate yet coreos MD will have failures when they update MC, having a transition phase is useful in that respect.

This PR does not change much either if we decide to drop coreos or not.

kdomanski:
We will drop CoreOS entirely within the next quarter or two. CoreOS is therefore frozen in time, while the code for Flatcar might change.

irozzo-1A:
> We will drop CoreOS entirely within the next quarter or two. CoreOS is therefore frozen in time, while the code for Flatcar might change.

I see your point, indeed if we need to introduce a breaking change for Flatcar during the time we still support CoreOS we should add conditions, that is less handy than having them split. On the other hand even if there will be no new CoreOS releases we could still have to do fixes that can profit to both.

Whatever we decide to do the gain/loss is not very big. I still prefer my approach and I think that makes more sense than the copy/paste approach used so far, but if there is consensus on keeping the two separated I'm ok to close this.

PR #778, opened 2020-06-09
Use cloud-init rh_subscription module to register and auto-attach subscriptions

### LINKED ISSUE #777: RHEL subscription repeated multiple times when setup script fails
The setup script for RHEL includes [registration of the machine](https://github.com/kubermatic/machine-controller/blob/master/pkg/userdata/rhel/provider.go#L190) using the subscription-manager. The problem of this operation is that it is not idempotent, if the script fails it gets restarted by systemd resulting in multiple subscriptions for the same machine.

irozzo-1A:
This PR removes the registration of the machine from the setup script (run by a systemd unit) and attempts to use `rh_subscription` cloud-init module as a replacement.

The advantage of doing this is that the module attempts the registration only if the system is not already registered.

Fixes #777

**Optional Release Note**:

```release-note
None
```

### DIFF
--- a/pkg/userdata/rhel/provider.go
+++ b/pkg/userdata/rhel/provider.go
@@ -186,10 +186,6 @@ write_files:
     hostnamectl set-hostname {{ .MachineSpec.Name }}
     {{ end }}
 
-    subscription-manager clean
-    subscription-manager register --username='{{.OSConfig.RHELSubscriptionManagerUser}}' --password='{{.OSConfig.RHELSubscriptionManagerPassword}}'
-    subscription-manager attach --auto
-
     dnf config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
 
     DOCKER_VERSION='18.09.1-3.el7'
@@ -314,6 +310,11 @@ write_files:
   content: |
     [Service]
     EnvironmentFile=-/etc/environment
+
+rh_subscription:
+  username: "{{.OSConfig.RHELSubscriptionManagerUser}}"
+  password: "{{.OSConfig.RHELSubscriptionManagerPassword}}"
+  auto-attach: true
 
 runcmd:
 - systemctl enable --now setup.service
--- a/pkg/userdata/rhel/provider.go
+++ b/pkg/userdata/rhel/provider.go
@@ -312,9 +312,9 @@ write_files:
     EnvironmentFile=-/etc/environment
 
 rh_subscription:
-  username: "{{.OSConfig.RHELSubscriptionManagerUser}}"
-  password: "{{.OSConfig.RHELSubscriptionManagerPassword}}"
-  auto-attach: true
+    username: "{{.OSConfig.RHELSubscriptionManagerUser}}"
+    password: "{{.OSConfig.RHELSubscriptionManagerPassword}}"
+    auto-attach: true
 
 runcmd:
 - systemctl enable --now setup.service
--- a/pkg/userdata/rhel/testdata/kubelet-v1.15-aws.yaml
+++ b/pkg/userdata/rhel/testdata/kubelet-v1.15-aws.yaml
@@ -69,10 +69,6 @@ write_files:
     swapoff -a
 
 
-    subscription-manager clean
-    subscription-manager register --username='' --password=''
-    subscription-manager attach --auto
-
     dnf config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
 
     DOCKER_VERSION='18.09.1-3.el7'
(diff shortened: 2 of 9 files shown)


pkg/userdata/rhel/provider.go
```diff
@@ -186,10 +186,6 @@ write_files:
     hostnamectl set-hostname {{ .MachineSpec.Name }}
     {{ end }}
 
-    subscription-manager clean
```
moadqassem, line 189: Using subscription clean was recommended from RedHat support.

irozzo-1A: This line is the root cause of our issue. Without cleaning the subsequent `subscription-manager register` command would fail the second time the script is executed.

I think that we should have added this not to leak registrations as mentioned [here](https://access.redhat.com/solutions/228883):
```
# subscription-manager remove --all
# subscription-manager unregister
# subscription-manager clean
```

but this would unregister and register the system all the times the script is run.  I think the solution proposed here is cleaner.

moadqassem: I see, actually this is what I was advised to do when the subscription manager didn't work.

pkg/userdata/rhel/provider.go
```diff
@@ -315,6 +311,11 @@ write_files:
     [Service]
     EnvironmentFile=-/etc/environment
 
+rh_subscription:
```
irozzo-1A, line 314: Would it be worthy to make this section optional?

irozzo-1A: After some more thinking I prefer to leave this aside for the moment, as we do not provide other means of registering.

pkg/userdata/rhel/provider.go
irozzo-1A: ok, anyway the [rh_subscription](https://github.com/canonical/cloud-init/blob/3c551f6ebc12f7729a2755c89b19b9000e27cc88/cloudinit/config/cc_rh_subscription.py) module is actively maintained also by RH folks, I think it should be ok.

moadqassem:
@irozzo-1A After testing the new change regarding the rh-subscription module, it appears that, if the cloud init script fails at some point the subscription won't be cleaned. So I am not sure about the value of this change?

irozzo-1A:
> @irozzo-1A After testing the new change regarding the rh-subscription module, it appears that, if the cloud init script fails at some point the subscription won't be cleaned. So I am not sure about the value of this change?

@moadqassem  Let me explain the issue this PR solves to avoid misunderstandings.

The registration of the machine is currently performed in the `setup` script, that gets re-run in case of failure. The problem is that when an error occurs after the node registration, the script is re-run (potentially lot of times as we observed) ending up in multiple registrations for the same machine. This basically means the script is not idempotent.

In order to fix this I found two main options:
- Fix the `setup` script to be idempotent either by re-subscribing cleanly et each run or by checking if a registration is already present.
- Move the registration out from the `setup` script, and the `rh-subscription` module is intended for this purpose. It also verifies if the node is already registered and in this case it does not register again.

IMO the second option is simpler and cleaner.

Note that currently we are not unregistering the systems in any case with the `setup` script, so nothing changes in that respect. Anyway this is not the central point of this PR as mentioned above.

moadqassem:
@irozzo-1A 
> It also verifies if the node is already registered and in this case it does not register again.

But actually this is not happening. I have tested it and I saw that whether the node is registered or not the subscription is recreated.

pkg/userdata/rhel/provider.go
irozzo-1A: > But actually this is not happening. I have tested it and I saw that whether the node is registered or not the subscription is recreated.

Well, there is a test based on `subsctiption-manager identity` command:
https://github.com/canonical/cloud-init/blob/3c551f6ebc12f7729a2755c89b19b9000e27cc88/cloudinit/config/cc_rh_subscription.py#L56
https://github.com/canonical/cloud-init/blob/master/tests/unittests/test_rh_subscription.py#L43

Maybe what you meant is that it does not perform a remote check with Red Hat Subscription Management API, but this does not seem to be necessary for the purpose of this PR.

irozzo-1A:
> But actually this is not happening. I have tested it and I saw that whether the node is registered or not the subscription is recreated.

The test that is performed is [here](
https://github.com/canonical/cloud-init/blob/3c551f6ebc12f7729a2755c89b19b9000e27cc88/cloudinit/config/cc_rh_subscription.py#L56), and relies on this command:
```
subsctiption-manager identity
```

It does not make use of the subscription management API if this is what you mean, it just checks locally. Note that this is not required to solve the issue reported here.

PR #782, opened 2020-06-23
Assign external IP and DNS name for GCP machines

### LINKED ISSUE #781: NodeCSRApprover doesn't approve requests on GCP
The NodeCSRApprover controller is not approving CSRs for GCP worker nodes due to lack of private IP address and DNS name in the `.status.Addresses` field in the Machine object.

The CSR includes both private IP address and DNS name and as those are not in the Machine object, the validation fails, therefore the CSR is not signed.

As a workaround, until the issue is not fixed, the operator can manually approve the CSR, by following the instructions from [the CSR docs](https://kubernetes.io/docs/tasks/tls/managing-tls-in-a-cluster/#approving-certificate-signing-requests).

As a fix, the [following function](https://github.com/kubermatic/machine-controller/blob/e18497b5a6096dcaf31d909d3ea22ed5e95a1db9/pkg/cloudprovider/provider/gce/instance.go#L60-L67) should be extended to include the other type of addresses.

xmudrii:

Assign external IP and DNS name for GCP machines.

Fixes #781

**Optional Release Note**:
```release-note
Add external IPs and DNS name to .status.Addresses for the GCP machines. This fixes the issue with NodeCSRApprover controller refusing to approve CSRs for GCP worker nodes.
```

### DIFF
--- a/pkg/cloudprovider/provider/gce/instance.go
+++ b/pkg/cloudprovider/provider/gce/instance.go
@@ -62,7 +62,11 @@ func (gi *googleInstance) Addresses() map[string]v1.NodeAddressType {
 	addrs := map[string]v1.NodeAddressType{}
 	for _, ifc := range gi.ci.NetworkInterfaces {
 		addrs[ifc.NetworkIP] = v1.NodeInternalIP
+		for _, ac := range ifc.AccessConfigs {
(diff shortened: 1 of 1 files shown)


xmudrii:
/hold
To prevent automatic retests, since RHEL tests are mostly like broken.

Issue #785, opened 2020-06-24
Creating Workers with custom disk size fails on Open Telekom Cloud


irozzo-1A:
See https://github.com/kubermatic/kubermatic/issues/4516

PR #786, opened 2020-06-24
Fix creation of servers with custom disk size on OpenStack

irozzo-1A:
This PR:

- Fix an bug related to machine creation in OpenStack creation
- Add log support for requests addressed to OpenStack API to ease investigation

Fixes # #785

**Optional Release Note**:

```release-note
NONE
```


pkg/cloudprovider/provider/openstack/provider_test.go
```diff
@@ -0,0 +1,379 @@
...
+
+type openstackProviderSpecConf struct {
+	IdentityEndpointURL string
+	RootDiskSizeGB      *int32
+}
+
+func (o openstackProviderSpecConf) rawProviderSpec(t *testing.T) []byte {
+	var out bytes.Buffer
+	tmpl, err := template.New("test").Parse(`{
+					"cloudProvider": "openstack",
```
kron4eg, line 112: This indentation is out of control!

pkg/cloudprovider/provider/openstack/http.go
```diff
@@ -0,0 +1,109 @@
...
+	"k8s.io/klog"
+)
+
+// TODO(irozzo): Make client timeout customizable
+const defaultClientTimeout = 15 * time.Second
+
+// newHTTPClient return a custom HTTP client that allows for logging
+// requests and responses with openstack API information before and after
+// the HTTP request.
+func newHTTPClient() http.Client {
```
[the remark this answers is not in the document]
kron4eg: Only if it will be you who will implement that next usecase. Other person will miss those bits here I guarantee

pkg/cloudprovider/provider/openstack/provider_test.go
irozzo-1A: :-) indeed it's slightly too deep

chrkl:
Thanks for the PR. Looks good for me so far. Since I'm not involved in the machine-controller development at all, I would prefer someone else to give a LGTM (also think I lack permissions).

irozzo-1A:
> Thanks for the PR. Looks good for me so far. Since I'm not involved in the machine-controller development at all, I would prefer someone else to give a LGTM (also think I lack permissions).

Yep no pb with that, it was just FYI. Thx for taking a look ;-)

Issue #788, opened 2020-06-25
Add support for marketplace image plans on Azure


chrkl:
Using a marketplace image on Azure requires to add a plan to the VM resource, e.g. for rhel-byos in Terraform:
```
plan {
  name      = "rhel-lvm81"
  publisher = "redhat"
  product   = "rhel-byos"
}
```

When using an image without the plan, the following error occurs:
`failed to create machine at cloudprovider, due to waiting for operation returned: Code="VMMarketplaceInvalidInput" Message="Creating a virtual machine from Marketplace image requires Plan information in the request. VM: '/subscriptions/.../resourceGroups/k1-tf-mk8s/providers/Microsoft.Compute/virtualMachines/k1-tf-worker-1-9b8b49c89-4pcmz'.`

Issue #780, opened 2020-06-17
Add support for Redhat Satellite server


chrkl:
RHEL provides the option to use a so called Satellite server for subscription management instead of the Redhat customer portal. This requires slightly different steps during the creation and deletion of machines. 

The required steps for the registration are ([Redhat docs](https://access.redhat.com/documentation/en-us/red_hat_satellite/6.6/html/managing_hosts/registering_hosts#registering-a-host-to-satellite)):
- Installing the katello-ca-consumer rpm from the Satellite server
- Calling `subscription-manager register` with `--activationkey` and `--org` arguments (user and password is not required with activation key
  - The `subscription-manager attach` step that we currently do is not required with an activation key
- Adding Satellite tools repo using subscription-manager and installing katello-agent (not sure yet if this is a strict requirement)

The host deletion can be done against the [Satellite API](https://access.redhat.com/documentation/en-us/red_hat_satellite/6.6/html-single/api_guide/index#sect-Working_with_Hosts):
- E.g., `curl -X DELETE --insecure --user user:password https://<satellite-url>/api/v2/hosts/<hostname>`
- The hostname is during registration set to the FQDN

We can do some of these steps during our image build. The minimal requirements are:
- [ ] Support `subscription-manager register` with activation key and organization
- [ ] Delete hosts (including subscription) from Satellite server

I can help with testing this on a Satellite server

PR #789, opened 2020-06-27
Add rhel satellite subscription

moadqassem:
Support Redhat Satellite Server Subscription in the machine controller while using rhel os, in addition to support Azure marketplace image plans. 

Fixes kubermatic/machine-controller#780
Fixes kubermatic/machine-controller#788

At the moment there is no available redhat satellite server support in our testing infrastructure thus the e2e test is disabled. The code has been tested in a different staging env. For more details about the testing please refer to @chrkl. 

**Optional Release Note**:
```release-note
Adding the support for Redhat satellite server hosts registering/subscription on machine creation and unregistering/unsubscribing on machine cleanup. 
```


moadqassem:
cc @chrkl

pkg/cloudprovider/provider/azure/types/types.go
```diff
@@ -38,11 +36,18 @@ type RawConfig struct {
...
 
 	ImageID        providerconfigtypes.ConfigVarString `json:"imageID"`
 	OSDiskSize     int32                               `json:"osDiskSize"`
 	DataDiskSize   int32                               `json:"dataDiskSize"`
 	AssignPublicIP providerconfigtypes.ConfigVarBool   `json:"assignPublicIP"`
 	Tags           map[string]string                   `json:"tags,omitempty"`
 }
+
+// imagePlan contains azure OS Plan fields for the marketplace images
+type imagePlan struct {
```
kron4eg, line 49: I suppose this structure should be public

moadqassem: why ?

kron4eg: how otherwise would you programmatically setup ImagePlan field?

pkg/userdata/rhel/rhel.go
```diff
@@ -28,6 +28,10 @@ type Config struct {
 	RHELSubscriptionManagerUser     string `json:"rhelSubscriptionManagerUser,omitempty"`
 	RHELSubscriptionManagerPassword string `json:"rhelSubscriptionManagerPassword,omitempty"`
 	RHSMOfflineToken                string `json:"rhsmOfflineToken,omitempty"`
+	UseRHSatelliteServer            bool   `json:"useRHSatelliteServer"`
+	RHSatelliteServer               string `json:"rhSatelliteServer"`
+	OrganizationName                string `json:"organizationName"`
+	ActivationKey                   string `json:"activationKey"`
```
kron4eg, line 34: I see nice patter here to prefix RHEL specific fields with `RHEL`, let's continue to do so (including in JSON)

pkg/rhsm/satellite_subscription_manager.go
```diff
@@ -0,0 +1,107 @@
...
+// NewSatelliteSubscriptionManager creates a new Redhat satellite manager.
+func NewSatelliteSubscriptionManager() SatelliteSubscriptionManager {
+	transport := &http.Transport{
+		TLSClientConfig: &tls.Config{
+			InsecureSkipVerify: true,
+		},
+	}
+
+	client := http.DefaultClient
+	client.Transport = transport
```
kron4eg, line 50: Please ether create new http.Client or transport with proper timeouts

pkg/userdata/rhel/rhel.go
moadqassem: yeah I agree.

pkg/rhsm/satellite_subscription_manager.go
```diff
@@ -0,0 +1,107 @@
...
+		klog.Infof("subscription for machine %s deleted successfully", machineName)
+		return nil
+	}
+
+	return errors.New("failed to delete system profile after max retires number has been reached")
+}
+
+func (s *DefaultSatelliteSubscriptionManager) executeDeleteRequest(machineName, username, password, serverURL string) error {
+	var requestURL url.URL
+	requestURL.Scheme = "https"
```
kron4eg, line 84: Please make Scheme a field of DefaultSatelliteSubscriptionManager struct, then it will be possible to avoid TLS complications in unit-tests (that I see below).

pkg/rhsm/satellite_subscription_manager.go
```diff
@@ -0,0 +1,107 @@
...
+	requestURL.Host = serverURL
+	requestURL.Path = path.Join("api", "hosts", machineName)
```
kron4eg, line 86: in original issue (#780) it said that URL should look like `https://<satellite-url>/api/v2/hosts/<hostname>`, but here v2 is missed, is this intended?

moadqassem: We used without adding the `v2` and it worked. however, we would like to stick to what the api in the ticket has been documented. 

cc @chrkl

PR #793, opened 2020-07-15
Make it work for CentOS7/CentOS8/RHEL7/RHEL8

kron4eg:
This PR unblocks the ability to user CentOS8 / RHEL7.

* now it's possible to upgrade docker version
* CentOS8 / RHEL7 can be provisioned

While by default without externally provided image machine-controller will search for CentOS7 images (fixing this is out of scope of this PR), it's now possible to provide CentOS8 (or RHEL7) image in MachineDeployment.

**Optional Release Note**:
```release-note
support CentOS7/CentOS8/RHEL7/RHEL8
```

### DIFF
--- a/pkg/userdata/centos/provider.go
+++ b/pkg/userdata/centos/provider.go
@@ -149,7 +149,7 @@ write_files:
   content: |
 {{ kernelSettings | indent 4 }}
 
-- path: /etc/sysconfig/selinux
+- path: /etc/selinux/config
   content: |
     # This file controls the state of SELinux on the system.
     # SELINUX= can take one of these three values:
@@ -176,9 +176,7 @@ write_files:
     sysctl --system
 
 {{- /* Make sure we always disable swap - Otherwise the kubelet won't start */}}
-    cp /etc/fstab /etc/fstab.orig
-    cat /etc/fstab.orig | awk '$3 ~ /^swap$/ && $1 !~ /^#/ {$0="# commented out by cloudinit\n#"$0} 1' > /etc/fstab.noswap
-    mv /etc/fstab.noswap /etc/fstab
+    sed -i.orig '/.*swap.*/d' /etc/fstab
     swapoff -a
     {{ if ne .CloudProviderName "aws" }}
 {{- /*  The normal way of setting it via cloud-init is broken, see */}}
@@ -188,6 +186,9 @@ write_files:
 
     yum install -y yum-utils
     yum-config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
+{{- /*	Due to DNF modules we have to do this on docker-ce repo
+		More info at: https://bugzilla.redhat.com/show_bug.cgi?id=1756473 */}}
+    yum-config-manager --save --setopt=docker-ce-stable.module_hotfixes=true
 
 {{- /* We need to explicitly specify docker-ce and docker-ce-cli to the same version.
 	See: https://github.com/docker/cli/issues/2533 */}}
@@ -208,7 +209,7 @@ write_files:
       open-vm-tools \
       {{- end }}
       ipvsadm
-    yum versionlock docker-ce-*
+    yum versionlock add docker-ce-*
 
 {{ safeDownloadBinariesScript .KubeletVersion | indent 4 }}
     # set kubelet nodeip environment variable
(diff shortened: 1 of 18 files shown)


moadqassem:
@kron4eg did you update the e2e tests?

pkg/userdata/rhel/provider.go
```diff
@@ -176,20 +176,20 @@ write_files:
...
     {{ if ne .CloudProviderName "aws" }}
 {{- /*  The normal way of setting it via cloud-init is broken, see */}}
 {{- /*  https://bugs.launchpad.net/cloud-init/+bug/1662542 */}}
     hostnamectl set-hostname {{ .MachineSpec.Name }}
     {{ end }}
 
-    dnf config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
+    yum install -y yum-utils
+    yum-config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
+    yum-config-manager --save --setopt=docker-ce-stable.module_hotfixes=true
```
irozzo-1A, line 189: I think that a comment on this line explaining why this is needed would be useful.
BTW I think this is required for RHEL8 only.

pkg/userdata/centos/provider.go
```diff
@@ -188,6 +186,7 @@ write_files:
 
     yum install -y yum-utils
     yum-config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
+    yum-config-manager --save --setopt=docker-ce-stable.module_hotfixes=true
```
irozzo-1A, line 189: Same here, I think a comment would be useful to avoid forgetting about why this is needed.

pkg/userdata/rhel/provider.go
kron4eg: It's required on CentOS8 too. And on CentOS7 it's a noop.

moadqassem: The problem is, we can test RHEL whatsoever, due to lack of subscriptions. Our subscription has expired and it hasn't been updated for RHEL

irozzo-1A:
> @kron4eg did you update the e2e tests? I would like to see the tests run for centos 8 where it is supported :). about Kubevirt I can update our registry, but there should be some work done on the cloud providers which we have our own images(vSphere)

I agree, I think we also need tests for RHEL7, and updating this https://github.com/kubermatic/machine-controller/blob/master/docs/operating-system.md#support-matrix

kron4eg:
@moadqassem not exactly :D I mean, all the tests for centos7 are still valid. It's that __now__ it's possible to add new centos8 scenarios. OK, I'll add some.

pkg/userdata/rhel/provider.go
kron4eg: This command was tested manually on live system multiple times

irozzo-1A: my point was simply that we should have a comment on this line to explain the reason why we are doing this as it is not self evident

pkg/userdata/centos/provider.go
kron4eg: 👍

pkg/userdata/rhel/provider.go
kron4eg: indeed, added the comment 👍

kron4eg:
@moadqassem added "special case test" pull-machine-controller-e2e-aws-centos8 modeled after sles

moadqassem:
> @moadqassem added "special case test" pull-machine-controller-e2e-aws-centos8 modeled after sles

We also need one for rhel 7 in azure. 

Soon we have to think of away how to handle these os versions during testing. At the moment we have the `herlper.go` file which acts as a customiser for those tests, that file has grown in a very unproductive way and the semantics of how we run things is very confusing. In the close future we need to find a better away to handle the e2e tests and cloud-init injection in general. I have an idea in my head how we can achieve this  but we should discuss all to see if it makes sense.

kron4eg:
> We also need one for rhel 7 in azure since we have a customer who is using RHEL on Azure.

From customer point of view nothing will be changed, RHEL8 continues to function like it was.
RHEL7 can wait until we officially have a way for "OS versioning" (centos 7/8, rhel 7/8, ubuntu 18/20), <<- this is a major problem and definitely out of this PR scope :)

irozzo-1A:
LGTM as well, just https://github.com/kubermatic/machine-controller/blob/master/docs/operating-system.md#supported-os-versions has not been updated to include CentOS7, we can do it in a following PR though

xmudrii:
I assume this is a flake.
/retest

PR #798, opened 2020-07-17
Reduce disk storage consumption in Openstack e2e tests

irozzo-1A:
Lastly we hit some quota issues in the openstack tenant we use for our e2e tests. In order to reduce disk consumption this PR is:
* Moving the `rootDiskSizeGB` to the openstack upgrade test that is run with ubuntu OS only
* Reducing `rootDIskSizeGB` to 10G that should be enough

Fixes #

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/provisioning/testdata/machinedeployment-openstack.yaml
+++ b/test/e2e/provisioning/testdata/machinedeployment-openstack.yaml
@@ -34,7 +34,6 @@ spec:
             domainName: "<< DOMAIN_NAME >>"
             region: "<< REGION >>"
             network: "<< NETWORK_NAME >>"
-            rootDiskSizeGB: 20
           operatingSystem: "<< OS_NAME >>"
           operatingSystemSpec:
             distUpgradeOnBoot: false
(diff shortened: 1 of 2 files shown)


test/e2e/provisioning/testdata/machinedeployment-openstack.yaml
```diff
@@ -34,7 +34,6 @@ spec:
             domainName: "<< DOMAIN_NAME >>"
             region: "<< REGION >>"
             network: "<< NETWORK_NAME >>"
-            rootDiskSizeGB: 20
```
kron4eg, line 37: isn't this supposed to be changed to 10 instead of being completely removed?

irozzo-1A: We have two manifests we use for openstack. I wanted to have 1 in which we specify `rootDiskSizeGB` and one where we do not in order to cover both options. I moved it to the other one because we run it just with ubuntu, while this one is used many times. Not sure actually if without specifying anything we consume more or less disk, I need to double check this.

PR #801, opened 2020-07-31
New -node-kubelet-repository flag

### LINKED ISSUE #800: Move away from Hyperkube image providing Kubelet for Flatcar machines
For the same reasons as https://github.com/kubermatic/kubermatic/issues/5717, we have to get rid of hyperkube image, that is currently used to provide the Kubelet for Flatcar machines. An additional complication is that there is no official image for Kubelet at the moment.

The current strategy is to use an image maintained by some third parties e.g. [Poseidon Kubelet image](https://github.com/poseidon/kubelet#packages).

Some more context can be found here:
https://forum.kubermatic.com/t/migrate-away-from-hyperkube-in-machine-controller/70

kron4eg:
Attempt to replace hyperkube image

Fixes #800

Heavily inspired by the https://github.com/kinvolk/lokomotive/blob/master/assets/lokomotive-kubernetes/bootkube/resources/charts/kubelet/templates/kubelet-ds.yaml

**Optional Release Note**:

```release-note
on flatcar linux hyperkube image is partially replaced by the quay.io/poseidon/kubelet
```

### DIFF
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -82,6 +82,7 @@ var (
 	nodeRegistryMirrors    string
 	nodePauseImage         string
 	nodeHyperkubeImage     string
+	nodeKubeletRepository  string
 )
 
 const (
@@ -165,7 +166,8 @@ func main() {
 	flag.StringVar(&nodeInsecureRegistries, "node-insecure-registries", "", "Comma separated list of registries which should be configured as insecure on the container runtime")
 	flag.StringVar(&nodeRegistryMirrors, "node-registry-mirrors", "", "Comma separated list of Docker image mirrors")
 	flag.StringVar(&nodePauseImage, "node-pause-image", "", "Image for the pause container including tag. If not set, the kubelet default will be used: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/")
-	flag.StringVar(&nodeHyperkubeImage, "node-hyperkube-image", "k8s.gcr.io/hyperkube-amd64", "Image for the hyperkube container excluding tag.")
+	flag.StringVar(&nodeHyperkubeImage, "node-hyperkube-image", "k8s.gcr.io/hyperkube-amd64", "Image for the hyperkube container excluding tag. Only has effect on CoreOS Container Linux and Flatcar Linux, and for kubernetes < 1.18.")
+	flag.StringVar(&nodeKubeletRepository, "node-kubelet-repository", "quay.io/poseidon/kubelet", "Repository for the kubelet container. Only has effect on Flatcar Linux, and for kubernetes >= 1.18.")
 	flag.BoolVar(&nodeCSRApprover, "node-csr-approver", false, "Enable NodeCSRApprover controller to automatically approve node serving certificate requests.")
 
 	flag.Parse()
@@ -198,13 +200,23 @@ func main() {
 	if err := clusterv1alpha1.AddToScheme(scheme.Scheme); err != nil {
 		klog.Fatalf("failed to add clusterv1alpha1 api to scheme: %v", err)
 	}
+
 	// Check if the hyperkube image has a tag set
 	hyperkubeImageRef, err := reference.Parse(nodeHyperkubeImage)
 	if err != nil {
-		klog.Fatalf("failed to parse --node-hyperkube-image %s: %v", nodeHyperkubeImage, err)
+		klog.Fatalf("failed to parse -node-hyperkube-image %s: %v", nodeHyperkubeImage, err)
 	}
 	if _, ok := hyperkubeImageRef.(reference.NamedTagged); ok {
-		klog.Fatalf("--node-hyperkube-image must not contain a tag. The tag will be dynamically set for each Machine.")
+		klog.Fatalf("-node-hyperkube-image must not contain a tag. The tag will be dynamically set for each Machine.")
+	}
+
+	// Check if the kubelet image has a tag set
+	kubeletRepoRef, err := reference.Parse(nodeKubeletRepository)
+	if err != nil {
+		klog.Fatalf("failed to parse -node-hyperkube-image %s: %v", nodeHyperkubeImage, err)
+	}
+	if _, ok := kubeletRepoRef.(reference.NamedTagged); ok {
+		klog.Fatalf("-node-kubelet-image must not contain a tag. The tag will be dynamically set for each Machine.")
 	}
 
 	cfg, err := clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
@@ -246,11 +258,12 @@ func main() {
 		skipEvictionAfter:     skipEvictionAfter,
 		nodeCSRApprover:       nodeCSRApprover,
 		node: machinecontroller.NodeSettings{
-			ClusterDNSIPs:  clusterDNSIPs,
-			HTTPProxy:      nodeHTTPProxy,
-			NoProxy:        nodeNoProxy,
-			HyperkubeImage: nodeHyperkubeImage,
-			PauseImage:     nodePauseImage,
+			ClusterDNSIPs:     clusterDNSIPs,
+			HTTPProxy:         nodeHTTPProxy,
+			NoProxy:           nodeNoProxy,
+			HyperkubeImage:    nodeHyperkubeImage,
+			KubeletRepository: nodeKubeletRepository,
+			PauseImage:        nodePauseImage,
 		},
 	}
 	if parsedJoinClusterTimeout != nil {
--- a/docs/network-restrictions.md
+++ b/docs/network-restrictions.md
@@ -27,16 +27,35 @@ If that image won't be accessible from the node, a custom image can be specified
 -node-pause-image="192.168.1.1:5000/kubernetes/pause:3.1"
 ```
 
-For ContainerLinux nodes the [hyperkube](https://github.com/kubernetes/kubernetes/tree/master/cluster/images/hyperkube) image must be accessible as well.
-This is due to the usage of the [kubelet-wrapper](https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/kubelet-wrapper.md).
 
-By default the image `k8s.gcr.io/hyperkube-amd64` will be used.
-If that image won't be accessible from the node, a custom image can be specified on the machine-controller:
+## Kubelet images
+
+### CoreOS ContainerLinux
+For ContainerLinux nodes, the [hyperkube][1] image must be accessible as well. This is due to the usage of the
+[kubelet-wrapper][2].
+
+By default the image `k8s.gcr.io/hyperkube-amd64` will be used. If that image won't be accessible from the node, a
+custom image can be specified on the machine-controller:
 ```bash
 # Do not set a tag. The tag depends on the used Kubernetes version of a machine.
 # Example:
 # A Node using v1.14.2 would use 192.168.1.1:5000/kubernetes/hyperkube-amd64:v1.14.2
 -node-hyperkube-image="192.168.1.1:5000/kubernetes/hyperkube-amd64"
+```
+
+### Flatcar Linux
+For Flatcar Linux nodes, the [hyperkube][1] or [kubelet][3] image must be accessible as well. This is due to the fact
+that kubelet is running as a docker container. For kubelet version `< 1.18` hyperkube will be used, otherwise `kubelet`
+image.
+
+By default the image `quay.io/poseidon/kubelet` will be used. If that image won't be accessible from the node, a custom
+image can be specified on the machine-controller:
+```bash
+# Do not set a tag. The tag depends on the used Kubernetes version of a machine.
+# Example:
+# A Node using v1.14.2 would use 192.168.1.1:5000/kubernetes/hyperkube-amd64:v1.14.2
+-node-hyperkube-image="192.168.1.1:5000/kubernetes/hyperkube-amd64"
+-node-kubelet-image="192.168.1.1:5000/my-custom/kubelet-amd64"
 ```
 
 # Insecure registries
@@ -45,3 +64,7 @@ If nodes require access to insecure registries, all registries must be specified
 ```bash
 --node-insecure-registries="192.168.1.1:5000,10.0.0.1:5000"
 ```
+
+[1]: https://console.cloud.google.com/gcr/images/google-containers/GLOBAL/hyperkube
+[2]: https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/kubelet-wrapper.md
+[3]: https://quay.io/poseidon/kubelet
--- a/pkg/controller/machine/machine.go
+++ b/pkg/controller/machine/machine.go
@@ -124,8 +124,10 @@ type NodeSettings struct {
 	RegistryMirrors []string
 	// Translates to --pod-infra-container-image on the kubelet. If not set, the kubelet will default it.
 	PauseImage string
-	// The hyperkube image to use. Currently only Container Linux uses it.
+	// The hyperkube image to use. Currently only Container Linux and Flatcar Linux uses it.
 	HyperkubeImage string
+	// The kubelet repository to use. Currently only Flatcar Linux uses it.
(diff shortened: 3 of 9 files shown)


pkg/userdata/flatcar/provider.go
```diff
@@ -227,38 +234,42 @@ systemd:
...
+          -v /opt/cni/bin:/opt/cni/bin:ro \
+          -v /run:/run \
+          -v /sys:/sys \
+          -v /usr/sbin/iscsiadm:/usr/sbin/iscsiadm \
+          -v /var/lib/calico:/var/lib/calico:ro \
+          -v /var/lib/cni:/var/lib/cni \
+          -v /var/lib/docker:/var/lib/docker \
+          -v /var/lib/kubelet:/var/lib/kubelet:rshared \
+          -v /var/log/pods:/var/log/pods \
+          {{ .KubeletImage }}:v{{ .KubeletVersion }} \
```
xmudrii, line 270: Similar to the comment regarding when should we start using this image, should we continue using rkt to run Hyperkube, or are we sure that it'll work properly using this approach? Have you tested this with older versions?

pkg/userdata/flatcar/provider.go
```diff
@@ -73,37 +78,39 @@ func (p Provider) UserData(req plugin.UserDataRequest) (string, error) {
...
-		}
-	}
-
 	if flatcarConfig.DisableAutoUpdate {
 		flatcarConfig.DisableLocksmithD = true
 		flatcarConfig.DisableUpdateEngine = true
 	}
 
+	kubeletImage := poseidonImage
+	lessThen118, err := semver.NewConstraint(lessThen118Check)
```
irozzo-1A, line 87: no need to compile this each time and doing an error check as it is a constant, I would rather use a global variable initialised in an init block.

pkg/userdata/flatcar/provider.go
```diff
@@ -33,6 +32,12 @@ import (
 	userdatahelper "github.com/kubermatic/machine-controller/pkg/userdata/helper"
 )
 
+const (
+	poseidonImage    = "quay.io/poseidon/kubelet"
```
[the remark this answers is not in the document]
irozzo-1A: I would rather call it `kubeletImage` in such case, giving an option to override it could turn useful indeed

pkg/userdata/flatcar/provider.go
irozzo-1A: Considering that this applies to newly created machines only and that RKT and `kubelet-wrapper` are deprecated I think that this approach makes sense. I presume we will have to backport this to the oldest `machine-controller` version used by the oldest supported `kubermatic`, I agree with @xmudrii that we need to ensure everything works there.

pkg/userdata/flatcar/provider.go
kron4eg: I'm 100% do not agree with anything global or func init(). That's considered as bad coding practices by the community (for a good reason).

pkg/userdata/flatcar/provider.go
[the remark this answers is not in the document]
kron4eg: then we'd need 2 flags, 1 for "< 1.18" kubelets and 1 for "> 1.18".

pkg/userdata/flatcar/provider.go
kron4eg: `rkt` is dead long ago, `kubelet-wrapper` is effectively dead. Fear should drive away from `kubelet-wrapper` since "I fear flatcar car remove long deprecated kubelet-wrapper". Please refer to https://github.com/flatcar-linux/Flatcar/issues/15 for more information why it's a good idea to forget about both rkt and kubelet-wrapper.

pkg/userdata/flatcar/provider.go
```diff
@@ -33,6 +32,12 @@ import (
...
+	hyperkubeImage   = "k8s.gcr.io/hyperkube-amd64"
+	lessThen118Check = "< 1.18"
```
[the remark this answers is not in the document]
kron4eg: I'd use poseidon even for 1.17+ if it would be accessible because hyperkube was officially deprecated in 1.17 https://github.com/kubernetes/kubernetes/pull/85094, but poseidon is only consistently accessible since 1.18+.

kron4eg:
Yeah, I've passed certification mode using sonobuoy

[202008031520_sonobuoy_36b1dab1-375e-431c-b969-338dbad2e2e5.tar.gz](https://github.com/kubermatic/machine-controller/files/5017261/202008031520_sonobuoy_36b1dab1-375e-431c-b969-338dbad2e2e5.tar.gz)

pkg/userdata/flatcar/provider.go
irozzo-1A: @kron4eg Could you please elaborate a bit more about the "good reason" you mentioned referring to this specific case? Just to be clear, I was suggesting something like:

```go
var (
	lessThen118 = *constraintOrDie("< 1.18")
)

func constraintOrDie(c string) *semver.Constraints {
	o, err := semver.NewConstraint(c)
	if err != nil {
		panic(err)
	}
	return o
}
```

Compiling the constraint and checking the error every time would make more sense if it was dynamically provided, but as it is a constant it seems redundant to me.
The drawback I see with this approach is that `semver.Constraints` is mutable, so the theoretical risk of having it modified at runtime holds.

That said I'm not going to fight for this because it does not change much, it's just to understand your point.

pkg/userdata/flatcar/provider.go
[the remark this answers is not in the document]
irozzo-1A: I think we could have just one that applies to both cases. We could default to poseidon for "> 1.18" and hyperkube for "< 1.18", but again I don't think this is fundamental.

pkg/userdata/flatcar/provider.go
kron4eg: In general, go community condemn globals, I don't want to retype here all the reasons, you can find them here:
* https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html
* https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables


There is even a linter to check that your code has no globals https://github.com/leighmcculloch/gochecknoglobals

irozzo-1A: Thx for the pointers, but IMO they don't add much to the discussion. I was more interested in your concerns regarding this specific case not the general one. Best practices can be useful tools, but it's important to separate fact from bias.  

> The drawback I see with this approach is that semver.Constraints is mutable, so the theoretical risk of having it modified at runtime holds.

This was the only theoretical drawback I found that could apply here, but after better looking it's not even valid, because the mutable internals of https://godoc.org/github.com/Masterminds/semver#Constraints are not exposed and no mutator is given.

If I did not convince you it's not a problem, I hope at least that your 100% become a 98% :-)

pkg/userdata/flatcar/provider.go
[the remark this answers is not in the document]
xmudrii: > I think we could have just one that applies to both cases. We could default to poseidon for "> 1.18" and hyperkube for "< 1.18", but again I don't think this is fundamental.

The problem with this approach is that someone can use MC to create both > 1.18 and < 1.18 nodes at the same time, so I'm not sure that one flag would work. I think we can have just one flag for Poseidon if someone wants to use their own image instead of a third-party one, and always use the upstream image for Hyperkube.

On the other hand, I don't think this is very important, so we can skip it unless there's no an request for this.

pkg/userdata/flatcar/provider.go
xmudrii: I definitely agree with both of you. I was mostly thinking about testing does everything work correctly and since conformance tests are passing, I think it's a good sign to proceed with this change. :shipit:

pkg/userdata/flatcar/provider.go
```diff
@@ -73,37 +78,39 @@ func (p Provider) UserData(req plugin.UserDataRequest) (string, error) {
...
+		KubeletVersion   string
+		NodeIPScript     string
 	}{
-		UserDataRequest:        req,
-		ProviderSpec:           pconfig,
-		FlatcarConfig:          flatcarConfig,
-		Kubeconfig:             kubeconfigString,
-		KubernetesCACert:       kubernetesCACert,
-		KubeletVersion:         kubeletVersion.String(),
-		InsecureHyperkubeImage: insecureHyperkubeImage,
```
xmudrii, line 105: This is a [machine-controller flag (`-node-hyperkube-image`)](https://github.com/kubermatic/machine-controller/blob/7e26d77173e9b4ef3b8bdf137e4d1f3a82469be2/cmd/machine-controller/main.go#L168). We should either continue using it (that would fix [this comment](https://github.com/kubermatic/machine-controller/pull/801/files#r464395765)) or we should deprecate it.

pkg/userdata/flatcar/provider.go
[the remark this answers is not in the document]
irozzo-1A: > The problem with this approach is that someone can use MC to create both > 1.18 and < 1.18 nodes at the same time, so I'm not sure that one flag would work. I think we can have just one flag for Poseidon if someone wants to use their own image instead of a third-party one, and always use the upstream image for Hyperkube.

makes sense, anyway this flag could be useful for scenarios with limited internet access, thus it's probably better to have it from the beginning. As you pointed out we have already a flag for `< 1.18` so we could add something like `kubeletRepository` for `>= 1.18`

pkg/userdata/flatcar/provider.go
kron4eg: yeah, that's the way I'm working currently on flags support

pkg/userdata/flatcar/provider.go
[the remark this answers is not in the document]
kron4eg: In addition to the `-node-hyperkube-image`, there is new flag `-node-kubelet-image`, which regulates the second image. Someday we will deprecate and remove `-node-hyperkube-image`, but for now looks like we need them both.

cmd/machine-controller/main.go
```diff
@@ -166,6 +167,7 @@ func main() {
 	flag.StringVar(&nodeRegistryMirrors, "node-registry-mirrors", "", "Comma separated list of Docker image mirrors")
 	flag.StringVar(&nodePauseImage, "node-pause-image", "", "Image for the pause container including tag. If not set, the kubelet default will be used: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/")
 	flag.StringVar(&nodeHyperkubeImage, "node-hyperkube-image", "k8s.gcr.io/hyperkube-amd64", "Image for the hyperkube container excluding tag.")
+	flag.StringVar(&nodeKubeletImage, "node-kubelet-image", "quay.io/poseidon/kubelet", "Image for the kubelet container excluding tag.")
```
xmudrii, line 170: We should probably add to the description that Hyperkube is used only for < 1.18 and Kubelet for >= 1.18.

docs/network-restrictions.md
```diff
@@ -27,16 +27,35 @@ If that image won't be accessible from the node, a custom image can be specified
...
 ```
 
-For ContainerLinux nodes the [hyperkube](https://github.com/kubernetes/kubernetes/tree/master/cluster/images/hyperkube) image must be accessible as well.
-This is due to the usage of the [kubelet-wrapper](https://github.com/coreos/coreos-kubernetes/blob/master/Documentation/kubelet-wrapper.md).
 
-By default the image `k8s.gcr.io/hyperkube-amd64` will be used.
-If that image won't be accessible from the node, a custom image can be specified on the machine-controller:
+## Kubelet images
+
+### CoreOS ContainerLinux
```
xmudrii, line 33: We should mention that CoreOS will not work with v1.19 and that we highly advise migrating to Flatcar.

pkg/controller/machine/machine.go
```diff
@@ -126,6 +126,8 @@ type NodeSettings struct {
 	PauseImage string
 	// The hyperkube image to use. Currently only Container Linux uses it.
```
xmudrii, line 127: Nit: this comment is outdated, both Container Linux and Flatcar use it.

pkg/userdata/flatcar/provider.go
```diff
@@ -164,7 +168,7 @@ systemd:
             Environment=ALL_PROXY={{ .HTTPProxy }}
 {{- end }}
 
-    - name: download-healthcheck-script.service
+    - name: download-script.service
```
xmudrii, line 171: Why we renamed this service? AFAIK, it downloads the health-check script, so mentioning it might make sense.

docs/network-restrictions.md
kron4eg: CoreOS does not work already

pkg/userdata/flatcar/provider.go
kron4eg: Because it's "download all stuff", not just health-check
`ExecStart=/opt/bin/download.sh`

cmd/machine-controller/main.go
kron4eg: There is such comment in the updated doc

xmudrii: I think having it in both places would be useful, but I'll leave it up to you and @irozzo-1A for the final decision.

pkg/userdata/flatcar/provider.go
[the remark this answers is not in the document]
irozzo-1A: Sorry for the cosmetic comment, but wouldn't `kubelet-repository` be a better name? `node-` seems superfluous and `repository` is more precise as the tag shouldn't be specified.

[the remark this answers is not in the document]
xmudrii: I believe that `node-kubelet-repository` would be a better name because it matches other related flags.

kron4eg:
@xmudrii OK, comments addressed

pkg/userdata/flatcar/provider.go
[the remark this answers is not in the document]
irozzo-1A: @kron4eg I think this is the last point, don't you think that `repository` would be more appropriate than `image`? It would avoid you to specify that the tag should not be specified.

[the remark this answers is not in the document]
kron4eg: the `-node-*` family of flags inherit this `node` prefix to signify that those settings directly control what will land at the node.

```
  -node-http-proxy string
        If set, it configures the 'HTTP_PROXY' & 'HTTPS_PROXY' environment variable on the nodes.
  -node-hyperkube-image string
        Image for the hyperkube container excluding tag. Only has effect on CoreOS Container Linux and Flatcar Linux, and for kubernetes < 1.18. (default "k8s.gcr.io/hyperkube-amd64")
  -node-insecure-registries string
        Comma separated list of registries which should be configured as insecure on the container runtime
  -node-kubelet-image string
        Image for the kubelet container excluding tag. Only has effect on Flatcar Linux, and for kubernetes >= 1.18. (default "quay.io/poseidon/kubelet")
  -node-no-proxy string
        If set, it configures the 'NO_PROXY' environment variable on the nodes. (default ".svc,.cluster.local,localhost,127.0.0.1")
  -node-pause-image string
        Image for the pause container including tag. If not set, the kubelet default will be used: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/
  -node-registry-mirrors string
```

so we ether review this "prefix" grouping or leave `-node-kubelet-image` alone.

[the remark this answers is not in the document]
xmudrii: @kron4eg The prefix is okay. I think @irozzo-1A is talking about renaming the flag to `-node-kubelet-repository`, i.e. replace `image` with `repository`.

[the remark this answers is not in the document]
irozzo-1A: yep, I'm not fond of this `node-` prefix but lets keep it for consistency. The proposal was to have `node-kubelet-repository` instead of `node-kubelet-image` as the term `repository` looks more appropriated.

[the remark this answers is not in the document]
kron4eg: Renamed

PR #802, opened 2020-07-31
Support Flatcar in Packet cloud provider

dongsupark:
To allow Flatcar to be used as one of the operating systems, we need to update `getNameForOS()` so it also accepts the input value `flatcar` in `operatingSystem` of the provider config.

### Optional Release Note:

```release-note
support Flatcar in Packet cloud provider
```


dongsupark:
Done. Updated the Packet e2e test.
Thanks for the review.

xmudrii:
/hold
we are aware of the problem and we are working on fixing it

dongsupark:
Thanks!

PR #807, opened 2020-08-07
Fix free floating IP detection logic for OTC

### LINKED ISSUE #803: [Openstack] False positives in check for free FIPs
The current check for free FIPs does seem to have some false positives, at least in OTC. The check is the following:

https://github.com/kubermatic/machine-controller/blob/7e26d77173e9b4ef3b8bdf137e4d1f3a82469be2/pkg/cloudprovider/provider/openstack/helper.go#L337

Some details about this check can be found [here](https://github.com/kubermatic/machine-controller/pull/28)

The issue I encountered is with a FIP with status `Active` and no port_id set, that is considered free but actually associated with an ELB.

The results is the following:
```
	failed to create machine at cloudprovider, due to failed to assign a floating ip to instance 8b0582b8-6c42-4f4a-a04b-9b29485080a7: failed to update FloatingIP <xxx>(x.x.21.144): Expected HTTP response code [200] when accessing [PUT https://vpc.eu-de.otc.t-systems.com/v2.0/floatingips/<xxx>], but got 409 instead {"NeutronError":{"message":"Resource is busy. Please try it again later.","type":" HTTPInternalServerError","detail":""}}
```

irozzo-1A:
This is a tentative fix for #803. I have still doubts about the possible side-effects. It could possibly result in available FIPs not been considered eligible.
See https://github.com/kubermatic/machine-controller/pull/28#discussion_r163773619 for more details.

Fixes #803 

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/helper.go
+++ b/pkg/cloudprovider/provider/openstack/helper.go
@@ -43,7 +43,7 @@ var (
 )
 
 const (
-	openstackFloatingIPErrorStatusName = "ERROR"
+	floatingIPErrorStatus = "ERROR"
 
 	floatingReassignIPCheckPeriod = 3 * time.Second
 )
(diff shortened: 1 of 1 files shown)


irozzo-1A:
@xrstf

xrstf:
I'm preparing an OTC provider, which will have its own copy of this functionality. I would propose that we do not change the Openstack logic and close this PR.

/hold

xrstf:
I have included the change from this PR in https://github.com/kubermatic/machine-controller/pull/807.

irozzo-1A:
@xrstf This new fix should fix the issue without causing regressions according to my observations on the Openstack providers we use in our dev platform.

irozzo-1A:
@xrstf PTAL

xrstf (approved):
/approve

kron4eg:
@irozzo-1A please check if this is still needed and cancel hold in case if yes.

PR #809, opened 2020-08-13
Add Open Telekom Cloud (OTC) Support

xrstf:
This copies the Openstack code and adjusts it to work with the Open Telekom Cloud (OTC).

**Optional Release Note**:
```release-note
Add Support for Open Telekom Cloud
```

### DIFF
--- a/pkg/cloudprovider/provider/otc/provider.go
+++ b/pkg/cloudprovider/provider/otc/provider.go
@@ -1,0 +1,936 @@
+/*
+Copyright 2019 The Machine Controller Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package otc
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strings"
+	"sync"
+	"time"
+
+	gophercloud "github.com/RainbowMango/huaweicloud-sdk-go"
+	"github.com/RainbowMango/huaweicloud-sdk-go/auth/aksk"
+	"github.com/RainbowMango/huaweicloud-sdk-go/openstack"
+	"github.com/RainbowMango/huaweicloud-sdk-go/openstack/compute/v2/extensions/bootfromvolume"
+	osservers "github.com/RainbowMango/huaweicloud-sdk-go/openstack/compute/v2/servers"
+	osfloatingips "github.com/RainbowMango/huaweicloud-sdk-go/openstack/networking/v2/extensions/layer3/floatingips"
+	osnetworks "github.com/RainbowMango/huaweicloud-sdk-go/openstack/networking/v2/networks"
+	"github.com/RainbowMango/huaweicloud-sdk-go/pagination"
+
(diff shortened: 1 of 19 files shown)


pkg/cloudprovider/provider/otc/provider.go
```diff
@@ -0,0 +1,942 @@
...
+	}
+
+	var rawConfig otctypes.RawConfig
+	err = json.Unmarshal(pconfig.CloudProviderSpec.Raw, &rawConfig)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
+	c := Config{}
+	c.AccessKey, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.AccessKey, "OTC_ACCESS_KEY")
```
kron4eg, line 139: Those names, `OTC_*`, where do they come from? I'm asking because for example in [terraform](https://www.terraform.io/docs/providers/opentelekomcloud/index.html#configuration-reference) they use standard openstack names

xrstf: I came up with them, thinking it's sensible to have separate names from Openstack, so that you can have an Openstack and an OTC preset in the same container.

kron4eg: My problem with custom (diverged from the upstream) names, is that later we going to need to ask users of kubeone to copy their terraform OS_* to OTC_*, or even worse, to convert them manually (it's worse because it's implicit and there will be no way to prevent it). So I'd prefer very much the way it's represented in terraform. Not because it's right, but because it's established.

xrstf: I have no strong feelings either way, but if we *ever* run into a conflict, I'll reserve the right to say "I tOlD yOu So!" ;-)

xmudrii: I agree with @kron4eg that we should stick with commonly used variables. If they use `OS_*` variables in their Terraform provider, I think that we should use the same variables in machine-controller. If we want to be sure, we can check what they use in other projects (e.g. CCM or any other project they have), and decide based on that.

> that you can have an Openstack and an OTC preset in the same container.

Do we have some idea when we might want to do that?

xrstf: I could ask the same question: Why did they prefix their access key with "OS_"? (OS_ACCESS_KEY)? Do they really think that there might be other access keys? They obviously did it to be careful and a good citizen and not clobber the program environment.

We want to treat OTC as its own provider and not just an openstack flavor. So it IMHO made and still makes sense to have variables specific to that provider. IMHO it's a terrible choice from Terraform to re-use "OS_" for the OTC provider, but alas, that ship has sailed.

I don't mind the name for the env variables either way, but I find it funny that I suddenly have to defend assining project-specific prefixes to avoid naming conflicts :-D Either OTC is 100% compatible, then we can re-use the Openstack stuff, or we actually treat it as its own thing.

> Do we have some idea when we might want to do that?

Have an e2e test that uses both the openstack and OTC presets in Prow, because you want to test in multiple clouds. Boom. Have the same .direnv in your terraform directory that manages nodes in multiple clusters. Boom. Is it in our multi-cluster multi-cloud world really so hard to imagine that one would for whatever reason have an environment for multiple cloud providers? :-D

xrstf: Changed to OS_.

xrstf: Need to change the preset, so the tests fill fail for now.

xrstf: Done and done.

youssefazrak:
@xrstf is this one still relevant?

shaase-ctrl:
We use another implementation for OTC and therefore have no need to work on this.

PR #812, opened 2020-08-22
Use c1.small.x86 instance type on Packet

xmudrii:

We've used t1.small.x86 instances for running E2E tests because we
don't need large instances. The t1.small.x86 instance type is first
generation, and as the first generation is deprecated, it's not
available anymore. The c1.small.x86 instances are bit more expensive,
but it's the smallest instance type currently available. Besides that,
we now run Packet E2E tests in the Amsterdam region.

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/provisioning/testdata/machinedeployment-packet.yaml
+++ b/test/e2e/provisioning/testdata/machinedeployment-packet.yaml
@@ -26,9 +26,9 @@ spec:
           cloudProviderSpec:
             apiKey: << PACKET_API_KEY >>
             projectID: << PACKET_PROJECT_ID >>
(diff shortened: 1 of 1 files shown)


dongsupark:
FYI, recently Packet's `t1.small.x86` became deprecated, and Packet simply does not allow new projects to access to `t1.small.x86`.
The type might be still accessible in existing projects in the facility ams1, but it could disappear soon.
That's why projects like Lokomotive changed the default type to [`c3.small.x86`](https://github.com/kinvolk/lokomotive/pull/612).

xrstf (approved):
/approve

PR #813, opened 2020-08-23
Delete Nodes by NodeRef and compare node and machine names when getting a Hetzner node

xmudrii:

This PR attempts to fix the bug which occurs when a Machine is rolled-out and the new Machine has the same IP addresses as the old one. It could happen that the machine-controller sets the NodeOwner label on the old Node object to the UID of the new Machine. That can break the logic for removing the old Node object, as it depends on the NodeOwner label having the UID of the Machine that's being deleted.

There are two fixes:

* The getNode function which is used when adding the NodeRef now confirms that the instance and node names match for **Hetzner Machines/instances**. Previously, we only checked IP addresses, so if we a new Machine has the same addresses as the old one, it can adopt a node that's to-be-deleted
  * This assumes the name of the Node object will always be the same as the Machine name.
    * The Machine name is **not always** same as the Node name. For example, on AWS, the Machine name is based on the MachineDeployment name, while the Node name is the internal DNS name.
  * On providers with CCM (internal or external), that's not a problem because CCM is responsible for setting the ProviderID on the Node object. If the ProviderID is present, it will be used by the `getNode` function instead of IP addresses
* The deleteNodeForMachine function now uses NodeRef if it's present. We use NodeRef for all other things when deleting a machine, such as evicting the node and deleting the cloud provider instance
  * The old logic (based on the NodeOwner label) is kept in the place but it's used only as a fall-back if the NodeRef is not present

**Optional Release Note**:
```release-note
* Delete the Node object by NodeRef if it's present
* Compare node and instances names when applying the NodeOwner label for Hetzner machines/instances
```

### DIFF
--- a/pkg/controller/machine/machine.go
+++ b/pkg/controller/machine/machine.go
@@ -608,19 +608,39 @@ func (r *Reconciler) deleteCloudProviderInstance(prov cloudprovidertypes.Provide
 }
 
 func (r *Reconciler) deleteNodeForMachine(machine *clusterv1alpha1.Machine) error {
-	selector, err := labels.Parse(NodeOwnerLabelName + "=" + string(machine.UID))
-	if err != nil {
-		return fmt.Errorf("failed to parse label selector: %v", err)
-	}
-	listOpts := &ctrlruntimeclient.ListOptions{LabelSelector: selector}
-	nodes := &corev1.NodeList{}
-	if err := r.client.List(r.ctx, nodes, listOpts); err != nil {
-		return fmt.Errorf("failed to list nodes: %v", err)
-	}
+	// If there's NodeRef on the Machine object, remove the Node by using the
+	// value of the NodeRef. If there's no NodeRef, try to find the Node by
+	// listing nodes using the NodeOwner label selector.
+	if machine.Status.NodeRef != nil {
+		objKey := ctrlruntimeclient.ObjectKey{Name: machine.Status.NodeRef.Name}
+		node := &corev1.Node{}
+		if err := r.client.Get(r.ctx, objKey, node); err != nil {
+			return fmt.Errorf("failed to get node %s: %v", machine.Status.NodeRef.Name, err)
+		}
 
-	for _, node := range nodes.Items {
-		if err := r.client.Delete(r.ctx, &node); err != nil {
+		if err := r.client.Delete(r.ctx, node); err != nil {
 			return err
+		}
+	} else {
+		selector, err := labels.Parse(NodeOwnerLabelName + "=" + string(machine.UID))
+		if err != nil {
+			return fmt.Errorf("failed to parse label selector: %v", err)
+		}
+		listOpts := &ctrlruntimeclient.ListOptions{LabelSelector: selector}
+		nodes := &corev1.NodeList{}
+		if err := r.client.List(r.ctx, nodes, listOpts); err != nil {
+			return fmt.Errorf("failed to list nodes: %v", err)
+		}
+		if len(nodes.Items) == 0 {
+			// We just want log that we didn't found the node. We don't want to
+			// return here, as we want to remove finalizers at the end.
+			klog.V(3).Infof("No node found for the machine %s", machine.Spec.Name)
+		}
+
+		for _, node := range nodes.Items {
+			if err := r.client.Delete(r.ctx, &node); err != nil {
+				return err
+			}
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/machine/machine.go
```diff
@@ -608,19 +608,42 @@ func (r *Reconciler) deleteCloudProviderInstance(prov cloudprovidertypes.Provide
...
-	}
-	listOpts := &ctrlruntimeclient.ListOptions{LabelSelector: selector}
-	nodes := &corev1.NodeList{}
-	if err := r.client.List(r.ctx, nodes, listOpts); err != nil {
-		return fmt.Errorf("failed to list nodes: %v", err)
-	}
+	// If there's NodeRef on the Machine object, remove the Node by using the
+	// value of the NodeRef. If there's no NodeRef, try to find the Node by
+	// listing nodes using the NodeOwner label selector.
+	if machine.Status.NodeRef != nil {
```
irozzo-1A, line 614: The logic is not very complicated, but a unit test would be welcome :-)

pkg/controller/machine/machine.go
```diff
@@ -890,6 +913,15 @@ func (r *Reconciler) getNode(instance instance.Instance, provider providerconfig
 		}
 		for _, nodeAddress := range node.Status.Addresses {
 			for instanceAddress := range instance.Addresses() {
+				// Hetzner reuses IP addresses, so if we rollout a Machine, it can happen that a new
```
irozzo-1A, line 916: Every cloud provider reuses public IP addresses when released :-) This is a mitigation for when the providerID is not specified in the node and we fallback to the logic based on IP matching. I think that the comment could be clearer.

pkg/controller/machine/machine.go
xmudrii: Added a bunch of tests.

pkg/controller/machine/machine.go
xmudrii: Fixed.

pkg/controller/machine/machine.go
```diff
@@ -888,8 +911,31 @@ func (r *Reconciler) getNode(instance instance.Instance, provider providerconfig
...
+		// then cause cluster stability issues in some cases.
 		for _, nodeAddress := range node.Status.Addresses {
 			for instanceAddress := range instance.Addresses() {
+				// Hetzner is known to assign old IP addresses to new instances right after
+				// the old instance has been deleted. On Hetzner, the described issue happens much
+				// more often than for other providers, and as we know that the node name will
+				// always match the instance name, we can check that those two names match.
+				//
+				// TODO: Ideally, we should check this for other providers too, but it's not the
+				// case that node name and instance name will be same for all providers (e.g. AWS).
```
xmudrii, line 935: As far as I've seen, AWS is actually the only provider that has a different node and instance names. However, I'm not sure how to reliably confirm this.

irozzo-1A: I think it's ok until we have users reporting issues. Here we are covering a corner case that should only happen in very rare circumstances. Normally CCM is deployed and `providerID` is assigned.

pkg/controller/machine/machine.go
```diff
@@ -888,8 +911,31 @@ func (r *Reconciler) getNode(instance instance.Instance, provider providerconfig
...
+		// IP addresses as the instance that has just been deleted. If machine-controller
+		// processes the new Machine before removing the old Machine and the corresponding
+		// Node object, machine-controller could update the NodeOwner label on the old Node
+		// object to point to the Machine that just got created, as IP addresses would match.
+		// This causes machine-controller to fail to delete the old Node object, which could
+		// then cause cluster stability issues in some cases.
 		for _, nodeAddress := range node.Status.Addresses {
 			for instanceAddress := range instance.Addresses() {
+				// Hetzner is known to assign old IP addresses to new instances right after
+				// the old instance has been deleted. On Hetzner, the described issue happens much
```
irozzo-1A, line 930: > On Hetzner, the described issue happens much more often than for other providers

I don't think we have evidences that it happens more often with Hetzner than with other providers we support, simply we observed it with Hetzner. I think that almost all clusters we use for manual or automated tests have a CCM deployed. That said it's a picky comment, I let you decide if you want to rectify or keep it ;-)

xmudrii:
```
// We observed that the issue described above happens often on Hetzner.
// As we know that the Node and the instance name will always be same
// on Hetzner, we can use it as an additional check to prevent this
// issue.
// TODO: We should do this for other providers, but there are providers where
// the node and the instance names will not match, so it requires further
// investigation (e.g. AWS).
```

@irozzo-1A What do you think about the comment above?

xmudrii: @irozzo-1A Pushed the comment update.

PR #816, opened 2020-09-01
Ignore Node when deleting by NodeRef if it doesn't exist anymore

xmudrii:

There's a regression introduced in #813, which prevents the Machine object from being deleted if the Node referred by NodeRef is deleted.

**Optional Release Note**:
```release-note
Ignore Node when deleting by NodeRef if it doesn't exist anymore
```

### DIFF
--- a/pkg/controller/machine/machine.go
+++ b/pkg/controller/machine/machine.go
@@ -614,15 +614,22 @@ func (r *Reconciler) deleteNodeForMachine(machine *clusterv1alpha1.Machine) erro
 	if machine.Status.NodeRef != nil {
 		objKey := ctrlruntimeclient.ObjectKey{Name: machine.Status.NodeRef.Name}
 		node := &corev1.Node{}
+		nodeFound := true
 		if err := r.client.Get(r.ctx, objKey, node); err != nil {
-			return fmt.Errorf("failed to get node %s: %v", machine.Status.NodeRef.Name, err)
+			if !kerrors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


xrstf (approved):
/approve

pkg/controller/machine/machine.go
```diff
@@ -614,16 +614,23 @@ func (r *Reconciler) deleteNodeForMachine(machine *clusterv1alpha1.Machine) erro
 	if machine.Status.NodeRef != nil {
 		objKey := ctrlruntimeclient.ObjectKey{Name: machine.Status.NodeRef.Name}
 		node := &corev1.Node{}
+		nodeFound := true
 		if err := r.client.Get(r.ctx, objKey, node); err != nil {
```
irozzo-1A, line 618: I think it's more natural to express this in the opposite way:

```
if err := r.client.Get(r.ctx, objKey, node); err == nil {
			if err := r.client.Delete(r.ctx, node); err != nil {
				if !kerrors.IsNotFound(err) {
					return err
				}
				klog.V(2).Infof("node %q does not longer exist for machine %q", machine.Status.NodeRef.Name, machine.Spec.Name)
			}
} else {
			if !kerrors.IsNotFound(err) {
				return err					return fmt.Errorf("failed to get node %s: %v", machine.Status.NodeRef.Name, err)
			}				}
}
```

You don't need the boolean

irozzo-1A: Don't look at the indentation...

PR #821, opened 2020-09-21
Add support for setting arbitrary feature gates on the kubelet

xmudrii:

This PR adds support for setting (enabling and disabling) arbitrary feature gates on the Kubelet. This is useful in scenarios when an user needs to enable some cluster features that are disabled by default and/or provider-specific, such as CSI Migration.

The feature gates are enabled by setting the `--node-kubelet-feature-gates` flag on machine-controller. The feature gates are provided in the same format as to other Kubernetes components, i.e. key/value pair where the key is the feature gate name and the value is `true` or `false`. Currently, the feature gates are set on **all** machines/nodes, i.e. it's not possible to do it for a subset of machines.

Example usage: `./machine-controller --node-kubelet-feature-gates=CSIMigration=true,CSIMigrationAWS=true ...`

Relevant to https://github.com/kubermatic/kubeone/issues/616

**Optional Release Note**:
```release-note
Add the --node-kubelet-feature-gates flag to allow setting arbitrary feature gates on the kubelet
```

/assign @kron4eg

### DIFF
--- a/pkg/controller/machine/machine.go
+++ b/pkg/controller/machine/machine.go
@@ -128,6 +128,10 @@ type NodeSettings struct {
 	HyperkubeImage string
 	// The kubelet repository to use. Currently only Flatcar Linux uses it.
 	KubeletRepository string
+	// Translates to feature gates on the kubelet.
+	// Expected input: FeatureGateName=true|false,...
+	// Default: RotateKubeletServerCertificate=true
+	KubeletFeatureGates []string
 }
 
 type KubeconfigProvider interface {
@@ -699,6 +703,7 @@ func (r *Reconciler) ensureInstanceExistsForMachine(
 				PauseImage:            r.nodeSettings.PauseImage,
 				HyperkubeImage:        r.nodeSettings.HyperkubeImage,
 				KubeletRepository:     r.nodeSettings.KubeletRepository,
+				KubeletFeatureGates:   r.nodeSettings.KubeletFeatureGates,
 				NoProxy:               r.nodeSettings.NoProxy,
 				HTTPProxy:             r.nodeSettings.HTTPProxy,
 			}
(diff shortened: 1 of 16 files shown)


pkg/controller/machine/machine.go
```diff
@@ -128,6 +128,10 @@ type NodeSettings struct {
 	HyperkubeImage string
 	// The kubelet repository to use. Currently only Flatcar Linux uses it.
 	KubeletRepository string
+	// Translates to feature gates on the kubelet.
+	// Expected input: FeatureGateName=true|false,...
+	// Default: RotateKubeletServerCertificate=true
+	KubeletFeatureGates []string
```
kron4eg, line 134: Probably this need to be changed to `map[string]bool`, and `userdatahelper.KubeletFeatureGates()` used to init and mainly default this map.

xmudrii: We'd have to move `userdatahelper.KubeletFeatureGates()` to `cmd/`, as this struct is populated when setting up a controller in the `cmd/machine-controller` package. That makes sense to me, so should I go ahead and change this?

kron4eg: Or, we could have a "constructor" for `machinecontroller.NewNodeSettings(opts ...NodeSettingsOptions)` being called like 

```
machinecontroller.NewNodeSettings(machinecontroller.WithFeatureGates(kubeletFeatureGates...))
```

Where `WithFeatureGates()` is defined like `func WithFeatureGates(featureGates ...string) map[string]bool` and will default feature gates properly.

kron4eg: Of course all this will be called in `cmd/machine-controller/main.go`

xmudrii: I haven't implemented the constructor, but I moved the parsing and defaulting logic to `cmd/machine-controller/main.go`. PTAL and let me know what do you think about this solution.

kron4eg (approved):

xmudrii:
/hold
until I debug the E2E test failures

xmudrii:
Tests should be fixed now.
/retest
/hold cancel

PR #826, opened 2020-09-24
Fix releasever to 7 for RHEL and CentOS in yum docker repository

irozzo-1A:
Recently docker introduced `docker-ce` repository for CentOS 8 it only contains docker version `19.03.13-3`. This caused a regression when using RHEL/CentOS 8, because previously we were using CentOS 7 repo indistinctly for CentOS/RHEL 7/8 but now the docker versions we were installing are no longer available in CentOS 8 repo.

As we do not have OS versioning yet the quick and dirty solution is to always use CentOS 7 repo for the docker versions that are not available on both repos.

https://download.docker.com/linux/centos/8/x86_64/stable/Packages/
https://download.docker.com/linux/centos/7/x86_64/stable/Packages/

Fixes #

**Optional Release Note**:

```release-note
Fix docker installation for CentOS/RHEL 8
```

### DIFF
--- a/pkg/userdata/centos/provider.go
+++ b/pkg/userdata/centos/provider.go
@@ -195,6 +195,7 @@ write_files:
     yum-config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
 {{- /*	Due to DNF modules we have to do this on docker-ce repo
 		More info at: https://bugzilla.redhat.com/show_bug.cgi?id=1756473 */}}
+    sed -i 's/\$releasever/7/g' /etc/yum.repos.d/docker-ce.repo
     yum-config-manager --save --setopt=docker-ce-stable.module_hotfixes=true
 
 {{- /* We need to explicitly specify docker-ce and docker-ce-cli to the same version.
--- a/pkg/userdata/centos/testdata/kubelet-v1.15-aws.yaml
+++ b/pkg/userdata/centos/testdata/kubelet-v1.15-aws.yaml
@@ -69,6 +69,7 @@ write_files:
 
     yum install -y yum-utils
     yum-config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
+    sed -i 's/\$releasever/7/g' /etc/yum.repos.d/docker-ce.repo
     yum-config-manager --save --setopt=docker-ce-stable.module_hotfixes=true
 
     DOCKER_VERSION='18.09.9-3.el7'
--- a/pkg/userdata/centos/testdata/kubelet-v1.16-aws.yaml
+++ b/pkg/userdata/centos/testdata/kubelet-v1.16-aws.yaml
@@ -69,6 +69,7 @@ write_files:
 
     yum install -y yum-utils
     yum-config-manager --add-repo=https://download.docker.com/linux/centos/docker-ce.repo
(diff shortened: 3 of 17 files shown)


kron4eg:
It's impossible to deduct CentOS release from the kubelet version, this logic is flawed.

For now I propose to patch .repo unconditionally, once we have somehow figured out OS versioning we could return to this part.

irozzo-1A:
> It's impossible to deduct CentOS release from the kubelet version, this logic is flawed.
> 
> For now I propose to patch .repo unconditionally, once we have somehow figured out OS versioning we could return to this part.

@kron4eg I'm not deducing the CentOS release from the kubelet version, I'm unconditionally patching the repo for docker versions that are not present in both CentOS 7 and 8 repos, that is 19.03.13-3 only. The problem is that kubernetes versions previous to 1.17 do not work with this docker version.

kron4eg:
> versions that are not present in both CentOS 7 and 8 repos, that is all but 19.03.13-3.

```bash
[root@bc944ddd6770 ~]# yum install docker-ce-${DOCKER_VERSION} docker-ce-cli-${DOCKER_VERSION}
Failed to set locale, defaulting to C.UTF-8
Last metadata expiration check: 0:01:02 ago on Thu Sep 24 20:00:30 2020.
Dependencies resolved.
================================================================================
 Package                      Arch   Version             Repository        Size
================================================================================
Installing:
 docker-ce                    x86_64 3:19.03.13-3.el7    docker-ce-stable  24 M
 docker-ce-cli                x86_64 1:19.03.13-3.el7    docker-ce-stable  38 M
```

This is with patched .repo file. Latest an greatest `19.03.13` in $releasever=7 repo.

irozzo-1A:
> This is with patched .repo file. Latest an greatest `19.03.13` in $releasever=7 repo.

I'm not sure to understand your point, let me try to explain my logic again. 

At the moment `19.03.13` is the only version that is present on both centos 7 and 8 repos (it's also the only version present in centos 8 docker stable repo). It is preferable and safer to to use the appropriate repo instead of enforcing centos 7 when we can use such version, that is for k8s versions >= 1.17.

Unfortunately k8s versions < 1.17 do not support docker versions more recent than 18.06, in those cases we are obliged to use centos 7 repo.

kron4eg:
My point is that we have everything what we need in `centos/7` repo, we don't need any guessing and please always patch the .repo file. We will do the proper fix once we decided on OS versioning.

xrstf:
When there are no downsides to always using the repo for 7, I would prefer the simpler solution. It doesn't introduce a preliminary OS versioning hack that we would have to revert later on.

irozzo-1A:
@xrstf  Well using the repository for another version and doing a [hack](https://bugzilla.redhat.com/show_bug.cgi?id=1756473) for installing docker is something I consider a downside.

 It's true that it is slightly more complex, but has the benefit of using the appropriate repository for all k8s versions >= 1.17 and it's already implemented. 

What kind of risks you foresee with this approach?

xmudrii:
I agree with @xrstf and @kron4eg. I'd prefer that we keep it simple and just use the CoreOS7 repo for both versions until we don't get the OS versioning.

irozzo-1A:
@xrstf @kron4eg As the consensus it to always use the centos7 repo I will conform to it. I don't fully understand the logic behind this position, but this has to be fixed soon.

xrstf (approved):
/approve

PR #830, opened 2020-10-05
Fix polkit rules filename for flatcar reboot.

eduardostalinho:

This bug was introduced in https://github.com/kubermatic/machine-controller/pull/824.
The name of the file to be created must be suffixed with `.rules` instead of `.rule`

**Optional Release Note**:

```release-note
Fix polkit rule filename to allow `core` user to rebbot machine for auto updates.
```


xmudrii (approved):
/lgtm
/approve

PR #837, opened 2020-10-16
Use Hyperkube Kubelet for 1.18 clusters

xmudrii:

We already use Hyperkube Kubelet for clusters running Kubernetes
1.17 and older. This PR extends this to 1.18, as there are no
Poseidon Kubelet images for the latest 1.18 releases any longer.

**Optional Release Note**:
```release-note
Use Hyperkube Kubelet for 1.18 clusters
```

/assign @kron4eg

### DIFF
--- a/pkg/userdata/flatcar/provider.go
+++ b/pkg/userdata/flatcar/provider.go
@@ -33,7 +33,7 @@ import (
 )
 
 const (
-	lessThen118Check = "< 1.18"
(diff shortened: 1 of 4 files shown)


irozzo-1A:
From what I saw we have not e2e tests for k8s 1.18/1.19 probably it would be better to add them.

kron4eg (approved):

PR #840, opened 2020-10-21
Support cloud-init for flatcar

mfranczy:
There is a requirement from one of our customers to use cloud-init together with flatcar,
thus this PR offers such a possibility.

Based on `privosioningUtility` option in operating system spec in machine deployment, cloud-init or igniton will be used.
```
operatingSystemSpec:
    privosioningUtility: cloud-init
```

By default ignition is set as a primary provisioner.

Fixes #

Before I go depper with the implementation let's sync if this solution is acceptable.

**Optional Release Note**:

```release-note
NONE
```


mfranczy:
@moadqassem the implementation is a bit different than we spoke, instead of extending `Plugins` struct, I think it would be better to decide if you want to use cloud-init or ignition directly in machine deployment. What do you think?

pkg/userdata/flatcar/provider.go
```diff
@@ -430,3 +456,6 @@ storage:
 {{ safeDownloadBinariesScript .KubeletVersion | indent 10 }}
           systemctl disable download-script.service
 `
+
+// coreos-cloud-config template
+const userDataCloudInitTemplate = ``
```
mfranczy, line 461: I will provide that soon.

pkg/userdata/flatcar/testdata/v1.17.0.yaml
```diff
@@ -1,464 +1 @@
...
...n/iscsiadm \\\n  -v /var/lib/calico:/var/lib/calico:ro \\\n  -v /var/lib/cni:/var/lib/cni \\\n  -v /var/lib/docker:/var/lib/docker \\\n  -v /var/lib/kubelet:/var/lib/kubelet:rshared \\\n  -v /var/log/pods:/var/log/pods \\\n  for-kubernetes-less-then-1.19/hyperkubeImage:v1.17.0 \\\n  --bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf \\\n  --kubeconfig=/var/lib/kubelet/kubeconfig \\\n  --config=/etc/kubernetes/kubelet.conf \\\n  --network-plugin=cni \\\n  --cni-conf-dir=/etc/cni/net.d \\\n  --cni-bin-dir=/opt/cni/bin \\\n  --cert-dir=/etc/kubernetes/pki \\\n  --cloud-provider=vsphere \\\n  --cloud-config=/etc/kubernetes/cloud-config \\\n  --hostname-override=node1 \\\n  --dynamic-config-dir=/etc/kubernetes/dynamic-config-dir \\\n  --exit-on-lock-contention \\\n  --lock-file=/tmp/kubelet.lock \\\n  --volume-plugin-dir=/var/lib/kubelet/volumeplugins \\\n  --node-ip ${KUBELET_NODE_IP}\nExecStop=-/usr/bin/docker stop %n\nRestart=always\nRestartSec=10\n[Install]\nWantedBy=multi-user.target\n","enabled":true,"name":"kubelet.service"},{"dropins":[{"contents":"[Service]\nEnvironmentFile=-/etc/environment\n","name":"10-environment.conf"}],"enabled":true,"name":"docker.service"}]}}
```
mfranczy, line 1: I will change that to more readable form. For now i just wanted to see that unit tests are passing.

moadqassem:
> @moadqassem the implementation is a bit different than we spoke, instead of extending `Plugins` struct, I think it would be better to decide if you want to use cloud-init or ignition directly in machine deployment. What do you think?

Absolutely, this is a very valid approach and I think it is better, due to the ability of accessing the cloud provider information while deciding which cloud-config format to use. Will review the changes and let you know about what we need to adjust(if something needs to be adjusted).

mfranczy:
@moadqassem do we have flaky tests in machine-controller? Before OpenStack tests failed, now it's ok.

mfranczy:
The PR is not WIP anymore, from my perspective is done and good for review, there is one more thing to confirm, that the cloud-config template is ok. I will confirm it with @erdii very soon.

moadqassem:
@mfranczy I found out why the flatcar tests on aws are failing and I am working on a fix atm.

pkg/userdata/flatcar/provider.go
```diff
@@ -430,3 +456,289 @@ storage:
 {{ safeDownloadBinariesScript .KubeletVersion | indent 10 }}
           systemctl disable download-script.service
 `
+
+// Coreos cloud-config template
+const userDataCloudInitTemplate = `#cloud-config
```
moadqassem, line 461: Is it possible to simply merge values where both format are similar? for example, the units in coreos-cloudinit are one to one, to those in ignition. So why not simply have a var called `units` and `files` and add them to both templates.

mfranczy: It's not possible as there are differences with naming both for units and files, like `content` vs `contents`, `mode` vs `permissions` and so on.

pkg/userdata/flatcar/flatcar.go
```diff
@@ -22,11 +22,22 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+type ProvisioningUtility string
```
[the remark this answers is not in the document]
mfranczy: done

pkg/userdata/convert/ignition-converter.go
```diff
@@ -55,7 +55,7 @@ func ToIgnition(s string) (string, error) {
 		return "", fmt.Errorf("failed to convert container linux config to ignition: %s", report.String())
 	}
 
-	out, err := json.Marshal(ignCfg)
+	out, err := json.MarshalIndent(ignCfg, "", "  ")
```
moadqassem, line 58: Please revert this line here as it applies Indent to format the output and this increases the size of the generated cloud config. AWS cloud provider doesn't permit more than 16384 bytes

mfranczy: damn, nice catch!

mfranczy:
@moadqassem all tests finally passed.

PR #843, opened 2020-10-29
Implement a configurable timeout for OpenStack

cedi:
Make instanceReadyCheckTimeout for OpenStack Provider configurable

#819 

This haven't been tested properly yet.
It is still work in progress.

**Optional Release Note**:

```release-note
[action required] instanceReadyCheckTimeout `instanceReadyCheckPeriod` and `instanceReadyCheckTimeout` are now configurable to allow alternative Timeouts and Re-Tries. 120s is the default 

```

### DIFF
--- a/pkg/apis/cluster/v1alpha1/conversions/testdata/machinesv1alpha1machine/openstack.yaml
+++ b/pkg/apis/cluster/v1alpha1/conversions/testdata/machinesv1alpha1machine/openstack.yaml
@@ -35,7 +35,7 @@ spec:
           namespace: kube-system
           name: machine-controller-openstack
           key: domainName
-    # If empty, ca be set via OS_TENANT_NAME env var
+   # If empty, ca be set via OS_TENANT_NAME env var
       tenantName:
         secretKeyRef:
           namespace: kube-system
@@ -58,6 +58,10 @@ spec:
       network: ""
       # Only required if the network has more than one subnet
       subnet: ""
+      # Can be increased for slower OpenStack setups. No values below 1m (60s) possible as this makes no sense.
+      instanceReadyCheckPeriod: 2m
+      # Can be increased for slower OpenStack setups. No values below 1m (60s) possible as this makes no sense.
+      instanceReadyCheckTimeout: 2m
       # the list of tags you would like to attach to the instance
       tags:
         tagKey: tagValue
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -62,7 +62,7 @@ type clientGetterFunc func(c *Config) (*gophercloud.ProviderClient, error)
 
 // serverReadinessWaiterFunc waits for the server with the given ID to be
 // ACTIVE.
-type serverReadinessWaiterFunc func(computeClient *gophercloud.ServiceClient, serverID string) error
+type serverReadinessWaiterFunc func(computeClient *gophercloud.ServiceClient, serverID string, instanceReadyCheckPeriod time.Duration, instanceReadyCheckTimeout time.Duration) error
 
 type provider struct {
 	configVarResolver     *providerconfig.ConfigVarResolver
@@ -101,15 +101,15 @@ type Config struct {
 	RootDiskSizeGB        *int
 	NodeVolumeAttachLimit *uint
 
+	InstanceReadyCheckPeriod  time.Duration
+	InstanceReadyCheckTimeout time.Duration
+
 	Tags map[string]string
 }
 
 const (
 	machineUIDMetaKey = "machine-uid"
 	securityGroupName = "kubernetes-v1"
-
-	instanceReadyCheckPeriod  = 2 * time.Second
-	instanceReadyCheckTimeout = 2 * time.Minute
 )
 
 // Protects floating ip assignment
@@ -146,6 +146,35 @@ func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *providerconfigt
 	c.Region, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Region, "OS_REGION_NAME")
 	if err != nil {
 		klog.V(6).Infof("Region from configuration or environment variable not found")
+	}
+
+	// Load timeout
+	instanceReadyCheckPeriodStr, err := p.configVarResolver.GetConfigVarStringValue(rawConfig.InstanceReadyCheckPeriod)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"InstanceReadyCheckPeriod\" field, error = %v", err)
+	}
+
+	c.InstanceReadyCheckPeriod, err = time.ParseDuration(instanceReadyCheckPeriodStr)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to parse the value of \"InstanceReadyCheckPeriod\" field (%s), error = %v", instanceReadyCheckPeriodStr, err)
+	}
+
+	if c.InstanceReadyCheckPeriod < 60*time.Second {
+		c.InstanceReadyCheckPeriod = 120 * time.Second
+	}
+
+	instanceReadyCheckTimeoutStr, err := p.configVarResolver.GetConfigVarStringValue(rawConfig.InstanceReadyCheckTimeout)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"InstanceReadyCheckTimeout\" field, error = %v", err)
+	}
+
+	c.InstanceReadyCheckTimeout, err = time.ParseDuration(instanceReadyCheckTimeoutStr)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to parse the value of \"InstanceReadyCheckTimeout\" field (%s), error = %v", instanceReadyCheckTimeoutStr, err)
+	}
+
+	if c.InstanceReadyCheckPeriod < 60*time.Second {
+		c.InstanceReadyCheckPeriod = 120 * time.Second
 	}
 
 	// We ignore errors here because the OS domain is only required when using Identity API V3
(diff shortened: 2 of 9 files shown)


test/e2e/provisioning/all_e2e_test.go
```diff
@@ -500,9 +504,11 @@ func TestUbuntuProvisioningWithUpgradeE2E(t *testing.T) {
 	osUsername := os.Getenv("OS_USERNAME")
 	osTenant := os.Getenv("OS_TENANT_NAME")
 	osNetwork := os.Getenv("OS_NETWORK_NAME")
+	osInstanceReadyCheckPeriod := os.Getenv("INSTANCE_READY_CHECK_PERIOD")
+	osInstanceReadyCheckTimeout := os.Getenv("INSTANCE_READY_TIMEOUT")
 
-	if osAuthURL == "" || osUsername == "" || osPassword == "" || osDomain == "" || osRegion == "" || osTenant == "" {
-		t.Fatal("unable to run test, all of OS_AUTH_URL, OS_USERNAME, OS_PASSOWRD, OS_REGION, OS_TENANT and OS_DOMAIN must be set!")
+	if osAuthURL == "" || osUsername == "" || osPassword == "" || osDomain == "" || osRegion == "" || osTenant == "" || osInstanceReadyCheckPeriod == "" || osInstanceReadyCheckTimeout == "" {
+		t.Fatal("unable to run test suite, all of OS_AUTH_URL, OS_USERNAME, OS_PASSOWRD, OS_REGION, OS_TENANT OS_DOMAIN, INSTANCE_READY_CHECK_PERIOD, and INSTANCE_READY_TIMEOUT must be set!")
```
kron4eg, line 511: I don't think this should be true, those timeouts can and should be defaulted to some value (like they do in normal flow) so they can avoid being required.

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -128,9 +128,11 @@ func TestOpenstackProvisioningE2E(t *testing.T) {
 	osUsername := os.Getenv("OS_USERNAME")
 	osTenant := os.Getenv("OS_TENANT_NAME")
 	osNetwork := os.Getenv("OS_NETWORK_NAME")
+	osInstanceReadyCheckPeriod := os.Getenv("INSTANCE_READY_CHECK_PERIOD")
+	osInstanceReadyCheckTimeout := os.Getenv("INSTANCE_READY_TIMEOUT")
 
-	if osAuthURL == "" || osUsername == "" || osPassword == "" || osDomain == "" || osRegion == "" || osTenant == "" {
-		t.Fatal("unable to run test suite, all of OS_AUTH_URL, OS_USERNAME, OS_PASSOWRD, OS_REGION, OS_TENANT and OS_DOMAIN must be set!")
+	if osAuthURL == "" || osUsername == "" || osPassword == "" || osDomain == "" || osRegion == "" || osTenant == "" || osInstanceReadyCheckPeriod == "" || osInstanceReadyCheckTimeout == "" {
+		t.Fatal("unable to run test suite, all of OS_AUTH_URL, OS_USERNAME, OS_PASSOWRD, OS_REGION, OS_TENANT OS_DOMAIN, INSTANCE_READY_CHECK_PERIOD, and INSTANCE_READY_TIMEOUT must be set!")
```
kron4eg, line 135: Same here with timeouts defaulting.

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -148,6 +148,25 @@ func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *providerconfigt
 		klog.V(6).Infof("Region from configuration or environment variable not found")
 	}
 
+	// Load timeout
+	instanceReadyCheckPeriod, err := p.configVarResolver.GetConfigVarIntValueOrEnv(rawConfig.InstanceReadyCheckPeriod, "INSTANCE_READY_CHECK_PERIOD")
```
kron4eg, line 152: Instead of plain int type and introducing new family of functions (GetConfigVarInt...), I'd prefer for https://golang.org/pkg/time/#ParseDuration to be used on a string value.

kron4eg:
Thanks @cedi!

pkg/cloudprovider/provider/openstack/provider.go
cedi: Ah, very good catch! I was unhappy with the int too, as I felt it was too ambiguous in regarding what it represents (timeout in seconds or minutes).

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -148,6 +148,25 @@ func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *providerconfigt
...
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"InstanceReadyCheckPeriod\" field, error = %v", err)
+	}
+	if instanceReadyCheckPeriod < 60 {
+		instanceReadyCheckPeriod = 120
+	}
+	c.InstanceReadyCheckPeriod = time.Duration(instanceReadyCheckPeriod) * time.Second
+
+	instanceReadyCheckTimeout, err := p.configVarResolver.GetConfigVarIntValueOrEnv(rawConfig.InstanceReadyCheckTimeout, "INSTANCE_READY_TIMEOUT")
```
moadqassem, line 161: Please change this to ` p.configVarResolver.GetConfigVarIntValue` as this value can be added explicitly: 
```go
instanceReadyCheckPeriod, err := p.configVarResolver.GetConfigVarIntValue(rawConfig.InstanceReadyCheckPeriod)
```

pkg/cloudprovider/provider/openstack/provider.go
moadqassem, line 152: Please change this to ` p.configVarResolver.GetConfigVarIntValue` as this value can be added explicitly: 
```go
instanceReadyCheckPeriod, err := p.configVarResolver.GetConfigVarIntValue(rawConfig.InstanceReadyCheckPeriod)
```

moadqassem (commented):
@cedi Thanks for the PR. The current PR introduces new helper functions in order to treat int values which got from secrets. Where I find this as a good utility to have, however, for the fields which these functions are created, shouldn't be treated that way and should be added explicitly instead of deriving them from secrets. I am not quite sure that there is a case where we have(or will have) a value which holds a sensitive information and being of a type int. Thus I would suggest to remove these helper functions.

test/e2e/provisioning/all_e2e_test.go
cedi: could you give me a pointer where to add the default values for this variables?

pkg/apis/cluster/v1alpha1/conversions/testdata/machinesv1alpha1machine/openstack.yaml
```diff
@@ -35,6 +35,18 @@ spec:
           namespace: kube-system
           name: machine-controller-openstack
           key: domainName
+    # If empty, ca be set via INSTANCE_READY_CHECK_PERIOD env var
+      instanceReadyCheckPeriod:
```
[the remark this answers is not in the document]
cedi: Fair. Changed it

pkg/cloudprovider/provider/openstack/provider.go
cedi: lgtm

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -148,6 +148,37 @@ func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *providerconfigt
...
+	}
+
+	instanceReadyCheckTimeoutStr, err := p.configVarResolver.GetConfigVarStringValue(rawConfig.InstanceReadyCheckTimeout)
+	if err != nil {
+		return nil, nil, nil, fmt.Errorf("failed to get the value of \"InstanceReadyCheckTimeout\" field, error = %v", err)
+	}
+
+	c.InstanceReadyCheckTimeout, err = time.ParseDuration(instanceReadyCheckTimeoutStr)
+
+	c.InstanceReadyCheckPeriod, err = time.ParseDuration(instanceReadyCheckPeriodStr)
```
kron4eg, line 173:
```suggestion
```

is already parsed and error is handled previously.

cedi: ufff. copy/paste error... Thanks for pointing this one out!

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -500,9 +504,11 @@ func TestUbuntuProvisioningWithUpgradeE2E(t *testing.T) {
 	osUsername := os.Getenv("OS_USERNAME")
 	osTenant := os.Getenv("OS_TENANT_NAME")
 	osNetwork := os.Getenv("OS_NETWORK_NAME")
+	osInstanceReadyCheckPeriod := os.Getenv("INSTANCE_READY_CHECK_PERIOD")
```
moadqassem, line 507: Why `INSTANCE_READY_CHECK_PERIOD` and `INSTANCE_READY_TIMEOUT` are is read from the env vars? I would suggest that you simply set them in the template: 
https://github.com/kubermatic/machine-controller/blob/master/test/e2e/provisioning/testdata/machinedeployment-openstack-upgrade.yml#L26-L38

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -128,9 +128,11 @@ func TestOpenstackProvisioningE2E(t *testing.T) {
...
```
moadqassem, line 131: Why INSTANCE_READY_CHECK_PERIOD and INSTANCE_READY_TIMEOUT are is read from the env vars? I would suggest that you simply set them in the template:
https://github.com/kubermatic/machine-controller/blob/master/test/e2e/provisioning/testdata/machinedeployment-openstack.yaml#L26-L36

moadqassem (changes requested):

moadqassem:
@cedi Thank you very much for you PR. 
/approve 
/lgtm

PR #844, opened 2020-11-02
Create workaround for kubelet stopped posting node status bug

phiphi282:

Sometimes the kubelet suddenly starts to hang and the kubelet needs to be restarted, or even the node if there is no ssh access.
This pr adds a script to cloud init to check the kubelet logs periodically and restarts the kubelet if needed.

This is because of this bug in kubernetes: https://github.com/kubernetes/kubernetes/issues/87615
That will probably be fixed with the first kubernetes version that will use golang 1.16 (probably earliest version will be 1.21).

Fixes #

We have only tested this for ubuntu, flatcar and coreos.
It is possible that there are some issues with RHEL, SLES and centos.
If there are any e2e tests that should be extended as well then please let me know.

**Optional Release Note**:

```release-note
automatically restart kubelets that stop posting node status
```


pkg/userdata/coreos/testdata/v1.17.0-vsphere-proxy.yaml
```diff
@@ -176,6 +190,25 @@ systemd:
...
   files:
+    - path: /opt/kubelet-restart.sh
+      filesystem: root
+      mode: 0755
+      contents:
+        inline: |
+          #!/bin/bash
+
+          while true; do
+              output=$(journalctl -u kubelet -n 1 | grep "use of closed network connection")
```
moelsayed, line 201: nit: you might want to use `grep -q`

pkg/userdata/helper/kubelet.go
```diff
@@ -243,3 +243,36 @@ ExecStart=/opt/bin/health-monitor.sh container-runtime
 [Install]
 WantedBy=multi-user.target`
 }
+
+// KubeletRestartOnNotReadyScript script to check kubelet stopped posting node status every 10 minutes
+func KubeletRestartOnNotReadyScript() string {
+	return `#!/bin/bash
+
+while true; do
+    output=$(journalctl -u kubelet -n 1 | grep "use of closed network connection")
```
moelsayed, line 252: nit: Please use `grep -q`

pkg/userdata/helper/kubelet.go
```diff
@@ -243,3 +243,36 @@ ExecStart=/opt/bin/health-monitor.sh container-runtime
 [Install]
 WantedBy=multi-user.target`
 }
+
+// KubeletRestartOnNotReadyScript script to check kubelet stopped posting node status every 10 minutes
+func KubeletRestartOnNotReadyScript() string {
+	return `#!/bin/bash
```
moelsayed, line 249: nit: please use `#!/usr/bin/env bash`

pkg/userdata/rhel/provider.go
```diff
@@ -40,7 +40,7 @@ type Provider struct{}
 func (p Provider) UserData(req plugin.UserDataRequest) (string, error) {
 	tmpl, err := template.New("user-data").Funcs(userdatahelper.TxtFuncMap()).Parse(userDataTemplate)
 	if err != nil {
-		return "", fmt.Errorf("failed to parse user-data template: %v", err)
+		return "", fmt.Errorf("FAILED TO PARSE USER-DATA TEMPLATE: %V", err)
```
moelsayed, line 43: IDE going haywire?

moelsayed (changes requested):
Thank you for the PR. I left a few minor comments.
I am assuming, since the script only checks the last line in the logs, kubelet stops completely at this point?
Also, is running this every 10 minutes is enough? Wouldn't the node be unavailable at this point?

@kron4eg what do you think about RHEL/CentOS? Would this be safe?

kron4eg:
But what about `/etc/systemd/system/kubelet-healthcheck.service` and its `/opt/bin/health-monitor.sh kubelet`? I suppose we can modify health-monitor.sh instead of introducing new scripts (and wasting precious user-data space).

pkg/userdata/rhel/provider.go
phiphi282: Yeah looks like it 😅

pkg/userdata/scripts/health-monitor.sh
```diff
@@ -77,10 +77,18 @@ function kubelet_monitoring {
...
+      failed=true
+      echo "Kubelet stopped posting node status. Restarting"
+    elif ! output=$(curl -m "${max_seconds}" -f -s -S http://127.0.0.1:10248/healthz 2>&1); then
+      failed=true
       # Print the response and/or errors.
-      echo $output
+      echo "$output"
+    fi
+
+    if [ $failed ]; then
```
kron4eg, line 91: there are no booleans in shell, the `if` body will always evaluate unconditionally.

kron4eg (commented):
https://github.com/kubermatic/machine-controller/blob/master/pkg/userdata/helper/download_binaries_script.go#L88 here it's hardcoded the git revision to download. So after merging this PR the download_binaries_script.go has to be updated too (but only after since the revision is unknown).

kron4eg:
This PR is replaced by the #901

PR #863, opened 2020-12-12
Update to controller-runtime 0.7.0

xrstf:
What it says on the tin. Lots of improvements worth having and also a couple of important, user-visible changes:

* the example manifests have been updated:
  * RBAC is using `rbac.authorization.k8s.io/v1`
  * MutatingWebhook is using `admissionregistration.k8s.io/v1`
* the leader election now uses the new `configmapsleases` resourcelock, which means that the machine-controller will hold *both* a ConfigMap and a coordination.k8s.io/v1/Lease as a means of slowly transitioning to only using Leases. This requires an updated Role in the kube-system namespace to grant the permissions. See https://github.com/kubernetes-sigs/controller-runtime/pull/1144 for more info.
* The metrics and the health endpoints cannot be on the same port anymore. `-internal-listen-address` has been replaced by `-health-probe-address` and `-metrics-address`.

Besides that, internally the changes are:

* lots and lots of contexts everywhere
* the admission handler now handles v1 reviews
* the home-grown leaderelection stuff in the main() has been replaced by just using controller-runtime features

**Optional Release Note**:
```release-note
BREAKING: leaderelection now uses both ConfigMaps and Leases, RBAC updates are required
BREAKING: `-internal-listen-address` has been replaced by `-health-probe-address` and `-metrics-address`; both addresses cannot use the same port
update to rbac.authorization.k8s.io/v1 and admissionregistration.k8s.io/v1
```


xrstf:
cc @alvaroaleman

pkg/controller/machine/machine_test.go
```diff
@@ -195,7 +197,7 @@ func TestController_GetNode(t *testing.T) {
 			client := ctrlruntimefake.NewFakeClient(nodes...)
 			reconciler := Reconciler{client: client}
 
-			node, exists, err := reconciler.getNode(test.instance, test.provider)
+			node, exists, err := reconciler.getNode(ctx, test.instance, test.provider)
```
moadqassem, line 200: This means you are using the same context for each tests case(probably wouldn't have a major effect). I would suggest to have a distinct context for each test cases.

```go 
node, exists, err := reconciler.getNode(context.Background(), test.instance, test.provider)
```

Or move the context in the for loop.

pkg/controller/machine/machine_test.go
```diff
@@ -276,6 +278,8 @@ func TestControllerDeletesMachinesOnJoinTimeout(t *testing.T) {
 		},
 	}
 
+	ctx := context.Background()
```
moadqassem, line 281: Please move the context in the tests for loop

moadqassem: Same goes to the lines below

cmd/machine-controller/main.go
```diff
@@ -299,246 +274,134 @@ func main() {
...
+		klog.Fatalf("failed to create runtime manager: %v", err)
+	}
+
+	if err := mgr.Start(ctx); err != nil {
+		klog.Errorf("failed to start kubebuilder manager: %v", err)
 	}
+}
 
-	id, err := os.Hostname()
+func createManager(ctx context.Context, syncPeriod time.Duration, options controllerRunOptions) (manager.Manager, error) {
```
moadqassem, line 290: The context parameter is not used, can we please get rid of it.

moadqassem:
@xrstf Really nice work! I left few minor comments

PR #867, opened 2020-12-16
WIP - Add contexts to cloudprovider interface

xrstf:
The admission webhook so far did not use contexts to control the lifetime of cloud provider API calls. This meant that these calls could block forever and slow the webhook down.

This PR adds an explicit context to the cloudprovider interface and uses the incoming request's context for all validation logic, cancelling all outgoing requests after a certain period of time. *However*, this leads to a problem: For slow cloud providers like Hetzner, the request is rarely processed within the kube-apiserver's timeout (10 seconds). Previously, our caching saved us (by accident?), as the validation logic continued and then cached the result long after the kube-apiserver stopped listening. And on the next admission request, it would respond with the cached result. But when the validation is cancelled thanks to the context timing out, this would not happen anymore.

I considered three approaches to generally handling slow cloud providers in the admission requests:

1. Ignore the problem, i.e. do not add contexts and keep yolo'ing. This is bad IMHO because most related Go APIs move towards having explicit contexts and it would be very nice if we can use those to actually control our logic. So far, there was no possibility to cancel a .Get() call for retrieving a machine.

    Also, I consider a cache to be optional by default, i.e. if you remove it, your application will run worse, potentially much worse, but it will still work. If we removed our caching however, the validation would never possibly work if a cloud provider is slow.

2. Be strict: Add contexts and cancel validation logic and then just blame cloud providers for being slow. This is easy, yet not very convenient for the user. I think we can do better.

3. Background processing. With this, we run a separate goroutine for each incoming request and then do some clever timeout handling. This allows us to let the validation complete on its own time and respond with a nicely worded "sorry, timeout, please try again later" to the user. Once the goroutine is done, the result is cached and the user will get a proper response.

I chose approach 3.

This makes the Hetzner tests look a bit more friendly already:

```
I1218 18:29:02  verify.go:150] Creating a new "ubuntu-1-18-10-hz-123" MachineDeployment
I1218 18:29:02  verify.go:150] Creating a new "centos-1-18-10-hz-123" MachineDeployment
I1218 18:29:02  verify.go:150] Creating a new "centos-1-19-3-hz-123" MachineDeployment
I1218 18:29:02  verify.go:150] Creating a new "centos-1-17-13-hz-123" MachineDeployment
I1218 18:29:02  verify.go:150] Creating a new "ubuntu-1-17-13-hz-123" MachineDeployment
I1218 18:29:02  verify.go:150] Creating a new "ubuntu-1-19-3-hz-123" MachineDeployment
I1218 18:29:02  verify.go:169] MachineDeployment "centos-1-19-3-hz-123" created
I1218 18:29:04  verify.go:169] MachineDeployment "centos-1-17-13-hz-123" created
W1218 18:29:11  verify.go:159] Creation of "ubuntu-1-18-10-hz-123" failed, retrying: admission webhook denied the request: validation timed out, please try again later
W1218 18:29:11  verify.go:159] Creation of "centos-1-18-10-hz-123" failed, retrying: admission webhook denied the request: validation timed out, please try again later
W1218 18:29:11  verify.go:159] Creation of "ubuntu-1-19-3-hz-123" failed, retrying: admission webhook denied the request: validation timed out, please try again later
W1218 18:29:11  verify.go:159] Creation of "ubuntu-1-17-13-hz-123" failed, retrying: admission webhook denied the request: validation timed out, please try again later
I1218 18:29:14  verify.go:169] MachineDeployment "centos-1-18-10-hz-123" created
I1218 18:29:22  verify.go:169] MachineDeployment "ubuntu-1-17-13-hz-123" created
W1218 18:29:23  verify.go:159] Creation of "ubuntu-1-18-10-hz-123" failed, retrying: admission webhook denied the request: validation timed out, please try again later
W1218 18:29:23  verify.go:159] Creation of "ubuntu-1-19-3-hz-123" failed, retrying: admission webhook denied the request: validation timed out, please try again later
I1218 18:29:26  verify.go:169] MachineDeployment "ubuntu-1-18-10-hz-123" created
I1218 18:29:26  verify.go:169] MachineDeployment "ubuntu-1-19-3-hz-123" created
```

**Optional Release Note**:
```release-note
NONE
```

### DIFF
--- a/pkg/admission/admission.go
+++ b/pkg/admission/admission.go
@@ -106,31 +106,102 @@ func createAdmissionResponse(original, mutated runtime.Object) (*admissionv1beta
 
 type mutator func(context.Context, admissionv1beta1.AdmissionReview) (*admissionv1beta1.AdmissionResponse, error)
 
+// handleFuncFactory wraps a mutator as an HTTP HandlerFunc.
 func handleFuncFactory(mutate mutator) func(http.ResponseWriter, *http.Request) {
+	/*
+		Within the handler we create below, the actual mutation logic is
+		running in a separate goroutine. This is because sometimes some
+		cloud provider APIs are slow, very slow, and the kube-apiserver is
+		impatient and will wait at most 30 seconds (10 by default) for a
+		response.
+
+		To ensure faster responses, we use our CloudproviderCache wrapper,
+		which will cache the results based on the MachineSpec's JSON
+		representation. While this just generally speeds up responses, this
+		becomes *critical* for slow cloud providers. For those, it can happen
+		that no validation ever succeeds within the given timeout. To ensure
+		that these providers can be used at all, we must let the validation run
+		in the background, cache its result and inform the user upon a timeout
+		that the validation is still happening and they should try again later,
+		when we have cached the result.
+
+		We plan the timing below based on a 10s timeout, as this is much less
+		painful for users that would have to wait up to 30s for any kind of
+		feedback for their `kubectl apply` command.
+	*/
 	return func(w http.ResponseWriter, r *http.Request) {
-		// cancel any validation after at most 30 seconds, which is the maximum time
-		// the kube-apiserver will every wait for a webhook to answer
-		ctx, cancel := context.WithTimeout(r.Context(), 30*time.Second)
-		defer cancel()
+		// this context is used for the validation logic
+		validationCtx := context.Background()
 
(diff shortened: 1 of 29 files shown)


xrstf:
So... err... this is weird, but it seems the Hetzner API has recovered and for a few days now, the API response times are sub 0.1 seconds, so the admission logic now simply works.

Let's close this, as we might not actually want or need it. But if the need arises in the future again, we can maybe use this PR as an inspiration.

PR #874, opened 2021-01-05
Specify resource group field for azure vnet

### LINKED ISSUE #836: Add the option to define a seperate ResourceGroup for the VNET in the Azure Cloud Provider
### Is your proposal related to a problem?

If the VNET in Azure is located in a different Ressource Group than the machines, the deployment of worker cluster is failing. The reason for this it, it is lookking for the network in the given Resource Group and it is not possible to add an optional RessourceGroup for the VNET.

### Describe the solution you'd like

The perfect solution would be to have a optional parameter vnetResourceGroup that if it is filled it uses the ResourceGroup that is given. If none is given just use the ResourceGroup wich is defined before. 

### Additional context

I created a gist with the changes i guess need to be made. Also i'm not shure if it is working and if the default option is working by falling back to the normal RessourceGroup if the VNetRessourceGroup is not given.
[You can find the gist here](https://gist.github.com/mfahlandt/70284ade3f5ad45b0803ac889d0bbeeb)

moadqassem:
Support specifying a different resource groups for vnets in Azure.

Fixes #836

**Optional Release Note**:
```release-note
Vnet Resource group support
```

### DIFF
--- a/docs/cloud-provider.md
+++ b/docs/cloud-provider.md
@@ -237,6 +237,8 @@ subscriptionID: "<< AZURE_SUBSCRIPTION_ID >>"
 location: "westeurope"
 # Azure resource group
 resourceGroup: "<< YOUR_RESOURCE_GROUP >>"
+# Azure resource group of the vnet	
+vnetResourceGroup: "<< YOUR_VNET_RESOURCE_GROUP >>"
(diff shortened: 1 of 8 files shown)


xrstf (approved):
/approve

PR #875, opened 2021-01-07
Include Kubernetes 1.20 in e2e tests

xrstf:
This makes sure we test against 1.20 as well. This PR also removes all manifests and tests for pre-1.17, as we do not support this anymore.

**Optional Release Note**:
```release-note
Support Kubernetes 1.20
```


test/e2e/provisioning/helper.go
```diff
@@ -33,9 +33,10 @@ var (
 	scenarios = buildScenarios()
 
 	versions = []*semver.Version{
-		semver.MustParse("v1.17.13"),
-		semver.MustParse("v1.18.10"),
-		semver.MustParse("v1.19.3"),
+		semver.MustParse("v1.17.16"),
```
irozzo-1A, line 36: In the test we are using K8S version 1.20, using kubelet 1.17 is out of skew policy. It should work anyway, but it's not guaranteed. We should invest some time to address this point (it can be a followup PR if is seems to work).

xrstf: Good point, we should fix this in the near future when we remove 1.17 support.

PR #892, opened 2021-02-04
Support image reference in azure

moadqassem:
Support azure custom image reference. This is needed in order to create worker nodes that run ubuntu 20.04 e.g and to support other os versions. The current `image_id` field doesn't work with public images and it only works with custom images.

Fixes #

**Optional Release Note**:
```release-note
Image Reference Support in Azure
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -312,6 +312,29 @@ presubmits:
               memory: 1Gi
               cpu: 500m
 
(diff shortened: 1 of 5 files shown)


kron4eg (approved):
❤️

kron4eg (approved):

moadqassem:
@kron4eg PTAL

kron4eg (approved):

PR #895, opened 2021-02-15
add -ca-bundle flag and support for custom CAs to vSphere, Openstack

xrstf:
This makes it possible to completely override the set of CA certificates used to communicate with the vSphere/Openstack APIs. If `-ca-bundle` is set, the certificates from the machine-controller(-webhook) pods are not used anymore.

The first commit shows the naive approach, where the CA bundle file is given through the stack until it arrives at the cloud provider instances. This was mainly because at the time, I thought the `SetRootCAs()` function was the only way to inject the CAs in the vSphere provider.

However the second commit retracts that approach and favors setting the CA as a global variable. I hate global state, but

* this approach uses much less ugly code
* this approach loads the CA bundle _once_ and then re-uses the x509.CertPool
* the CA bundle will not change during the runtime anyway

A new e2e test has been added that is doing a "negative test": We override the CA bundle with a useless set of random certificates and then _expect_ Machine validation from the webhook to _fail_. This demonstrates that the CA bundle is indeed loaded, parsed and used by the cloud provider (openstack in this case).

**Optional Release Note**:
```release-note
Add support for custom CAs for Openstack, vSphere
```

### DIFF
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -18,10 +18,8 @@ package main
 
 import (
 	"context"
-	"crypto/x509"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"net"
 	"net/http"
 	"net/http/pprof"
@@ -41,6 +39,7 @@ import (
 	clusterv1alpha1 "github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
 	"github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1/migrations"
 	cloudprovidertypes "github.com/kubermatic/machine-controller/pkg/cloudprovider/types"
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider/util"
 	"github.com/kubermatic/machine-controller/pkg/clusterinfo"
 	"github.com/kubermatic/machine-controller/pkg/containerruntime"
 	machinecontroller "github.com/kubermatic/machine-controller/pkg/controller/machine"
@@ -126,9 +125,6 @@ type controllerRunOptions struct {
 	// Enable NodeCSRApprover controller to automatically approve node serving certificate requests.
 	nodeCSRApprover bool
 
-	// caBundleFile is an optional path to a PEM-encoded CA certificate file
-	caBundleFile string
-
 	node machinecontroller.NodeSettings
 }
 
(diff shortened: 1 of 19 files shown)


pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -282,6 +283,15 @@ func getClient(c *Config) (*gophercloud.ProviderClient, error) {
...
 	}
 
+	// use the util's HTTP client to benefit from its CA bundle
+	httpConfig := util.HTTPClientConfig{
+		// before machine-controller v1.25 this used the default client with no timeout,
+		// so this is deliberately set to a high value to avoid too many disruptions
+		Timeout: 60 * time.Second,
+	}
+
+	pc.HTTPClient = httpConfig.New()
```
kron4eg, line 293: Here `pc.HTTPClient` is unconditionally overwritten from what might been set just few lines above. Looks wrong.

xrstf: Oh wow, yeah, that was a total brainfart.

cmd/webhook/main.go
```diff
@@ -46,10 +48,17 @@ func main() {
 	flag.StringVar(&admissionListenAddress, "listen-address", ":9876", "The address on which the MutatingWebhook will listen on")
 	flag.StringVar(&admissionTLSCertPath, "tls-cert-path", "/tmp/cert/cert.pem", "The path of the TLS cert for the MutatingWebhook")
 	flag.StringVar(&admissionTLSKeyPath, "tls-key-path", "/tmp/cert/key.pem", "The path of the TLS key for the MutatingWebhook")
+	flag.StringVar(&caBundleFile, "ca-bundle", "", "(optional) path to a file containing all PEM-encoded CA certificates (will be used instead of the host's certificates)")
```
moadqassem, line 51: Nit: Just to follow the pattern as the previous flags(I don't think the `optional` hint)
```suggestion
	flag.StringVar(&caBundleFile, "ca-bundle", "", "Path to a file containing all PEM-encoded CA certificates (will be used instead of the host's certificates if set)")
```

cmd/machine-controller/main.go
```diff
@@ -156,6 +158,7 @@ func main() {
 	flag.StringVar(&nodeKubeletRepository, "node-kubelet-repository", "quay.io/kubermatic/kubelet", "Repository for the kubelet container. Only has effect on Flatcar Linux, and for kubernetes >= 1.18.")
 	flag.StringVar(&nodeKubeletFeatureGates, "node-kubelet-feature-gates", "RotateKubeletServerCertificate=true", "Feature gates to set on the kubelet. Default: RotateKubeletServerCertificate=true")
 	flag.StringVar(&nodeContainerRuntime, "node-container-runtime", "docker", "container-runtime to deploy")
+	flag.StringVar(&caBundleFile, "ca-bundle", "", "(optional) path to a file containing all PEM-encoded CA certificates (will be used instead of the host's certificates)")
```
moadqassem, line 161: Nit: Just to follow the pattern as the previous flags(I don't think the optional hint)
```suggestion
	flag.StringVar(&caBundleFile, "ca-bundle", "", "Path to a file containing all PEM-encoded CA certificates (will be used instead of the host's certificates if set)")
```

pkg/cloudprovider/provider/vsphere/client.go
```diff
@@ -47,6 +50,11 @@ func NewSession(ctx context.Context, config *Config) (*Session, error) {
 		return nil, fmt.Errorf("failed to build client: %v", err)
 	}
 
+	// inject our global set of CA certificates
+	client.DefaultTransport().TLSClientConfig = &tls.Config{
```
kron4eg, line 54: Check this out what I've found: https://pkg.go.dev/github.com/vmware/govmomi/vim25/soap#Client.SetRootCAs

xrstf: Check out this PR's description:

> The first commit shows the naive approach, where the CA bundle file is given through the stack until it arrives at the cloud provider instances. This was mainly because at the time, I thought the SetRootCAs() function was the only way to inject the CAs in the vSphere provider.

kron4eg: well, yeah, but this global state is disgusting :D

xrstf:
/hold

still want to find a way to add some rudimentary tests

xrstf:
/hold cancel

e2e test has been added and seems to work pretty well :)

moadqassem (commented):
/approve
/lgtm

xrstf:
Thanks guys, we got this :-) :tada:

PR #898, opened 2021-02-18
[WIP] Save node settings in machine annotation

### LINKED ISSUE #897: Save node settings in annotation when machine is created
Right now there are some settings that are used to configure the worker nodes (e.g. Kubelet feature gates) that are passed as mc flags and are not persisted anywhere.
This makes it difficult to observe which configuration was used when machines are created.
In order to overcome this limitation I propose the following changes:

- When a new machine is created by the `machineset_controller` it should add an annotation containing a serialized version (e.g. json) of the [node settings](https://github.com/kubermatic/machine-controller/blob/ed7aeda28b232eb808693727250259f8cee22af1/pkg/controller/machine/machine_controller.go#L115-L137) (say `machine-controller.k8c.io/node-settings`). The ideal would be to have a mutating webhook taking care of this, but adding this logic in the [machineset_controller](https://github.com/kubermatic/machine-controller/blob/ed7aeda28b232eb808693727250259f8cee22af1/pkg/controller/machineset/machineset_controller.go#L322) directly is also an option.
- The `machine_controller` verifies if such annotation is present, and in such case it deserialize the node settings and uses them for preparing  the [userdata request](https://github.com/kubermatic/machine-controller/blob/ed7aeda28b232eb808693727250259f8cee22af1/pkg/controller/machine/machine_controller.go#L687-L701), instead of the ones provided at construction time.

irozzo-1A:

Fixes #897

**Optional Release Note**:

```release-note

```

### DIFF
--- a/pkg/controller/machine/machine_controller.go
+++ b/pkg/controller/machine/machine_controller.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"net"
 	"strings"
 	"time"
 
@@ -29,12 +28,12 @@ import (
 
 	"github.com/kubermatic/machine-controller/pkg/apis/cluster/common"
 	clusterv1alpha1 "github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
+	"github.com/kubermatic/machine-controller/pkg/apis/nodesettings"
 	"github.com/kubermatic/machine-controller/pkg/apis/plugin"
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider"
 	cloudprovidererrors "github.com/kubermatic/machine-controller/pkg/cloudprovider/errors"
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider/instance"
 	cloudprovidertypes "github.com/kubermatic/machine-controller/pkg/cloudprovider/types"
-	"github.com/kubermatic/machine-controller/pkg/containerruntime"
 	"github.com/kubermatic/machine-controller/pkg/node/eviction"
 	"github.com/kubermatic/machine-controller/pkg/providerconfig"
 	providerconfigtypes "github.com/kubermatic/machine-controller/pkg/providerconfig/types"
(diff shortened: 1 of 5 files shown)


pkg/controller/machine/machine_controller.go
```diff
@@ -683,21 +653,22 @@ func (r *Reconciler) ensureInstanceExistsForMachine(
 			if err != nil {
 				return nil, fmt.Errorf("failed to render cloud config: %v", err)
 			}
+			nodeSettings, err := nodesettings.FromMachine(machine)
```
irozzo-1A, line 656: check error

pkg/controller/machine/machine_controller.go
```diff
@@ -174,11 +146,9 @@ func Add(
 		kubeconfigProvider:               kubeconfigProvider,
 		providerData:                     providerData,
 		joinClusterTimeout:               joinClusterTimeout,
-		externalCloudProvider:            externalCloudProvider,
 		name:                             name,
 		bootstrapTokenServiceAccountName: bootstrapTokenServiceAccountName,
 		skipEvictionAfter:                skipEvictionAfter,
-		nodeSettings:                     nodeSettings,
```
irozzo-1A, line 181: keep for backward compatibility. e.g. machine created by old `machineset_controller` and processed by new `machine_controller`

xmudrii:
We discussed this PR internally. I'll try to recap the most important points here:

* The first problem is that we're applying the annotation in the `machineset_controller`. In most cases, this is okay, however, if someone would create a Machine object directly (without MachineDeployment/MachineSet), they wouldn't have the annotation on that Machine object
  * That's because if we just create the Machine object, then `machineset_controller` wouldn't do anything
  * @irozzo-1A proposed that we can pass NodeSettings to both the `machine_controller` and the `machineset_controller`, but we will still apply the annotation in the `machineset_controller`. In that case, `machine_controller` will have the needed data to create a Machine (if it's created directly without MachineDeployment/MachineSet), but there will be no annotation on that Machine object
  * The reason for applying that annotation in the `machineset_controller` is that we can do it in an atomic way, when creating the Machine object
  * Additionally, the reason for applying that annotation is that we get to keep the settings that were used for creating the Machine
  * I think that ideally that should be handled by `machine_controller`. We don't create Machine objects directly (we use MachineDeployments in both KKP and KubeOne), but I believe that Machines created manually should still have the annotation
* The second problem is should we keep the whole configuration JSON in the annotation
  * The main purpose of this is to detect is a Machine provisioned with ExternalCloudProvider enabled and with CSIMigration annotations set
  * For that use case, I think a much better solution would be to apply an annotation that will have the `true/false` semantics. For example, if the Machine was created with ExternalCloudProvider enabled, we would apply an annotation such as `k8c.io/external-cloud-provider: ""`
  * With such annotation, we can easily filter instances that have and that don't have the external cloud provider support enabled. It would be done just with a simple Kubernetes API call
  * If we would put a whole configuration in the annotation, we would have to:
    * Get the annotation
    * Unmarshal the annotation into the NodeSetting struct (this requires importing machine-controller and that struct)
    * Read the value from the struct
  * Additionally, this approach might require us to version the NodeSettings struct

As an additional note, we've discussed moving NodeSettings away from the machine-controller, but this is still to be decided.

irozzo-1A:
Thx for the comment @xmudrii, let me give more details about the choices I made here.

The idea of this PR is to complement the `Machine` spec with some settings, in order to avoid relying on implicit ephemeral information. As those settings should be considered as part of the spec I don't consider the `machine_controller` as the right place to set it. Note that I don't put this info in status (e.g. using conditions ) because it cannot be easily reconstructed by observation.
That said, the point you raised regarding individual machines created is valid. And I don't see at the moment any practical issue in adding the annotation here if it is not present:

https://github.com/kubermatic/machine-controller/blob/1eb46774f48f62d8aa785241e5f95b7fca662b74/pkg/controller/machine/machine_controller.go#L657

The cleanest would be to use a mutating webhook, but the drawback is that we should provide the `NodeSettings` to the webhook somehow and keep the flags in mc for backward compatibility.

Regarding the point about using structured data in the annotation, IMO as I already mentioned this has not to be seen as status but as spec integration (or simply a way of enabling some features/behavious). We could have a bunch of annotations, as some projects do (e.g. https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/), or adding the annotations for kubelet feature gates and external cloud provider flag only to start with, but IMO this is cleaner and a bit more consistent.

kron4eg:
#914 is inspired by this PR and will replace it.

PR #900, opened 2021-02-20
Add application credentials to openstack provider

### LINKED ISSUE #899: Support Openstack Application Credentials
In order to improve the security of the cloud.conf on Openstack provider, we should intergrate the option to add application credentials instead of username/password. 

An application credentials are limited to one project/tenant and belong to a user. This is not true for username/password what has certain security implications.

Acceptance criteria:
Allow the user to specify application-credential-id and application-credential-secret in the provider specs of Openstack.

Docs:
CCM: https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#global
Openstack: https://docs.openstack.org/keystone/latest/user/application_credentials.html

Happy2C0de:
Signed-off-by: Happy2C0de <46957159+Happy2C0de@users.noreply.github.com>

Add support for openstack application credentials

Fixes #899

Addition of `getConfigAuth` function in provider.go limits the cyclomatic complexity that was reached with a direct integration into `getConfig` function. From my point of view, excluding the retrieval of authentication information makes sense.

Please to proper e2e tests.

**Optional Release Note**:
```release-note
Add application credentials support to the Openstack provider
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -124,8 +124,9 @@ func (p *provider) getConfigAuth(c *Config, rawConfig *openstacktypes.RawConfig)
 		klog.V(6).Infof("applicationCredentialID from configuration or environment was found.")
 		c.ApplicationCredentialSecret, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.ApplicationCredentialSecret, "OS_APPLICATION_CREDENTIAL_SECRET")
 		if err != nil {
-			return fmt.Errorf("failed to get the value of \"applicationCredentialSecret\" field, error = %v", err)
+			return fmt.Errorf(`failed to get the value of "applicationCredentialSecret" field, error = %v`, err)
 		}
+		return nil
 	} else {
 		c.Username, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Username, "OS_USER_NAME")
 		if err != nil {
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -124,9 +124,10 @@ func (p *provider) getConfigAuth(c *Config, rawConfig *openstacktypes.RawConfig)
 		klog.V(6).Infof("applicationCredentialID from configuration or environment was found.")
 		c.ApplicationCredentialSecret, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.ApplicationCredentialSecret, "OS_APPLICATION_CREDENTIAL_SECRET")
 		if err != nil {
-			return fmt.Errorf(`failed to get the value of "applicationCredentialSecret" field, error = %v`, err)
+			return fmt.Errorf("failed to get the value of \"applicationCredentialSecret\" field, error = %v", err)
 		}
 		return nil
+	}
 	c.Username, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Username, "OS_USER_NAME")
 	if err != nil {
 		return fmt.Errorf("failed to get the value of \"username\" field, error = %v", err)
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -127,23 +127,21 @@ func (p *provider) getConfigAuth(c *Config, rawConfig *openstacktypes.RawConfig)
 			return fmt.Errorf(`failed to get the value of "applicationCredentialSecret" field, error = %v`, err)
 		}
 		return nil
-	} else {
-		c.Username, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Username, "OS_USER_NAME")
-		if err != nil {
-			return fmt.Errorf("failed to get the value of \"username\" field, error = %v", err)
-		}
-		c.Password, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Password, "OS_PASSWORD")
-		if err != nil {
-			return fmt.Errorf("failed to get the value of \"password\" field, error = %v", err)
-		}
-		c.TenantName, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.TenantName, "OS_TENANT_NAME")
-		if err != nil {
-			return fmt.Errorf("failed to get the value of \"tenantName\" field, error = %v", err)
-		}
-		c.TenantID, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.TenantID, "OS_TENANT_ID")
-		if err != nil {
-			return fmt.Errorf("failed to get the value of \"tenantID\" field, error = %v", err)
-		}
+	c.Username, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Username, "OS_USER_NAME")
+	if err != nil {
+		return fmt.Errorf("failed to get the value of \"username\" field, error = %v", err)
+	}
+	c.Password, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Password, "OS_PASSWORD")
+	if err != nil {
+		return fmt.Errorf("failed to get the value of \"password\" field, error = %v", err)
+	}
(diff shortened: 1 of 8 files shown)


pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -115,6 +117,36 @@ const (
...
 
+func (p *provider) getConfigAuth(c *Config, rawConfig *openstacktypes.RawConfig) error {
+	var err error
+	c.ApplicationCredentialID, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.ApplicationCredentialID, "OS_APPLICATION_CREDENTIAL_ID")
+	if err == nil {
+		klog.V(6).Infof("applicationCredentialID from configuration or environment was found.")
+		c.ApplicationCredentialSecret, _ = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.ApplicationCredentialSecret, "OS_APPLICATION_CREDENTIAL_SECRET")
+		if err != nil {
+			return fmt.Errorf("failed to get the value of \"applicationCredentialSecret\" field, error = %v", err)
+		}
```
kron4eg, line 128:
```suggestion
		if err != nil {
			return fmt.Errorf(`failed to get the value of "applicationCredentialSecret" field, error = %v`, err)
		}
		return nil
```

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -115,6 +117,36 @@ const (
...
+		c.TenantName, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.TenantName, "OS_TENANT_NAME")
+		if err != nil {
+			return fmt.Errorf("failed to get the value of \"tenantName\" field, error = %v", err)
+		}
+		c.TenantID, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.TenantID, "OS_TENANT_ID")
+		if err != nil {
+			return fmt.Errorf("failed to get the value of \"tenantID\" field, error = %v", err)
+		}
+	}
+	return nil
```
kron4eg, line 147:
```suggestion
	c.Username, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Username, "OS_USER_NAME")
	if err != nil {
		return fmt.Errorf("failed to get the value of \"username\" field, error = %v", err)
	}
	c.Password, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.Password, "OS_PASSWORD")
	if err != nil {
		return fmt.Errorf("failed to get the value of \"password\" field, error = %v", err)
	}
	c.TenantName, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.TenantName, "OS_TENANT_NAME")
	if err != nil {
		return fmt.Errorf("failed to get the value of \"tenantName\" field, error = %v", err)
	}
	c.TenantID, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.TenantID, "OS_TENANT_ID")
	if err != nil {
		return fmt.Errorf("failed to get the value of \"tenantID\" field, error = %v", err)
	}
	return nil
```

kron4eg (changes requested):

kron4eg:
@Happy2C0de thank you for the contribution. Really great to have it!

Happy2C0de:
> @Happy2C0de thank you for the contribution. Really great to have it!

@kron4eg No problem.
Thanks for the valuable change request. 👍

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -115,6 +117,35 @@ const (
...
+func (p *provider) getConfigAuth(c *Config, rawConfig *openstacktypes.RawConfig) error {
+	var err error
+	c.ApplicationCredentialID, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.ApplicationCredentialID, "OS_APPLICATION_CREDENTIAL_ID")
+	if err == nil {
+		klog.V(6).Infof("applicationCredentialID from configuration or environment was found.")
+		c.ApplicationCredentialSecret, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.ApplicationCredentialSecret, "OS_APPLICATION_CREDENTIAL_SECRET")
+		if err != nil {
+			return fmt.Errorf(`failed to get the value of "applicationCredentialSecret" field, error = %v`, err)
+		}
+		return nil
```
kron4eg, line 129: this will not compile

kron4eg:
```
$ make
GOOS=darwin go build -v \
        -ldflags '-s -w' \
        -o machine-controller \
        github.com/kubermatic/machine-controller/cmd/machine-controller
github.com/kubermatic/machine-controller/pkg/cloudprovider/provider/openstack
# github.com/kubermatic/machine-controller/pkg/cloudprovider/provider/openstack
pkg/cloudprovider/provider/openstack/provider.go:149:32: syntax error: unexpected v1alpha1, expecting comma or )
pkg/cloudprovider/provider/openstack/provider.go:154:6: no new variables on left side of :=
pkg/cloudprovider/provider/openstack/provider.go:158:6: rawConfig redeclared in this block
        previous declaration at pkg/cloudprovider/provider/openstack/provider.go:120:82
pkg/cloudprovider/provider/openstack/provider.go:163:4: no new variables on left side of :=
gmake: *** [Makefile:53: machine-controller] Error 2
```

kron4eg:
@Happy2C0de please see build log from build failure

Happy2C0de:
@kron4eg 
Thanks for reviewing and I apologise for the delayed reply.
/test pull-machine-controller-build

kron4eg:
@Happy2C0de webhook is not happy 

`verify.go:159] Creation of "ubuntu-upgrade-1365061762797277184" failed, retrying: admission webhook "machinedeployments.machine-controller.kubermatic.io" denied the request: failed to default machineSpec: failed to get a openstack client, due to You must provide a password to authenticate`

Happy2C0de:
@kron4eg 
Does your Openstack test cluster already support ApplicationCredentials?
See https://pkg.go.dev/github.com/gophercloud/gophercloud#AuthOptions
I think authentication should work now but it fails with the given error: "You must provide a password to authenticate" that comes from the provider. Username and Password are empty in `provider.go` at authentication time. LGTM
Local e2e tests are not possible for me such that I have to rely on your test pipelines.

kron4eg:
It looks like the error comes from here
https://github.com/gophercloud/gophercloud/blob/v0.14.0/auth_options.go#L285

means that no `ApplicationCredentialID` nor `ApplicationCredentialName` are set (or empty).

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -115,6 +117,36 @@ const (
 // Protects floating ip assignment
 var floatingIPAssignLock = &sync.Mutex{}
 
+func (p *provider) getConfigAuth(c *Config, rawConfig *openstacktypes.RawConfig) error {
+	var err error
+	c.ApplicationCredentialID, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.ApplicationCredentialID, "OS_APPLICATION_CREDENTIAL_ID")
+	if err == nil {
```
phiphi282, line 123: `GetConfigVarStringValueOrEnv` will actually never return an error.
Thus you will always return before reading in all the non applicationCredential related variables.

This is probably causing the e2e test error because the username and password are never set.

An option would be to just remove the return command or you could check if the applicationCredentialsID is not empty

Happy2C0de: @phiphi282 Good catch, I didn't see it. Let's see if this fixes the problem.

I would also suggest to add in a next PR an e2e test for applicationCredentials, see: 
https://github.com/kubermatic/machine-controller/blob/d7e05daefec4d9ba7a2a872ab1e4a50b8d4d08c9/test/e2e/provisioning/all_e2e_test.go#L70

e.g. OSAppCredsManifest, but that depends on the e2e test infra (OS_APPLICATION_CREDENTIAL_ID/SECRET are required)

kron4eg (approved):

kron4eg:
@Happy2C0de thanks for contributing!

PR #901, opened 2021-02-23
Create workaround for kubelet stopped posting node status bug

kron4eg:

text by @phiphi282 
Sometimes the kubelet suddenly starts to hang and the kubelet needs to be restarted, or even the node if there is no ssh access.
This pr adds a script to cloud init to check the kubelet logs periodically and restarts the kubelet if needed.

This is because of this bug in kubernetes: kubernetes/kubernetes#87615
That will probably be fixed with the first kubernetes version that will use golang 1.16 (probably earliest version will be 1.21).

Replaces #844

```release-note
Create workaround for kubelet stopped posting node status bug
```


xrstf (approved):
/approve

PR #902, opened 2021-02-24
Updated health monitor

kron4eg:
Use updated health monitor from #901 

**Optional Release Note**:

```release-note
health-monitor.sh also watches for kubelet network disconnects
```


xrstf (approved):
/approve

PR #911, opened 2021-03-03
handle misconfigured hetzner machines better

xrstf:
Fun fact: all these HCloud-Go API methods document

> If the ... does not exist, nil is returned.

As piping nil values through to hcloud-go will cause panics, this PR improves the error handling. Nonexisting resources are not treated as terminal errors, as their status can change at any time (e.g. if the user notices and creates the missing network).

**Optional Release Note**:
```release-note
Improve Hetzner error handling
```

### DIFF
--- a/pkg/cloudprovider/provider/hetzner/provider.go
+++ b/pkg/cloudprovider/provider/hetzner/provider.go
@@ -221,17 +221,25 @@ func (p *provider) Create(machine *v1alpha1.Machine, _ *cloudprovidertypes.Provi
 	}
 
 	if c.Datacenter != "" {
-		serverCreateOpts.Datacenter, _, err = client.Datacenter.Get(ctx, c.Datacenter)
+		dc, _, err := client.Datacenter.Get(ctx, c.Datacenter)
 		if err != nil {
 			return nil, hzErrorToTerminalError(err, "failed to get datacenter")
 		}
+		if dc == nil {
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -241,19 +249,30 @@ func (p *provider) Create(machine *v1alpha1.Machine, _ *cloudprovidertypes.Provi
...
 
-	serverCreateOpts.ServerType, _, err = client.ServerType.Get(ctx, c.ServerType)
+	serverType, _, err := client.ServerType.Get(ctx, c.ServerType)
 	if err != nil {
 		return nil, hzErrorToTerminalError(err, "failed to get server type")
 	}
+	if serverType == nil {
+		return nil, fmt.Errorf("server type %q does not exist", c.ServerType)
+	}
+	serverCreateOpts.ServerType = serverType
```
xrstf, line 275: I _hope_ that having nil for Image and ServerType was an error and not intended behaviour for when nothing is configured.

xrstf: Looking at https://github.com/hetznercloud/hcloud-go/blob/master/hcloud/server.go#L286-L295, not having these two fields should definitly blow up.

xmudrii (approved):
/lgtm

PR #914, opened 2021-03-09
Machine settings in annotations

### LINKED ISSUE #897: Save node settings in annotation when machine is created

kron4eg:
Save important settings (kubelet feature gates / external cloud provider) into annotation for consumption by other components.

Fixes #897

**Optional Release Note**:

```release-note
* Machine settings in annotations
* -external-cloud-provider -node-kubelet-feature-gates flags moved from controller to the webhook
```

### DIFF
--- a/pkg/apis/cluster/common/consts.go
+++ b/pkg/apis/cluster/common/consts.go
@@ -16,6 +16,14 @@ limitations under the License.
 
 package common
 
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
 // Constants aren't automatically generated for unversioned packages.
 // Instead share the same constant for all versioned packages
 type MachineStatusError string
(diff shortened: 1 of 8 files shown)


pkg/apis/cluster/common/consts.go
```diff
@@ -119,3 +127,63 @@ const (
...
+const (
+	// Annotation prefixes, used on Machine objects to indicate the parameters that been used to create those Machines
+	KubeletFeatureGatesAnnotationPrefixV1 = "kubelet-featuregates.machine-controller.kubermatic.io/v1"
+	KubeletFlagsGroupAnnotationPrefixV1   = "kubelet-flags.machine-controller.kubermatic.io/v1"
+)
+
+// SetKubeletFeatureGates marshal and save featureGates into metaobject annotations with
+// KubeletFeatureGatesAnnotationPrefixV1 prefix
+func SetKubeletFeatureGates(metaobj metav1.Object, featureGates map[string]bool) {
+	annts := metaobj.GetAnnotations()
```
xmudrii, line 146: Should we check if `annts` is `nil`?

xmudrii (approved):
/lgtm

PR #915, opened 2021-03-11
Update to Go 1.16.1

xrstf:
:tada:

**Optional Release Note**:
```release-note
Update to Go 1.16.1
```

### DIFF
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -8,7 +8,7 @@ presubmits:
       preset-goproxy: "true"
     spec:
       containers:
-        - image: golang:1.15.1
+        - image: golang:1.16.1
           command:
             - make
           args:
@@ -26,7 +26,7 @@ presubmits:
       preset-goproxy: "true"
     spec:
       containers:
-        - image: golang:1.15.1
+        - image: golang:1.16.1
           command:
             - make
           args:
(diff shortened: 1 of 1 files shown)


xrstf:
/hold

holding so i can time this properly.

kron4eg (approved):

uvwild:
> **What this PR does / why we need it**:
> tada
> 
> **Optional Release Note**:
> 
> ```
> Update to Go 1.16.1
> ```
just wondering if there is a deeper reason behind the go update. e.g. language capabilities?
otherwise this is rather confusing and tada does not quite carry a meaning for me.

In the real world there people usually don't have the choice to always update to the newest versions....
so knowing for what reason this has been updated would be quite beneficial.

xrstf:
We bump dependencies from time to time, just to gain bugfixes and other improvements. Updating Go can result in faster, smaller binaries. Go updates also contain fixes to its stdlib. It can also happen that we might need a recent Go version because some cloud provider SDK makes use of more recent features.

> In the real world there people usually don't have the choice to always update to the newest versions....

I'm sorry to hear that. You can still try to build your own machine-controller binaries with older Go versions, because if you take a peek at the [go.mod](https://github.com/kubermatic/machine-controller/blob/master/go.mod#L3) you will notice that we only depend on Go 1.13. :tada:

PR #917, opened 2021-03-11
use base64 encoded secret for kubevirt kubeconfig e2e tests

moadqassem:
Use base64 encoded kubeconfig for the KubeVirt cluster,
Fixes #

**Optional Release Note**:

```release-note
None
```


kron4eg (approved):

xrstf:
So the original problem is the way we inject values into the MachineDeployment testdata templates. In https://github.com/kubermatic/machine-controller/blob/master/test/e2e/provisioning/testdata/machinedeployment-kubevirt.yaml#L35 we have a singleline placeholder, but would inject a multiline YAML document. That breaks of course, because we break the indentation completely.

Our solution for now is to keep the kubeconfig as a singleline JSON string in Vault. This makes the injection harmless. There is a comment in the kubeconfig JSON that hints at this reason for it being JSON.

PR #930, opened 2021-03-23
Add tinkerbell as baremetal driver

### LINKED ISSUE #932: Add TinkerBell as a Baremetal driver in Machine Controller
**User Story**
As part of the Baremetal cluster provisioning, TinkerBell(TB) is considered one of good options to bootstrap and provision baremetal machines using containerized env. Thus, Machine Controller should provide a way to communicate with TB services to provision those servers/machines.

**Acceptance Criteria**
- Adding TB as a driver option in the machine controller Baremetal provider
- The driver is able to register hardwares(machines) run workflows to provision those machines
- Machine controller should persist an os based on the workflow and image repo setup in TB
- Adding new machine deployment types for the TB related specs

moadqassem:
Support tinkerbell as a baremetal driver.
Fixes #932

This PR only introduces tinkerbell as a baremetal driver in addition to provision servers using tinkerbell. Next step would be injecting the cloud configuration to the provisioned machines which have all the needed configs to join the user cluster.
**Optional Release Note**:

```release-note
Provisioning baremetal clusters
```


pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/hardware.go
```diff
@@ -0,0 +1,66 @@
...
+
+	return ""
+}
+
+func (h *HardwareSpec) GetStatus() string {
+	metadata := struct {
+		State string `json:"state"`
+	}{}
+
+	if err := json.Unmarshal([]byte(h.Hardware.Metadata), &metadata); err != nil {
```
mfranczy, line 60: `json.Unmarshal` is quite expensive operation, maybe would be better to unmarshal the whole hardware including metadata in the GetServer function where you already execute unmarshall operation. I am not sure about that to be honest as I understand why you decided to do it like that since the way how tinkerbell folks expose metadata (string) is not the best.

mfranczy: Maybe let's leave it for now and see how their API evolves.

pkg/cloudprovider/provider/baremetal/provider.go
```diff
@@ -126,36 +147,54 @@ func (p provider) Validate(spec v1alpha1.MachineSpec) error {
...
 		return nil, cloudprovidererrors.TerminalError{
 			Reason:  common.InvalidConfigurationMachineError,
 			Message: fmt.Sprintf("Failed to parse MachineSpec, due to %v", err),
 		}
 	}
 
-	return &bareMetalServer{}, nil
+	server, err := c.driver.GetServer(context.Background(), machine.UID, c.driverSpec)
+	if err != nil {
+		if err == cloudprovidererrors.ErrInstanceNotFound {
```
mfranczy, line 161: I think worth to change to `cloudprovidererrors.IsNotFound(err)`

mfranczy (commented):
Looks good to me, I've added a few minor things

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/workflow.go
```diff
@@ -0,0 +1,220 @@
...
+// NewWorkflowClient returns a Workflow client.
+func NewWorkflowClient(client workflow.WorkflowServiceClient, hClient *Hardware) *Workflow {
+	return &Workflow{client: client, hardwareClient: hClient}
+}
+
+// Get returns a Tinkerbell Workflow.
+func (t *Workflow) Get(ctx context.Context, id string) (*workflow.Workflow, error) {
+	tinkWorkflow, err := t.client.GetWorkflow(ctx, &workflow.GetRequest{Id: id})
+	if err != nil {
+		if err.Error() == sqlErrorString || err.Error() == sqlErrorStringAlt {
```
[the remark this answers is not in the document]
moadqassem: This part is ported from the cluster api client. This should have been vendored, but it wasn't possible import the cluster api lib in a clean way thus, I ported it as it is. I would say lets not touch that for now and once they release a tag we will import it directly.

mfranczy (approved):
so as agreed with @moadqassem I approve it as it is and we will improve implementation in the next PRs as we have to test stuff at ND.


/approve
/lgtm

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/workflow.go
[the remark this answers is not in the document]
mfranczy: then that should not be in the cluster api too, let's fix this in the following PRs.

PR #941, opened 2021-04-07
Support cloud-init bootstrap generating

moadqassem:
Generate cloud-init bootstrap, to fetch the provisioning cloud-init over the api server. There are couple of issues regarding injecting the original cloud-init configs, which configure and install all the needed libs and bins to provision a worker node. One of those issues is the cloud-init size, as some cloud providers limit the size. 

Fixes #

At the moment, this utility is not used, however in a following up PR, it would be used by the baremetal provisioner.
**Optional Release Note**:

```release-note
Fetch the provisioning cloud-init over the api-server 
```


pkg/cloudprovider/util/cloud_init_settings.go
```diff
@@ -0,0 +1,149 @@
...
+    [Unit]
+    Requires=network-online.target
+    After=network-online.target
+    [Service]
+    Type=oneshot
+    RemainAfterExit=true
+    ExecStart=/opt/bin/bootstrap
+runcmd:
+- systemctl restart bootstrap.service
+- systemctl daemon-reload
```
xmudrii, line 71: Shouldn't we first reload the daemon and then restart the service?

moadqassem: Not really, the bootstrap.service will be reloaded because it is loaded during the boot process.

pkg/cloudprovider/util/cloud_init_settings.go
```diff
@@ -0,0 +1,149 @@
...
+	"html/template"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/clientcmd"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	CloudInitNamespace = "cloud-init-settings"
```
xmudrii, line 34: Nit: the namespace may be too general. Maybe we should call it `machine-controller` or `machine-controller-cloud-init`.

xmudrii (commented):
Left some nits, otherwise LGTM.

pkg/cloudprovider/util/cloud_init_settings.go
```diff
@@ -0,0 +1,149 @@
...
+	jwtTokenNamePrefix = "cloud-init-getter-token"
+
+	bootstrapCloudConfig = `#cloud-config
```
moelsayed, line 37: Unless you are planning to create a secret per node, I recommend that you set the node hostname using this cloud-init instead of the provisioning one.

pkg/cloudprovider/util/cloud_init_settings.go
```diff
@@ -0,0 +1,149 @@
...
+
+write_files:
+- path: "/opt/bin/bootstrap"
+  permissions: "0755"
+  content: |
+    #!/bin/bash
+    set -xeuo pipefail
+    apt-get update
+    apt-get install jq -y
+    wget --no-check-certificate --quiet \
```
moelsayed, line 47: We should probably retry here.

moelsayed: Also, some recommendations:
- You can include the cluster cacert in this cloud-cloud init and use it instead of `--no-check-certificates`. You will need it later anyway
- You can either use `curl` or use `wget -O -` to right to stdout directly. and avoid the temp file.
- curl/wget might exist on the node based on the image. So might be a good idea to apt install it anyway.

pkg/cloudprovider/util/cloud_init_settings.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+func ExtractCloudInitSettingsToken(ctx context.Context, client ctrlruntimeclient.Client) (string, error) {
+	secretList := corev1.SecretList{}
+	if err := client.List(ctx, &secretList, &ctrlruntimeclient.ListOptions{Namespace: CloudInitNamespace}); err != nil {
+		return "", fmt.Errorf("failed to list secrets in namespace %s: %v", CloudInitNamespace, err)
+	}
+
+	for _, secret := range secretList.Items {
+		if strings.HasPrefix(secret.Name, jwtTokenNamePrefix) {
```
moelsayed, line 82: Are we sure this is exactly one of those secrets?

moelsayed (commented):
Just a few suggestions. Looks good :+1:

pkg/cloudprovider/util/cloud_init_settings.go
moadqassem: hmm I am not sure, as this is a user cluster and there is no other place where we get the cloud init from. So not quite sure if this is really general :)

pkg/cloudprovider/util/cloud_init_settings.go
moadqassem: why? I won't to keep the script as clean as possible and also hostnames might differ for each cloud provider, which means adding conditions. The moment you start adding conditions, this means this is slim anymore.

pkg/cloudprovider/util/cloud_init_settings.go
moadqassem: I don't really understand the question, What do you mean ?

pkg/cloudprovider/util/cloud_init_settings.go
moelsayed: I agree with that in principle. However, in practice, it's just AWS that has a different behavior. The alternatives would be:
- Create a cloud-init-secret per node which is not really scalable.
- Write some magic that would be executed by the cloud-init-secret to figure out and apply the hostname. This will be tricky because you will need to figure out the hostname for kubelet configuration. It has to be "known".

xmudrii:
/lgtm cancel
@moadqassem The `yamllint` failure seems legitimate, PTAL.

xmudrii (approved):
/lgtm

PR #948, opened 2021-04-20
Support Nautobot client in machine controller

moadqassem:
Support Nautobot minimal client which will be used to provision baremetal clusters
 
Fixes #

The client is not being used in the baremetal provider. will be added once tested in a separate commit 
**Optional Release Note**:

```release-note
Nautobot client support
```

### DIFF
--- a/pkg/nautobot/client.go
+++ b/pkg/nautobot/client.go
@@ -23,10 +23,20 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 )
 
-type deviceClient struct {
+const httpScheme = "http"
+
+type Client interface {
+	RequestActiveDevice() (*NetworkDevice, error)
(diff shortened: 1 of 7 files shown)


mfranczy:
In my opinion, the entire nautobot client should be in a different repository and used in the machine controller.

moadqassem:
> In my opinion, the entire nautobot client should be in a different repository and used in the machine controller.

Well, this is what I have mentioned earlier, that this client shouldn't be part of the machine-controller because it can be bounded directly to specific DC setup. For now, there is no other place that suits this client however in the future we can find another place or even a service that should contain this client

mfranczy:
> For now, there is no other place that suits this client however in the future we can find another place or even a service that should contain this client

I meant a dedicated repositroy only for nautobot client

mfranczy:
I still think that should be a separate client, although others thing in a different way, evenutally we might move this client once we have other consuments.

/lgtm

PR #949, opened 2021-04-22
Integrate nautobot client in baremetal provider

moadqassem:
Integrate Nautobot client in the baremetal provider so that we can use it as a metadata server.

Fixes #

**Optional Release Note**:

```release-note
Baremetal cluster provisioning 
```

### DIFF
--- a/pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/common.go
+++ b/pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/common.go
@@ -28,4 +28,5 @@ var ErrNotFound = errors.New("resource not found")
 const (
 	sqlErrorString    = "rpc error: code = Unknown desc = sql: no rows in result set"
 	sqlErrorStringAlt = "rpc error: code = Unknown desc = SELECT: sql: no rows in result set"
+	sqlErrorNotFound  = "rpc error: code = NotFound desc = not found"
 )
--- a/pkg/nautobot/device.go
+++ b/pkg/nautobot/device.go
@@ -21,155 +21,159 @@ import "time"
 type DeviceStatus string
 
 const (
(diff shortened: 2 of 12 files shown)


pkg/nautobot/device.go
```diff
@@ -21,155 +21,159 @@ import "time"
...
-	Planned DeviceStatus = "planned"
-	Staged  DeviceStatus = "staged"
+	Active DeviceStatus = "active"
+	Staged DeviceStatus = "staged"
 )
 
 type NetworkDevice struct {
-	Count   int          `json:"count"`
-	Results []DeviceInfo `json:"results"`
+	Count   int           `json:"count,omitempty,omitempty"`
```
mfranczy, line 29: repeated `omitempty`

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/common.go
```diff
@@ -28,4 +28,5 @@ var ErrNotFound = errors.New("resource not found")
 const (
 	sqlErrorString    = "rpc error: code = Unknown desc = sql: no rows in result set"
 	sqlErrorStringAlt = "rpc error: code = Unknown desc = SELECT: sql: no rows in result set"
+	sqlErrorNotFound  = "rpc error: code = NotFound desc = not found"
```
mfranczy, line 31: maybe worth to add a comment why you decided to check errors in that way, as you described here https://github.com/kubermatic/machine-controller/pull/930#discussion_r604979119

mfranczy:
for the ND PoC it's ok, although, we will have to revisit the whole bare metal implementation once we work on the production solution, some assumptions might not work (like taking the first active interface). Just posting this here that others are aware why we decided to proceed in that way.

PR #950, opened 2021-04-26
Fix RHEL subscription deregistration on OpenStack

chrkl:
The deregistration of a RHEL subscription failed if the machines FQDN
did not match its node name. This happened for instance on OpenStack
clusters that add the .novalocal domain suffix.

This change sets the fqdn field in the cloud-init userdata to the hostname to
ensure that the machine name matches the registered name in the
RHSM or Satellite server.

Fixes #

**Optional Release Note**:

```release-note
Fix RHEL subscription deregistration on OpenStack
```

### DIFF
--- a/pkg/userdata/rhel/provider.go
+++ b/pkg/userdata/rhel/provider.go
@@ -132,6 +132,7 @@ func (p Provider) UserData(req plugin.UserDataRequest) (string, error) {
 const userDataTemplate = `#cloud-config
 {{ if ne .CloudProviderName "aws" }}
 hostname: {{ .MachineSpec.Name }}
+fqdn: {{ .MachineSpec.Name }}
 {{- /* Never set the hostname on AWS nodes. Kubernetes(kube-proxy) requires the hostname to be the private dns name */}}
 {{ end }}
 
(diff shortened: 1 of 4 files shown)


pkg/userdata/rhel/provider.go
```diff
@@ -132,6 +132,7 @@ func (p Provider) UserData(req plugin.UserDataRequest) (string, error) {
 const userDataTemplate = `#cloud-config
 {{ if ne .CloudProviderName "aws" }}
 hostname: {{ .MachineSpec.Name }}
+fqdn: {{ .MachineSpec.Name }}
```
moadqassem, line 135: This means it is included on all cloud providers by aws. If you want to set it on OS only I would suggest that you add a new case for it:
```go
{{ if .CloudProviderName "openstack" }}
fqdn: {{ .MachineSpec.Name }}
{{ end }}
```

moadqassem: Have you actually check if the subscription has been removed after this change?

chrkl: Yes, works for OpenStack / OTC with a Satellite server.

PR #971, opened 2021-05-12
introduce insecure option for openstack cloud provider

### LINKED ISSUE #970: Consider implementing TLS Insecure Option for OpenStack Cloud Provider
Right now, there is no way to configure the `tls-inseucre` option for the OpenStack cloud Provider:

```go
type Config struct {
	IdentityEndpoint            string
	ApplicationCredentialID     string
	ApplicationCredentialSecret string
	Username                    string
	Password                    string
	DomainName                  string
	TenantName                  string
	TenantID                    string
	TokenID                     string
	Region                      string
[...]
}
```
meanwhile the OpenStack Terraform Provider does implement that option, making use of the [utils/auth](https://github.com/gophercloud/utils/blob/master/terraform/auth/config.go) package:

```go
v, ok := d.GetOkExists("insecure")
	if ok {
		insecure := v.(bool)
		config.Insecure = &insecure
	}
```
https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/cc49d3294434f9fce84fe9fc26e2917854b26b18/openstack/provider.go#L523

Would you mind implementing that option as well, otherwise I dont think I can fast bypass the self signed ssl certificate in our dev openstack environment.

FalcoSuessgott:

Fixes #970

**Optional Release Note**:

```release-note
Introduces the `insecure` option which can be set via `OS_INSECURE` or `insecure=true` in the cloud provider config
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -23,7 +23,8 @@ import (
 	"strings"
 	"sync"
 	"time"
-
+	"crypto/tls"
+	"net/http"
 	"github.com/gophercloud/gophercloud"
 	goopenstack "github.com/gophercloud/gophercloud/openstack"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/bootfromvolume"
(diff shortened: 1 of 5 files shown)


FalcoSuessgott:
Okay I just realized that we have to use a subet of available cloud.conf options.... https://v1-18.docs.kubernetes.io/docs/concepts/cluster-administration/cloud-providers/#openstack and cannot just implement an insecure option on our own.

Sorry for any inconvenience, but I learned something at least ..

PR #975, opened 2021-05-18
use the rhel ga instead of custom image on azure

### LINKED ISSUE #974: Fix RHEL image plan for e2e tests
**User Story**
As a user I would like to run all the required e2e tests in the machine controller, however at the moment, the azure rhel image is broken due to Azure plans. 

**Acceptance Criteria**
- Azure e2e tests are running without any issues 
- Fix the rhel image plan in Azure

moadqassem:
Using RHEL GA images instead of custom images. 

Fixes #974

**Optional Release Note**:

```release-note
None
```

### DIFF
--- a/pkg/userdata/rhel/provider.go
+++ b/pkg/userdata/rhel/provider.go
@@ -204,7 +204,9 @@ write_files:
 {{- /*  https://bugs.launchpad.net/cloud-init/+bug/1662542 */}}
     hostnamectl set-hostname {{ .MachineSpec.Name }}
     {{ end }}
-
+    {{ if eq .CloudProviderName "azure" }}
+    yum update -y --disablerepo='*' --enablerepo='*microsoft*'
+    {{ end }}
     yum install -y \
       device-mapper-persistent-data \
       lvm2 \
--- a/.prow.yaml
+++ b/.prow.yaml
@@ -313,7 +313,7 @@ presubmits:
               cpu: 500m
 
   - name: pull-machine-controller-e2e-aws-arm
-    always_run: true
+    always_run: false
     decorate: true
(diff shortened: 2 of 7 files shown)


pkg/userdata/rhel/provider.go
```diff
@@ -204,7 +204,9 @@ write_files:
 {{- /*  https://bugs.launchpad.net/cloud-init/+bug/1662542 */}}
     hostnamectl set-hostname {{ .MachineSpec.Name }}
     {{ end }}
-
+    {{ if eq .CloudProviderName "azure" }}
+    yum update -y --disablerepo='*' --enablerepo='*microsoft*'
```
kron4eg, line 208: Isn't this should be gated under `{{- if .OSConfig.DistUpgradeOnBoot }}`?

kron4eg: I mean, in addition to `if eq .CloudProviderName "azure"`

moadqassem: Why? the reason why I disabled those repos due to some openssl certificates that are not consistent by some mirrors. After upgrading to use the Gold Access images for azure and then enabled repos from MS mirrors all worked.

kron4eg: Since we use GA image, does it mean this line can be removed then?

moadqassem: No, because either cases whether you have a GA image or a custom one, rhui client would still try to access the repos. Thus this will end up with the same situation where rhui client certificates are expired and it would fail on all mirrors. I was expecting that updating `yum update` alone would actually update the certificate as well, however this is not the case for RHEL 8.x

PR #977, opened 2021-05-19
add aws-node-termination-handler label for node selector

moadqassem:
Label machines and nodes with aws-node-termination-handler for aws spot instances draining.

Fixes kubermatic/kubermatic#6723

Please have a look on this PR as well:
https://github.com/kubermatic/kubermatic/compare/master...moadqassem:supports-aws-node-termination-handler?expand=1

**Optional Release Note**:
```release-note
AWS Spot instances support
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -389,6 +389,12 @@ func (p *provider) AddDefaults(spec v1alpha1.MachineSpec) (v1alpha1.MachineSpec,
 	if rawConfig.AssignPublicIP == nil {
 		rawConfig.AssignPublicIP = aws.Bool(true)
 	}
+	if rawConfig.IsSpotInstance != nil && *rawConfig.IsSpotInstance {
+		if spec.Labels == nil {
+			spec.Labels = map[string]string{}
+		}
+		spec.Labels["k8c.io/aws-spot"] = "aws-node-termination-handler"
+	}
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -389,6 +389,12 @@ func (p *provider) AddDefaults(spec v1alpha1.MachineSpec) (v1alpha1.MachineSpec,
 	if rawConfig.AssignPublicIP == nil {
 		rawConfig.AssignPublicIP = aws.Bool(true)
 	}
+	if rawConfig.IsSpotInstance != nil && *rawConfig.IsSpotInstance {
+		if spec.Labels == nil {
+			spec.Labels = map[string]string{}
+		}
+		spec.Labels["k8s-app"] = "aws-node-termination-handler"
```
kron4eg, line 396: probably it's better to use something with OUR namespace (`k8c.io/aws-spot` for example, or anything else under our namespace).

moadqassem: The reason why actually added this, is due to the addon template. The addon had the same labels from upstream. If I change here then I need to change it also in KKP and keep propagate this change with each time we upgrade the addon

kron4eg: Yeah, the label IS needed no doubt, it's just the labels itself that should correspond to what happening.

kron4eg (approved):

PR #981, opened 2021-05-21
Add ssh config to cloud-init template - POC

emmanuel-kubermatic:
Implements a POC for the ssh signed certificates implementation with vault

Fixes #6671

**Optional Release Note**:

```release-note

```

### DIFF
--- a/pkg/apis/plugin/plugin.go
+++ b/pkg/apis/plugin/plugin.go
@@ -54,6 +54,7 @@ type UserDataRequest struct {
 	KubeletRepository     string
 	KubeletFeatureGates   map[string]bool
 	ContainerRuntime      containerruntime.Config
+	UseTrustedCA          bool
 }
 
 // UserDataResponse contains the responded user data.
--- a/pkg/userdata/ubuntu/provider.go
+++ b/pkg/userdata/ubuntu/provider.go
@@ -157,6 +157,14 @@ write_files:
 {{ proxyEnvironment .HTTPProxy .NoProxy | indent 4 }}
 {{- end }}
 
+{{- if .UseTrustedCA }}
+- path: "/etc/ssh/sshd_config"
+  content: |
+	TrustedUserCAKeys /etc/ssh/trusted-user-ca-keys.pem
+	CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
+  append: true
+{{- end }}
+
 - path: "/etc/systemd/journald.conf.d/max_disk_use.conf"
   content: |
 {{ journalDConfig | indent 4 }}


pkg/apis/plugin/plugin.go
```diff
@@ -54,6 +54,7 @@ type UserDataRequest struct {
 	KubeletRepository     string
 	KubeletFeatureGates   map[string]bool
 	ContainerRuntime      containerruntime.Config
+	UseTrustedCA          bool
```
kron4eg, line 57: Seems like this field is not used anymore

pkg/userdata/amzn2/provider.go
```diff
@@ -140,6 +140,20 @@ package_upgrade: true
 package_reboot_if_required: true
 {{- end }}
 
+{{- if .ProviderSpec.CAPublicKey }}
+- path: "/etc/ssh/sshd_config"
+  content: |
+	TrustedUserCAKeys /etc/ssh/trusted-user-ca-keys.pem
+	CASignatureAlgorithms ecdsa-sha2-nistp256,ecdsa-sha2-nistp384,ecdsa-sha2-nistp521,ssh-ed25519,rsa-sha2-512,rsa-sha2-256,ssh-rsa
```
kron4eg, line 147: Please move this into a template function and call it in all other places

pkg/userdata/amzn2/provider.go
```diff
@@ -140,6 +140,20 @@ package_upgrade: true
...
+  append: true
+{{- end }}
+
+{{- if .ProviderSpec.CAPublicKey }}
+- path: "/etc/ssh/trusted-user-ca-keys.pem"
+  content: |
+{{ .ProviderSpec.CAPublicKey | indent 4 }}
+{{- end }}
```
kron4eg, line 155:
```suggestion
{{- with .ProviderSpec.CAPublicKey }}
- path: "/etc/ssh/trusted-user-ca-keys.pem"
  content: |
{{ . | indent 4 }}
{{- end }}
```

pkg/apis/plugin/plugin.go
emmanuel-kubermatic: Yes, I'd remove that

pkg/userdata/amzn2/provider.go
emmanuel-kubermatic: Cool cool

pkg/userdata/amzn2/provider.go
```diff
@@ -140,6 +140,20 @@ package_upgrade: true
...
+
```
kron4eg, line 156:
```suggestion
{{- with .ProviderSpec.CAPublicKey }}
- path: "/etc/ssh/sshd_config"
  content: |
	{{ SshdConfigAddendum }} << this should be defined in FuncMap
  append: true

- path: "/etc/ssh/trusted-user-ca-keys.pem"
  content: |
{{ . | indent 4 }}
{{- end }}
```

emmanuel-kubermatic: Done

kron4eg: I meant add new function to the https://github.com/kubermatic/machine-controller/blob/b51933b55d3474a6f26014e56bee985a78aa0f13/pkg/userdata/helper/template_functions.go#L27-L46

emmanuel-kubermatic: I've done that now

I reverted the change and unresolved this conversation

kron4eg (approved):

pkg/userdata/amzn2/provider.go
```diff
@@ -300,6 +300,18 @@ write_files:
   content: |
 {{ kubeletHealthCheckSystemdUnit | indent 4 }}
 
+{{- with .ProviderSpec.CAPublicKey }}
+
```
irozzo-1A, line 304: Is this newline intentional?

irozzo-1A: I saw the rendered version and looks good, probably you could have skipped the dash and the newline:

```
{{ with .ProviderSpec.CAPublicKey }}
- path: "/etc/ssh/trusted-user-ca-keys.pem"
```
but it's ok.

irozzo-1A:
Overall LGTM, I would add a e2e test, but you could add it with a follow-up PR if you prefer.

pkg/userdata/amzn2/provider.go
emmanuel-kubermatic: Hello

It was intentional as the next block would fall on the next line with this so the spacing is to accomodate for legibility

```yaml
- path:dukfl;d
  key: value and some other content
- path: huf
```

instead of 
```yaml
- path:dukfl;d
  key: value and some other content
  
- path: huf
```

Nothing more

emmanuel-kubermatic:
> Overall LGTM, I would add an e2e test, but you could add it with a follow-up PR if you prefer.

I would definitely do this on another MR honestly

I added some unit tests already for the `CAPublicKey` addition but an E2E test works

PR #983, opened 2021-05-25
Hetzner firewall support

### LINKED ISSUE #982: Support Hetzner Firewalls
Using terraform + kubeone. If using Kubermatic machine-controller to manage worker nodes there is currently no way to assign a firewall to the worker nodes.

Desired outcome.

- If a firewall id is found in the `kubeone_workers` object...
- Run `kubeone apply --manifest kubeone.yaml --tfjson output.json` to create the worker nodes with the firewall assignment.
- Scaling with `kubectl scale -n kube-system machinedeployment development-pool1 --replicas=x` to manage firewall assignment.

I'm on kubernetes slack #kubeone incase the issue in not clear there is a discussion there. Also, with another user who has the same issue.

EarthlingDavey:
Support for firewalls on Hetzner. 

Without this feature there is no way to assign firewalls to Hetzner servers when using machine deployment.

Fixes #982

This is untested. @kron4eg asked for me to create the PR and testing will be done by kubermatic

**Optional Release Note**:

```release-note
This is the first go code that I've written. Please review the code with that in mind. Thanks.
```


test/e2e/provisioning/testdata/machinedeployment-hetzner.yaml
```diff
@@ -30,6 +30,8 @@ spec:
             location: "nbg1"
             networks:
               - "machine-controller-e2e"
+            firewall:
+              - "machine-controller-e2e"
```
xmudrii, line 34: This might fail our tests because I don't think we have created this firewall yet.

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -183,6 +192,14 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 		}
 	}
 
+	if len(c.Firewalls) != 0 {
+		for _, firewall := range c.Firewalls {
+			if _, _, err = client.Firewall.Get(ctx, firewall); err != nil {
+				return fmt.Errorf("failed to get firewall %q: %v", firewall, err)
+			}
```
kron4eg, line 199:
```suggestion
			if frwl, _, err := client.Firewall.Get(ctx, firewall); err != nil || frwl == nil {
				return fmt.Errorf("failed to get firewall %q: %v", firewall, err)
			}
```

kron4eg: this is how hcloud handles the 404: https://github.com/hetznercloud/cli/blob/bb30002002cd2c8af6b20269eff549d09f7204a5/internal/cmd/firewall/describe.go#L36-L42

404 of object doesn’t return an error but rather a nil object and we have to check for nilness

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -260,6 +277,20 @@ func (p *provider) Create(machine *v1alpha1.Machine, _ *cloudprovidertypes.Provi
 		}
 	}
 
+	if len(c.Firewalls) != 0 {
+		serverCreateOpts.Firewalls = []*hcloud.ServerCreateFirewall{}
```
kron4eg, line 281: nit, this is not necessary, first `append()` will initialize it.

pkg/cloudprovider/provider/hetzner/provider.go
EarthlingDavey: I'll change now. To pass spell checks, do you mind if I use fetchedFirewall instead of frwl?

pkg/cloudprovider/provider/hetzner/provider.go
EarthlingDavey: Deleted this line

pkg/cloudprovider/provider/hetzner/provider.go
kron4eg: sure, I used frwl just as a demo that there has to be something

EarthlingDavey: OK cool.

test/e2e/provisioning/testdata/machinedeployment-hetzner.yaml
```diff
@@ -30,6 +30,8 @@ spec:
             location: "nbg1"
             networks:
               - "machine-controller-e2e"
+            firewall:
```
xmudrii, line 33: Typo:

```suggestion
            firewalls:
```

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -183,6 +192,14 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 		}
 	}
 
+	if len(c.Firewalls) != 0 {
+		for _, firewall := range c.Firewalls {
+			if n, _, err := client.Firewall.Get(ctx, firewall); err != nil || n == nil {
+				return fmt.Errorf("failed to get firewall %q: %v", firewall, err)
+			}
```
xmudrii, line 199: Let split this check into two `if`s. Right now, if the firewall doesn't exist, this is going to produce an error such as:

```
failed to get firewall "machine-controller-e2e": <nil>
```

From this error, it's hard to find out that the firewall doesn't exist. Instead, we can do this:

```suggestion
			if f, _, err := client.Firewall.Get(ctx, firewall); err != nil {
				return fmt.Errorf("failed to get firewall %q: %v", firewall, err)
			}
			if f == nil {
				return fmt.Errorf("firewall %q does not exist", firewall)
			}
```

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -183,6 +192,17 @@ func (p *provider) Validate(spec v1alpha1.MachineSpec) error {
 		}
 	}
 
+	if len(c.Firewalls) != 0 {
+		for _, firewall := range c.Firewalls {
+			if f, _, err := client.Firewall.Get(ctx, firewall); err != nil {
```
xmudrii, line 197: Sorry, my previous suggestion was not fully correct. This should work.

```suggestion
			f, _, err := client.Firewall.Get(ctx, firewall)
			if err != nil {
```

xmudrii (approved):
The Hetzner tests are green. I've verified that the firewall is attached to the instances.
@EarthlingDavey Thank you for the contribution!

/lgtm
/approve

EarthlingDavey:
> The Hetzner tests are green. I've verified that the firewall is attached to the instances.
> @EarthlingDavey Thank you for the contribution!

Awesome, you're welcome. It's very nice to have made a tiny contribution to the project. 
And, as a bonus, I learned a little go and gained a feel for how the project works.

Cheers @xmudrii & @kron4eg - your guidance & patience was really appreciated.

PR #995, opened 2021-06-09
Fixed openstack client ca bundle handling

### LINKED ISSUE #994: Openstack provider client initialization authenticates before setting the custom CA Bundle
In the Openstack provider, we try to set the custom CA bundle by creating the openstack client and then overriding its default root/host CA bundle with our one.

```
pc, err := goopenstack.AuthenticatedClient(opts)
if pc != nil {
	// use the util's HTTP client to benefit, among other things, from its CA bundle
	pc.HTTPClient = cloudproviderutil.HTTPClientConfig{LogPrefix: "[OpenStack API]"}.New()
}

return pc, err
```
But the issue is that we are calling goopenstack.AuthenticatedClient(opts) which both creates the client and authenticates it, so the custom CA bundle is not used during authentication which leads to errors.

lsviben:
Fix for openstack custom ca bundle handling. Client was trying to authenticate before setting the custom CA bundle which was causing errors.

Fixes #994 

**Optional Release Note**:

```release-note
Fixed using a custom CA Bundle for Openstack by authenticating after setting the custom CA bundle
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -308,12 +308,16 @@ func getClient(c *Config) (*gophercloud.ProviderClient, error) {
 		ApplicationCredentialSecret: c.ApplicationCredentialSecret,
 	}
 
-	pc, err := goopenstack.AuthenticatedClient(opts)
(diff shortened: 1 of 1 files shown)


mfranczy (approved):
/lgtm
/approve

PR #1011, opened 2021-06-30
Dynamically choose which certificates.k8s.io version to support

kron4eg:
We had to update certificates API, now I'm returning the support for older certificates.k8s.io/v1beta1, but NOW it can be chosen based on API availability.

**Optional Release Note**:

```release-note
Dynamically choose which certificates.k8s.io version to support
```

### DIFF
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -44,7 +44,6 @@ import (
 	"github.com/kubermatic/machine-controller/pkg/node"
 	"github.com/kubermatic/machine-controller/pkg/signals"
 
-	certificatesv1 "k8s.io/api/certificates/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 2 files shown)


pkg/controller/nodecsrapprover/node_csr_approver.go
```diff
@@ -59,19 +61,72 @@ type reconciler struct {
...
 	if err != nil {
-		return fmt.Errorf("failed to create certificate client: %v", err)
+		return fmt.Errorf("failed to init discovery client: %w", err)
 	}
 
-	r := &reconciler{Client: mgr.GetClient(), certClient: certClient.CertificateSigningRequests()}
-	c, err := controller.New(ControllerName, mgr, controller.Options{Reconciler: r})
+	srvGroups, err := discoveryClient.ServerGroups()
+	if err != nil {
+		return fmt.Errorf("failed to server API groups: %w", err)
```
xmudrii, line 73: Tiny nit: it feels like this error message is missing a word.

```suggestion
		return fmt.Errorf("failed to get server API groups: %w", err)
```

pkg/controller/nodecsrapprover/node_csr_approver.go
```diff
@@ -59,19 +61,72 @@ type reconciler struct {
...
+
+	switch certificatesVersionFound {
+	case "v1":
+		certClient, err := certificatesv1client.NewForConfig(mgr.GetConfig())
+		if err != nil {
+			return fmt.Errorf("failed to create certificate client: %v", err)
+		}
+		rec = &reconciler{Client: mgr.GetClient(), certClient: certClient.CertificateSigningRequests()}
+		watchType = &certificatesv1.CertificateSigningRequest{}
+	case "v1beta1":
```
xmudrii, line 106: Is it expected that we don't assign `watchType` for v1beta1?

pkg/controller/nodecsrapprover/node_csr_approver.go
kron4eg: indeed :D

pkg/controller/nodecsrapprover/node_csr_approver.go
kron4eg: good catch, no, it was forgotten

xmudrii (approved):
/lgtm
/approve

PR #1026, opened 2021-07-22
Don't delete openstack instances on creation timeout

### LINKED ISSUE #1022: Openstack VM creation timeout
Currently openstack Create() call blocks (with wait.Poll) the call for `InstanceReadyCheckTimeout` (10 seconds by default) and recheck VM state every `InstanceReadyCheckPeriod` (5 seconds by default). If the timeouts are too short, this causes call to fail. And this timeout has proved to be too short in different OS installations. 

If this timeout is too long, then whole event loop of machine-controller is blocked.

kron4eg:
Fixes #1022

**Optional Release Note**:

```release-note
Don't delete openstack instances on creation timeout
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/helper.go
+++ b/pkg/cloudprovider/provider/openstack/helper.go
@@ -43,7 +43,7 @@ var (
 )
 
 const (
-	floatingIPErrorStatus = "ERROR"
+	errorStatus = "ERROR"
 
 	floatingReassignIPCheckPeriod = 3 * time.Second
 )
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -565,53 +579,50 @@ func (p *provider) Create(machine *v1alpha1.Machine, data *cloudprovidertypes.Pr
 		}
 	}
 
-	if err := p.serverReadinessWaiter(computeClient, server.ID, c.InstanceReadyCheckPeriod, c.InstanceReadyCheckTimeout); err != nil {
-		defer deleteInstanceDueToFatalLogged(computeClient, server.ID)
-		return nil, fmt.Errorf("instance %s became not active: %v", server.ID, err)
-	}
-
-	// Find a free FloatingIP or allocate a new one
 	if c.FloatingIPPool != "" {
```
kron4eg, line 582: poll and wait ONLY IN CASE if it's really needed (i.e. FloatingIPPool is set and thus floating IP is required).

pkg/cloudprovider/provider/openstack/helper.go
```diff
@@ -385,8 +363,8 @@ func getInstancePort(client *gophercloud.ProviderClient, region, instanceID, net
 	return nil, errNotFound
 }
 
-func getDefaultNetwork(client *gophercloud.ProviderClient, region string) (*osnetworks.Network, error) {
-	networks, err := getNetworks(client, region)
+func getDefaultNetwork(netClient *gophercloud.ServiceClient, region string) (*osnetworks.Network, error) {
```
moadqassem, line 366: Here the region parameter can be omitted as well

pkg/cloudprovider/provider/openstack/helper.go
```diff
@@ -418,14 +396,16 @@ NetworkLoop:
 	return nil, fmt.Errorf("%d candidate networks found", len(candidates))
 }
 
-func getDefaultSubnet(client *gophercloud.ProviderClient, network *osnetworks.Network, region string) (*string, error) {
+func getDefaultSubnet(netClient *gophercloud.ServiceClient, network *osnetworks.Network, region string) (*string, error) {
```
moadqassem, line 399: Here the region parameter can be omitted as well

kron4eg:
@moadqassem done, PTAL.

moadqassem (commented):
/approve
/lgtm

PR #1029, opened 2021-08-02
prevent AWS metrics from growing over time

xrstf:
Without resetting the metric, old label values will never get cleaned up, so the number of metrics will just grow without bounds.

Fixes #

**Optional Release Note**:

```release-note
Fix machine_controller_aws_instances_for_machine metric growing with new machines, but never forgetting old, deleted machines.
```


moadqassem (commented):
/approve
/lgtm

PR #1060, opened 2021-09-30
update to controller-runtime 0.10 and k8s 1.22, remove compat with k8s <1.19

xrstf:
This does a long overdue dependency bump for the controller-runtime. This also required bumping the `kubevirt.io/containerized-data-importer`, which in turn required a whole bunch of new replacements.

I also merged the two different semver modules into a single one, because I could not find out why we import both v2 and v3.

**Optional Release Note**:
```release-note
update Kubernetes dependencies to 1.22 and controller-runtime 0.10.1
support for machines running Kubernetes <1.19 has been removed
```


pkg/controller/machine/machine_controller.go
```diff
@@ -1217,7 +1217,7 @@ ExecStart=/opt/bin/bootstrap
 hostname: {{ .MachineSpec.Name }}
 {{- /* Never set the hostname on AWS nodes. Kubernetes(kube-proxy) requires the hostname to be the private dns name */}}
 {{ end }}
-ssh_pwauth: no
+ssh_pwauth: false
```
embik, line 1220: Why is this changing to `false`? Maybe I'm looking at the wrong documentation, but https://cloudinit.readthedocs.io/en/latest/topics/modules.html#set-passwords suggests that `no` is the correct value to set.

embik: This happens in a couple of places, but I'll leave this as one review comment to discuss this.

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -624,7 +624,7 @@ func TestAzureRedhatSatelliteProvisioningE2E(t *testing.T) {
 		name:              "Azure redhat satellite server subscription",
 		osName:            "rhel",
 		containerRuntime:  "docker",
-		kubernetesVersion: "v1.17.0",
+		kubernetesVersion: "1.21.5",
```
embik, line 627: This is getting rid of the `v` prefix in `kubernetesVersion`, is that intentional? In this file, it's sometimes there for a scenario and sometimes isn't.

embik (commented):
Two questions for changes I didn't fully understand, otherwise this looks quite good.

pkg/controller/machine/machine_controller.go
embik: After going over the code with @xrstf it looks like cloud-init is capable of understanding variations of true/false here, including yes/no, which is what is referenced in the docs for some reason. But this should be fine as well.

test/e2e/provisioning/all_e2e_test.go
xrstf: I think for the versions I changed, I removed the prefix. That was IIRC the only thought that went into it.

pkg/controller/machine/machine_controller.go
xrstf: https://github.com/number5/cloud-init/blob/main/cloudinit/config/cc_set_passwords.py#L101 shows that the Python code expects a true and https://github.com/number5/cloud-init/blob/820e3d1c1c8869aa84757bcd49b8619c17a350b1/cloudinit/util.py#L219 proves that a native bool is even faster (FASTER!!!11!!), so I think we're find. cloudinit is also translating this into the appropriate value, regardless whether it was "no" or "false".

I just got reminded the of YAML alias problem and wanted to avoid one more pitfall ^^

test/e2e/provisioning/all_e2e_test.go
embik: After doing some research, it seems that without the `v` prefix is correct. This is being templated into manifests for `MachineDeployments` [here](https://github.com/kubermatic/machine-controller/blob/a5b2d57b475da95de59368f20cfe936dac55d0a5/test/e2e/provisioning/helper.go#L143). In our examples, the kubelet version is given without the prefix.

embik (approved):
/lgtm

xrstf:
/override pull-machine-controller-e2e-openstack 

Flatcar image is missing.

vCenter is offline

PR #1067, opened 2021-10-06
Flatcar improvements

FlorinPeter:
The PR improves the Flatcar implementation to close the drift between other providers
- remove containerized kubelet
- removed docker service dependency
- use the same containerd config and kubelet.service as other providers
- allow to use the pause image as the sandbox image in containerd to gain more control over the used image

This PR was tested with Flatcar 2905.2.3 (Oklo)

**Optional Release Note**:

```release-note
Flatcar improvements
Containerd config improvement
```


FlorinPeter:
@moadqassem I rebased my Flatcar changes and closed the PR https://github.com/kubermatic/machine-controller/pull/1068 as we discussed today in favor of the upcoming OSM integration. 
Would be nice to get this PR merged to keep the gap between our fork and the upstream as minimal as possible.

pkg/userdata/flatcar/provider.go
```diff
@@ -254,66 +256,15 @@ systemd:
...
-        [Unit]
-        Description=Kubernetes Kubelet
-        Requires=docker.service
-        After=docker.service
-        [Service]
-        TimeoutStartSec=5min
-        CPUAccounting=true
-        MemoryAccounting=true
-        EnvironmentFile=-/etc/environment
-        EnvironmentFile=/etc/kubernetes/nodeip.conf
```
kron4eg, line 268: we need this file

FlorinPeter: this is done via the kubeletFlags macro.
You can see it from the expected config here https://github.com/kubermatic/machine-controller/pull/1067/files/a173b5a6aedbf5fbec5461c77ac6077f563db568#diff-1f679fb68a00fca3c52634cc0620d2dc5ee938e7e821bc6790f1233b8acf082cL99-R101

kron4eg: kubeletFlags uses the ENV that should be set by `EnvironmentFile=/etc/kubernetes/nodeip.conf`, but since it's removed, there is no `$KUBELET_NODE_IP` ENV.

FlorinPeter: @kron4eg got your point, will check tomorrow why it was left behind, thx for the finding

kron4eg (approved):

kron4eg:
```
 pkg/userdata/flatcar/provider.go:681: File is not `gofmt`-ed with `-s` (gofmt) 
```

FlorinPeter:
@kron4eg fixed the fmt and a typo on the drop-in, this branch was deployed on a test cluster without any issues

kron4eg (approved):

FlorinPeter:
@moadqassem something wrong with the PR or why did you put it on hold?

moadqassem:
@FlorinPeter atm vSphere tests are failing due to some internal issues, which means the test will never be green, and since it has LGTM it will keep running the jobs which spam vCenter with no success. We are working on so hopefully will unhold it soon.

FlorinPeter:
@moadqassem thx for the information

pkg/userdata/flatcar/provider.go
```diff
@@ -203,6 +205,17 @@ systemd:
...
 {{- end }}
 
+    - name: containerd.service
+      dropins:
+        - name: 10-custom.conf
+          contents: |
+            [Service]
+            Restart=always
+            EnvironmentFile=-/etc/environment
+            Environment=CONTAINERD_CONFIG=/etc/containerd/config.toml
```
moadqassem, line 215: Where are you getting this file from? it doesn't seem that you create it anywhere

pkg/userdata/flatcar/provider.go
```diff
@@ -203,6 +205,17 @@ systemd:
...
+            ExecStart=
+            ExecStart=/usr/bin/env PATH=${TORCX_BINDIR}:${PATH} ${TORCX_BINDIR}/containerd --config ${CONTAINERD_CONFIG}
```
moadqassem, line 217: Since the `/etc/containerd/config.toml` isn't actually created anywhere, this won't start

pkg/userdata/flatcar/provider.go
```diff
@@ -98,31 +98,35 @@ func (p Provider) UserData(req plugin.UserDataRequest) (string, error) {
...
+		UserDataRequest:                req,
+		ProviderSpec:                   pconfig,
+		FlatcarConfig:                  flatcarConfig,
+		Kubeconfig:                     kubeconfigString,
+		KubernetesCACert:               kubernetesCACert,
+		KubeletImage:                   kubeletImage,
+		KubeletVersion:                 kubeletVersion.String(),
+		NodeIPScript:                   userdatahelper.SetupNodeIPEnvScript(),
+		ExtraKubeletFlags:              crEngine.KubeletFlags(),
+		ContainerRuntimeConfigFileName: crEngine.ConfigFileName(),
```
moadqassem, line 128: It seems that this value here isn't loaded correctly

moadqassem (changes requested):
@FlorinPeter Well, this won't work, the reason behind that is, for any k8s version less than 1.22.x we actually deploy docker by default not containerd, this means the systemd services (such as kubelet) assumes that the cri running on the worker node is docker not containerd thus non of those services will start. Maybe I can open another PR which will fox this issue for you but I need to validate couple of things before doing that

moadqassem:
FYI the problems that you had with vCenter other than Flatcar has been resolved. However I am gonna remove the LGTM label since this won't work without couple of changes

FlorinPeter:
> @FlorinPeter Well, this won't work, the reason behind that is, for any k8s version less than 1.22.x we actually deploy docker by default not containerd, this means the systemd services (such as kubelet) assumes that the cri running on the worker node is docker not containerd thus non of those services will start. Maybe I can open another PR which will fix this issue for you but I need to validate couple of things before doing that 

@moadqassem not sure if I understand it correctly. I deployed this change with Kubeone v1.3 and K8s v1.20 without any issue.

moadqassem:
> > @FlorinPeter Well, this won't work, the reason behind that is, for any k8s version less than 1.22.x we actually deploy docker by default not containerd, this means the systemd services (such as kubelet) assumes that the cri running on the worker node is docker not containerd thus non of those services will start. Maybe I can open another PR which will fix this issue for you but I need to validate couple of things before doing that
> 
> @moadqassem not sure if I understand it correctly. I deployed this change with Kubeone v1.3 and K8s v1.20 without any issue.

Yeah sure and I can understand that. K1 deployed containerd as a default cri on all k8s versions unlike machine controller default options. Try change this line here from `docker` to `containerd` and see what you will get: 
https://github.com/kubermatic/machine-controller/blob/master/cmd/machine-controller/main.go#L167

FlorinPeter:
> > > @FlorinPeter Well, this won't work, the reason behind that is, for any k8s version less than 1.22.x we actually deploy docker by default not containerd, this means the systemd services (such as kubelet) assumes that the cri running on the worker node is docker not containerd thus non of those services will start. Maybe I can open another PR which will fix this issue for you but I need to validate couple of things before doing that
> 
> > 
> 
> > @moadqassem not sure if I understand it correctly. I deployed this change with Kubeone v1.3 and K8s v1.20 without any issue.
> 
> 
> 
> Yeah sure and I can understand that. K1 deploys containerd as a default cri on all k8s versions unlike machine controller default options. Try change this line here from `docker` to `containerd` and see what you will get: 
> 
> https://github.com/kubermatic/machine-controller/blob/master/cmd/machine-controller/main.go#L167

Ok now I got the issue you are referring to. Will try to see what will happen if I change it to docker to get a better understanding of the implications.

moadqassem (commented):
@FlorinPeter So I have opened a debugging PR based on your changes https://github.com/kubermatic/machine-controller/pull/1092. As you can see, this change https://github.com/kubermatic/machine-controller/pull/1092/files#diff-957f1608260802dae55acd0fed6cbd9009441b087b896d045928c0d0171e6b0fR167 would solve all the issues that you have in the flatcar tests as I suggested earlier. Please include that change in your PR it all tests will pass.

FlorinPeter:
@moadqassem but this will change the default to containerd and will break users that are using docker as default, or? 
I will try to check it tomorrow.

moadqassem:
> @moadqassem but this will change the default to containerd and will break users that are using docker as default, or?
> I will try to check it tomorrow.

How is that a breaking behaviour? In order to use containerd in existing machines, you need to rotate your cluster in the first place, or create a new one. The part where we specify what cri we are gonna use is only at the beginning of machine provisioning. Or what do you exactly mean ?

kron4eg:
I've rebased this PR and fixed issues mentioned by @moadqassem (albeit without changing flags defaults). So this PR is superseded by #1109.

Thanks again @FlorinPeter!

FlorinPeter:
@moadqassem @kron4eg thx a lot, was on sick leave last week and was not able to handle it.
I appreciate your efforts!

PR #1075, opened 2021-10-20
Use appropriate base64 flags

SimonTheLeg:
Signed-off-by: Simon Bein <simontheleg@gmail.com>

Adds detection whether gnu or bsd base64 package is available and
selects the correct flag. This is necessary as MacOs (or BSD) uses the
-b flag while gnu uses the -w flag. Unfortunately there is no flag for
line-wrapping that both of them support

```release-note
None
```


ahmedwaleedmalik (approved):
Tested it out locally(MacOs and linux) and the script works fine for me.

/lgtm
/approve

xrstf:

vCenter is currently down.

PR #1085, opened 2021-10-31
Change getImageByName to use the Image client

### LINKED ISSUE #1084: Change Openstack getImageByName to use Glance API
The current machine-controller uses `nova` API to get the image information about a specific image. 

Openstack Nova deprecated the proxy API in Nova computeAPIVersion 2.36 and recommends to use API from the image service (Glance) instead.

See:
- https://docs.openstack.org/api-ref/compute/?expanded=create-image-createimage-action-detail,list-images-with-details-detail#list-images

Quotation:
> These APIs are proxy calls to the Image service. Nova has deprecated all the proxy APIs and users should use the native APIs instead. All the Image services proxy APIs except image metadata APIs will fail with a 404 starting from microversion 2.36. The image metadata APIs will fail with a 404 starting from microversion 2.39.

Happy2C0de:
Implements the change from Nova API to Glance API to get the required image information.

Fixes #1084 

- I had to change the Mux return of the test environment in `provider_test.go`. The new implementation uses `/v2/images` to list the images. I used the official example to make things clear.
- Ideally, you test this implementation in an e2e-Test to assure that it works as expected.
- Image API v2 is the de-facto standard, I have never seen a v1 API in the community, because V2 was released at Openstack Folsom (2012). Glance API v1 was deprecated in Newton (2016). So, it's absolutely save to assume that all Openstack clouds have a V2 API exposed. Btw, the go client doesn't support V1 anyway.

**Optional Release Note**:

```release-note
Change the Openstack image client to use the Glance v2 API
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/helper.go
+++ b/pkg/cloudprovider/provider/openstack/helper.go
@@ -27,8 +27,8 @@ import (
 	goopenstack "github.com/gophercloud/gophercloud/openstack"
 	osavailabilityzones "github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
 	osflavors "github.com/gophercloud/gophercloud/openstack/compute/v2/flavors"
-	osimages "github.com/gophercloud/gophercloud/openstack/compute/v2/images"
 	osregions "github.com/gophercloud/gophercloud/openstack/identity/v3/regions"
+	osimagesv2 "github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 	osfloatingips "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
 	ossecuritygroups "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/groups"
 	osecruritygrouprules "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/security/rules"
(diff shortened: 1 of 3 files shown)


Happy2C0de:
@kron4eg Can you retest? 
The provider seems to have timed out.

Happy2C0de:
Thanks! :)

Happy2C0de:
@kron4eg Can you call retest?
The provider still has some issues to build the resources.

Happy2C0de:
Done, branch was rebased to current upstream master

Happy2C0de:
Very nice @kron4eg Thank you for pushing the tests through!

Happy2C0de:
@kron4eg 
LGTM too :)

kron4eg:
@Happy2C0de I wanted to let @moadqassem to see this too, since he is an assignee too.

Happy2C0de:
> @Happy2C0de I wanted to let @moadqassem to see this too, since he is an assignee too.

Got it, thanks @kron4eg 👍

moadqassem:
> @Happy2C0de I wanted to let @moadqassem to see this too, since he is an assignee too.

Will take a look in a second

Happy2C0de:
> > @Happy2C0de I wanted to let @moadqassem to see this too, since he is an assignee too.
> 
> Will take a look in a second

Thanks. I was just curious where we were standing here and I didn't realize that there is another assignee. :) No hurry!

moadqassem:
@Happy2C0de Those api calls, are they also compatible with microversion less than 2.36?

Happy2C0de:
@moadqassem 
The Glance API doesn't have microversions in contrast to Nova.

The change mainly queries Glance directly what is the default for years, the official Openstack CLI did this change in 2017.

So mainly, this PR changes:
`GET nova.region.example.com/images` -> `GET glance.region.example.com/v2/images`
this implementation does
`GET glance.region.example.com/v2/images` 
directly.
The proxy API was removed in 2016 but Glance V2 is there since 2012 or so. The go-client doesn't even support Glance V1 anymore, what makes sense. Glance belongs to the core services of each Openstack Cloud.
So, it's not a question of microversion but rather a question if there is a Glance V2 endpoint what is the case for all Openstack versions older than 2012/13 or so.

To see if a certain Openstack cloud works, you can call the Openstack CLI:
```
openstack image list --debug
....
GET call to image for https://glance.example.com/v2/images used request id req-$UUID
```

moadqassem (approved):
/approve 
/lgtm

PR #1086, opened 2021-11-01
Disable e2e tests for vSphere cloud provider temporarily

moadqassem:
At the moment, the current vSphere image that we have is not behaving as expected, die to some issue with the our subscription. Thus we are disabling the tests on vSphere until we get a valid subscription shortly.   
Fixes #

**Optional Release Note**:

```release-note
None
```

### DIFF
--- a/test/e2e/provisioning/all_e2e_test.go
+++ b/test/e2e/provisioning/all_e2e_test.go
@@ -766,7 +766,7 @@ func getVSphereTestParams(t *testing.T) []string {
 func TestVsphereProvisioningE2E(t *testing.T) {
 	t.Parallel()
 
-	selector := Not(OsSelector("sles", "amzn2"))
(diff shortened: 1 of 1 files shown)


test/e2e/provisioning/all_e2e_test.go
```diff
@@ -777,7 +777,7 @@ func TestVsphereProvisioningE2E(t *testing.T) {
 func TestVsphereDatastoreClusterProvisioningE2E(t *testing.T) {
 	t.Parallel()
 
-	selector := OsSelector("ubuntu", "centos")
+	selector := OsSelector("ubuntu", "centos", "rhel")
```
kron4eg, line 780: but this has enabled RHEL, instead of disabling. Is this intended?

kron4eg (approved):

PR #1087, opened 2021-11-01
Provide hostname to kubelet for AWS instances with external CCM

xmudrii:

The cloud controller manager is responsible for setting the Node object name. In the case of AWS, the Node object name is the instance private DNS name (FQDN).

When in-tree CCM is used, the CCM components integrated into Kubelet are responsible for setting the hostname. However, when running external CCM, it's solely up to the external CCM to properly name the Node object.

It appears that AWS in-tree and external CCMs have different logic for finding the instance object on AWS. In the case of the external CCM, there are two criteria:

* find by the providerID -- this doesn't work because the external CCM should set it, and it can't set it if it can't find the instance
* search instances by private DNS name while using the Node object name as the criteria -- this doesn't work because Node objects are not named properly (e.g. just `ip-172-31-38-11` instead of `ip-172-31-38-11.eu-west-3.compute.internal`)

Because of this, the external CCM doesn't work, which manifests as Node objects not named properly, having the uninitialized taint, and so on...

As a way to fix this, we can provide the hostname to Kubelet, which will be used to name the Node object. With this fix, the external CCM can find the instance, assign it all the metadata, and remove the taint at the end.

I decided to re-use the NodeIP script for this fix, but if there are any objections, I can put this in some other place.

**Optional Release Note**:
```release-note
Provide hostname to kubelet for AWS instances with external CCM
```

/assign @moadqassem @kron4eg 
cc: @embik @WeirdMachine


moadqassem (approved):
/approve
/lgtm

PR #1088, opened 2021-11-01
Enable e2e tests for vSphere cloud provider

moadqassem:
Adjusting the RHEL image to run without registering the nodes which led to faster provisioning and also fixed the issue with ubuntu replicas using the same parent IP address.   
Fixes #

**Optional Release Note**:

```release-note
None
```


xrstf (approved):
/approve

PR #1089, opened 2021-11-02
Use CentOS 7 images issued by CentOS CPE team

### LINKED ISSUE #1069: AMI filter for CentOS 7 uses legacy marketplace
As per https://wiki.centos.org/Cloud/AWS#Official_and_current_CentOS_Public_Images, the product code that is used by machine-controller to determine the CentOS image to use in AWS is considered legacy. We should update to the new marketplace product code provided on that page to get the officially latest CentOS 7 images when looking for an AMI.

Unfortunately, both the old and the new marketplace versions don't seem to provide a ARM64 AMI for CentOS 7. The page above mentions images published from Owner ID `125523088429` by CentOS directly, so we might need to use those instead of the marketplace version. According to the CentOS documentation, the marketplace version is not changed from what is published directly.

embik:
The current image filter for AMIs in AWS targets a legacy account for CentOS 7 images. CentOS (via their Community Platform Engineering team) started publishing their own images as per https://wiki.centos.org/Cloud/AWS. We should use those images instead.

The new location also has an ARM64 image.

Fixes #1069

**Optional Release Note**:

```release-note
Use new source of CentOS 7 images (issued by CentOS CPE) in AWS when no AMI is set
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -96,14 +96,18 @@ var (
 	)
 
 	amiFilters = map[providerconfigtypes.OperatingSystem]map[awstypes.CPUArchitecture]amiFilter{
+		// Source: https://wiki.centos.org/Cloud/AWS
 		providerconfigtypes.OperatingSystemCentOS: {
 			awstypes.CPUArchitectureX86_64: {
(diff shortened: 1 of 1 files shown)


kron4eg (approved):
/approve

PR #1093, opened 2021-11-08
fix migrate uid test

moadqassem:
Fix a bug in the migrate machine uid tests. We use a fake client in order to run the migarate uid test. In the current code base, we populate the fake client with a machine object that has a name and a namespace value. There are some modifiers functions that we use, in order to update machine cr(e.g: adding finalizers). Since the machine cr that we populate the client with, only has a name and namespace, the cloud provider would simply fail to parse the cloud provider spec(since it nil) and thus the test will fail. 

This PR change this by populating the full object in the client, thus once the modifiers functions execute an update it will have the full copy of the machine not just the machine name and namespace. 
Fixes #

**Optional Release Note**:

```release-note
None
```


xrstf (approved):
/approve

PR #1098, opened 2021-11-10
change flatcar image

moadqassem:
Currently we are using the auto-updated image for flatcar that we have in Openstack. Problem is, this image is changing all consistently due too upgrading from one release to another, thus we need to have our own custom image. The old image that we used, didn't support containerd cri thus we need to find an alternative. 
 
Fixes #

**Optional Release Note**:

```release-note
None
```


embik (approved):
/lgtm

I think this might need an override for the vsphere tests though.

xrstf:

vCenter is offline

PR #1100, opened 2021-11-12
Include optional subscription attachment in RHEL MachineDeployment 

moadqassem:
Once a rhel machine is created and provisioned using machine controller, if we are using the username and password to register this machine we attach a subscription to it automatically. This is however doesn't give the option for the users to choose whether they would like to attach a subscription to the provisioned machine or not. Also this imposes a limitation for RHEL SCA which is not optimal. This PR gives the user the possibility to choose whether they would like to attach a subscription to the provisioned machine or not.  
Fixes #

**Optional Release Note**:

```release-note
Optional RHEL subscription attachment 
```

### DIFF
--- a/pkg/userdata/rhel/provider.go
+++ b/pkg/userdata/rhel/provider.go
@@ -363,7 +363,7 @@ rh_subscription:
 {{- else }}
     username: "{{.OSConfig.RHELSubscriptionManagerUser}}"
     password: "{{.OSConfig.RHELSubscriptionManagerPassword}}"
-    auto-attach: "{{.OSConfig.AttachSubscription}}"
+    auto-attach: {{.OSConfig.AttachSubscription}}
 {{- end }}
 
 runcmd:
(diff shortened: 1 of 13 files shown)


pkg/userdata/rhel/provider.go
```diff
@@ -363,7 +363,7 @@ rh_subscription:
 {{- else }}
     username: "{{.OSConfig.RHELSubscriptionManagerUser}}"
     password: "{{.OSConfig.RHELSubscriptionManagerPassword}}"
-    auto-attach: true
+    auto-attach: "{{.OSConfig.AttachSubscription}}"
```
ahmedwaleedmalik, line 366:
```suggestion
    auto-attach: {{.OSConfig.AttachSubscription}}
```

Please remove the redundant inverted commas.

ahmedwaleedmalik (changes requested):

ahmedwaleedmalik (approved):
/approve
/lgtm

PR #1102, opened 2021-11-15
run e2e test Openstack Project Auth Provisioning  for one version of k8S

vgramer:
the test does not need to be launched for every k8s version and UUID migration. So it will reduce the number of necessary jobs to pass CI

Fixes #

**Optional Release Note**:

```release-note
NONE
```


moadqassem (approved):
/approve
/lgtm

PR #1104, opened 2021-11-16
osm-namespace flag

### LINKED ISSUE #1103: OSM namespace hardcoding removal
The Namespace in which OSM resources are created is inconsistently hardcoded in MC:
* the reconcile looks for the secret existence in `kube-system`
* the bootstrap secret fetches the provisioning secret from `cloud-init-settings`

Such a namespace shouldn't be hardcoded but parameterized with a new command-line flag.

mlavacca:
A new command-line flag has been introduced to make the OSM integration transparent to the namespace used by OSM.

Fixes #1103 

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -72,7 +72,8 @@ var (
 	nodeCSRApprover                  bool
 	caBundleFile                     string
 
-	useOSM bool
+	useOSM       bool
+	osmNamespace string
 
 	nodeHTTPProxy          string
 	nodeNoProxy            string
@@ -125,7 +126,11 @@ type controllerRunOptions struct {
 
 	node machinecontroller.NodeSettings
 
+	// useOSM tells whether the MC is used jointly with the Operating System Manager
 	useOSM bool
+
+	// The namespace from which to get OSM resources
+	osmNamespace string
 
 	// Assigns the POD networks that will be allocated.
 	podCidr string
@@ -169,6 +174,7 @@ func main() {
 	flag.StringVar(&nodePortRange, "node-port-range", "30000-32767", "A port range to reserve for services with NodePort visibility")
 
 	flag.BoolVar(&useOSM, "use-osm", false, "use osm controller for node bootstrap")
+	flag.StringVar(&osmNamespace, "osm-namespace", "", "the namespace from which to get OSM resources")
 
 	flag.Parse()
 	kubeconfig = flag.Lookup("kubeconfig").Value.(flag.Getter).Get().(string)
@@ -177,6 +183,10 @@ func main() {
 	clusterDNSIPs, err := parseClusterDNSIPs(clusterDNSIPs)
 	if err != nil {
 		klog.Fatalf("invalid cluster dns specified: %v", err)
+	}
+
+	if useOSM && len(osmNamespace) == 0 {
+		klog.Fatalf("-osm-namespace flag is mandatory when -use-osm is set: %v", err)
 	}
 
 	var parsedJoinClusterTimeout *time.Duration
@@ -279,6 +289,7 @@ func main() {
 			),
 		},
 		useOSM:        useOSM,
+		osmNamespace:  osmNamespace,
 		podCidr:       podCidr,
 		nodePortRange: nodePortRange,
 	}
@@ -415,6 +426,7 @@ func (bs *controllerBootstrap) Start(ctx context.Context) error {
 		bs.opt.skipEvictionAfter,
 		bs.opt.node,
 		bs.opt.useOSM,
+		bs.opt.osmNamespace,
 		bs.opt.podCidr,
 		bs.opt.nodePortRange,
 	); err != nil {
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -186,7 +186,7 @@ func main() {
 	}
 
 	if useOSM && len(osmNamespace) == 0 {
-		klog.Fatalf("-osm-namespace flag is mandatory when -use-osm is set: %v", err)
+		klog.Fatalf("-osm-namespace flag is mandatory when -use-osm is set")
 	}
 
 	var parsedJoinClusterTimeout *time.Duration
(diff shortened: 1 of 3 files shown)


vgramer (commented):
/lgtm

mlavacca:
> The namespace should be `cloud-init-settings` and it is added here as const because it's stared between multiple consumers, KKP, KubeOne and OSM thus it should be super consistent. /hold

Totally agree that it has to be super consistent, but as far as I get from the code it isn't now. We look for the provisioning secret in `kube-system` as preliminary check but the bootstrap service tries fetching it from `cloud-init-settings`.

moadqassem:
> > The namespace should be `cloud-init-settings` and it is added here as const because it's stared between multiple consumers, KKP, KubeOne and OSM thus it should be super consistent. /hold
> 
> Totally agree that it has to be super consistent, but as far as I get from the code it isn't now. We look for the provisioning secret in `kube-system` as preliminary check, but the bootstrap service tries fetching it from `cloud-init-settings`.

That's because this secret isn't only created for OSM, it is created for other resources as well. Thus you can't have it in the `cloud-init-settings`, especially that this namespace is kinda exposed to the worker nodes.

ahmedwaleedmalik:
I don't think this is required anymore since https://github.com/kubermatic/machine-controller/pull/1105/files#diff-8ff87bbeb3db2e4cd0d5652974ce6a647c7d38c0ec7efe07d0dd7a664fc335c0 should take care of this.

mlavacca:
> I don't think this is required anymore since #1105 should take care of this.
> 
> https://github.com/kubermatic/machine-controller/pull/1105/files#diff-8ff87bbeb3db2e4cd0d5652974ce6a647c7d38c0ec7efe07d0dd7a664fc335c0

I think this is still needed because the OSM namespace is parameterized. What is the point of having a customizable namespace in which we create resources, but a hardcoded namespace from which we fetch the same resources? At this point, if we want to have it hardcoded I think we should hardcode the OSM namespace as well.

moadqassem:
> > I don't think this is required anymore since #1105 should take care of this.
> 
> > 
> 
> > https://github.com/kubermatic/machine-controller/pull/1105/files#diff-8ff87bbeb3db2e4cd0d5652974ce6a647c7d38c0ec7efe07d0dd7a664fc335c0
> 
> 
> 
> I think this is still needed because the OSM namespace is parameterized. What is the point of having a customizable namespace in which we create resources, but a hardcoded namespace from which we fetch the same resources? At this point, if we want to have it hardcoded I think we should hardcode the OSM namespace as well.

To be honest I am not sure I get your point. What we have here is a namespace that is being used for everything that has to do with OSM and that's `cloud-init-settings` namespace, which purely conventional. Also the lifecycle of OSM related resources is super limited to only the provision of the machine, so only the initial step and after that OSM doesn't play a major role so why mixing up resources that are long lived with resources that have a very limited interaction. I honestly don't quite the point about customizable namespaces.

mlavacca:
> > > I don't think this is required anymore since #1105 should take care of this.
> > 
> > 
> > > 
> > 
> > 
> > > https://github.com/kubermatic/machine-controller/pull/1105/files#diff-8ff87bbeb3db2e4cd0d5652974ce6a647c7d38c0ec7efe07d0dd7a664fc335c0
> > 
> > 
> > I think this is still needed because the OSM namespace is parameterized. What is the point of having a customizable namespace in which we create resources, but a hardcoded namespace from which we fetch the same resources? At this point, if we want to have it hardcoded I think we should hardcode the OSM namespace as well.
> 
> To be honest I am not sure I get your point. What we have here is a namespace that is being used for everything that has to do with OSM and that's `cloud-init-settings` namespace, which purely conventional. Also the lifecycle of OSM related resources is super limited to only the provisioning of the machine, so only the initial step and after that OSM doesn't play a major role so why mixing up resources that are long lived with resources that have a very limited interaction. I honestly don't quite get the point about customizable namespaces.

My point is that the namespace used for everything related to OSM can be changed in the OSM. You can run OSM to make it create OSCs, secrets, etc. in a namespace different from `cloud-init-settings`. You don't have such a possibility in MC, so my point is that I don't see why we should enforce `cloud-init-settings` in the consumer component (MC) and make the namespace customizable in the producer component (OSM).

moadqassem:
> > > > I don't think this is required anymore since #1105 should take care of this.
> > > 
> > > 
> > > > 
> > > 
> > > 
> > > > https://github.com/kubermatic/machine-controller/pull/1105/files#diff-8ff87bbeb3db2e4cd0d5652974ce6a647c7d38c0ec7efe07d0dd7a664fc335c0
> > > 
> > > 
> > > I think this is still needed because the OSM namespace is parameterized. What is the point of having a customizable namespace in which we create resources, but a hardcoded namespace from which we fetch the same resources? At this point, if we want to have it hardcoded I think we should hardcode the OSM namespace as well.
> > 
> > 
> > To be honest I am not sure I get your point. What we have here is a namespace that is being used for everything that has to do with OSM and that's `cloud-init-settings` namespace, which purely conventional. Also the lifecycle of OSM related resources is super limited to only the provisioning of the machine, so only the initial step and after that OSM doesn't play a major role so why mixing up resources that are long lived with resources that have a very limited interaction. I honestly don't quite get the point about customizable namespaces.
> 
> My point is that the namespace used for everything related to OSM can be changed in the OSM. You can run OSM to make it create OSCs, secrets, etc. in a namespace different from `cloud-init-settings`. You don't have such a possibility in MC, so my point is that I don't see why we should enforce `cloud-init-settings` in the consumer component (MC) and make the namespace customizable in the producer component (OSM).

Okay now I understand about which customizable namespace you are talking about. Well, this was done that way, because we wanted to make compatible with K1 and KKP because the resources that are created such as OSP and OSC, are created differently for each software. For example in KKP, we need to hand in the namespace of the user cluster in the seed where in KubeOne you need to hand in one namespace only, got it?

mlavacca:
As discussed with @moadqassem, this PR is gonna be closed in favor of https://github.com/kubermatic/operating-system-manager/pull/72, which will hardcode cloud-init-settings as the namespace in which OSCs and secret will be created.

/close

PR #1106, opened 2021-11-19
Populate /etc/hostname during bootstrap for ignition

ahmedwaleedmalik:
Write machine name to `/etc/hostname`, which is then later on consumed by kubelet to populate the flag `--hostname-override`

Fixes #

**Optional Release Note**:

```release-note
NONE
```


ahmedwaleedmalik:
xref https://github.com/kubermatic/operating-system-manager/issues/65

moadqassem (approved):
@ahmedwaleedmalik I have actually an idea to overcome the absence of the machine name in OSM so we can populate it with Kubelet and other processes. For now we can do it this way though :)

PR #1109, opened 2021-11-23
Flatcar improvements

kron4eg:
The PR improves the Flatcar implementation to close the drift between other providers:

- remove containerized kubelet
- removed docker service dependency
- use the same containerd config and kubelet.service as other providers
- allow to use the pause image as the sandbox image in containerd to gain more control over the used image

This is another version of #1067

**Optional Release Note**:
```release-note
Flatcar improvements
Containerd config improvement
```


moadqassem (approved):
/approve
/lgtm

PR #1125, opened 2021-12-06
Add support for gp3 disk type in AWS

### LINKED ISSUE #1123: Support gp3 volume type in AWS
https://github.com/kubermatic/machine-controller/blob/07e3ca3aa199ca8ada73fada539ef565b45612a1/pkg/cloudprovider/provider/aws/provider.go#L91-L97

machine-controller currently doesn't support `gp3` as volume type in AWS. It was introduced a year ago (see https://aws.amazon.com/about-aws/whats-new/2020/12/introducing-new-amazon-ebs-general-purpose-volumes-gp3/), so having it available would be nice.

embik:
Keeping up with AWS, this adds support for the `gp3` disk type. The difference to `gp2` is that you can define the IOPS for the disk to not depend on disk size for guaranteed IOPS.

Fixes #1123

**Optional Release Note**:

```release-note
Add support for gp3 disk type in AWS
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -426,7 +426,7 @@ func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *providerconfigt
 			}
 		} else if c.DiskType == ec2.VolumeTypeGp3 {
 			if iops < 3000 || iops > 64000 {
-				return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 100, max: 64000)")
+				return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 3000, max: 64000)")
 			}
 		}
 
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -420,18 +420,22 @@ func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *providerconfigt
 		}
 		iops := *rawConfig.DiskIops
 
-		if c.DiskType == ec2.VolumeTypeIo1 {
-			if iops < 100 || iops > 64000 {
-				return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 100, max: 64000)")
-			}
-		} else if c.DiskType == ec2.VolumeTypeGp3 {
-			if iops < 3000 || iops > 64000 {
-				return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 3000, max: 64000)")
-			}
+		if iops < 100 || iops > 64000 {
+			return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 100, max: 64000)")
+		}
+
+		c.DiskIops = rawConfig.DiskIops
+	} else if c.DiskType == ec2.VolumeTypeGp3 && rawConfig.DiskIops != nil {
+		// gp3 disks start with 3000 IOPS by default, we _can_ pass better IOPS, but it is not a required field
+		iops := *rawConfig.DiskIops
+
+		if iops < 3000 || iops > 64000 {
+			return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 3000, max: 64000)")
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -419,8 +420,14 @@ func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *providerconfigt
...
-			return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 100, max: 64000)")
+		if c.DiskType == ec2.VolumeTypeIo1 {
+			if iops < 100 || iops > 64000 {
+				return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 100, max: 64000)")
+			}
+		} else if c.DiskType == ec2.VolumeTypeGp3 {
+			if iops < 3000 || iops > 64000 {
+				return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 3000, max: 64000)")
+			}
 		}
```
ahmedwaleedmalik, line 431: This needs to be moved out of the if block. Right now it looks like:

```
if c.DiskType == ec2.VolumeTypeIo1 {
	//....
	if c.DiskType == ec2.VolumeTypeIo1 {
	//....
	} else if c.DiskType == ec2.VolumeTypeGp3 {
	//....
	}
	//....
}
```

You'll never hit the condition `else if c.DiskType == ec2.VolumeTypeGp3`

ahmedwaleedmalik (changes requested):

pkg/cloudprovider/provider/aws/provider.go
embik: Aha! good catch, thanks. I actually did not see the outer if condition.

embik: Makes you wonder how many issues you can introduce with 10 lines of change ...

embik: Should be fixed now! PTAL.

pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -423,8 +424,18 @@ func (p *provider) getConfig(s v1alpha1.ProviderSpec) (*Config, *providerconfigt
 			return nil, nil, nil, errors.New("Invalid value for `diskIops` (min: 100, max: 64000)")
 		}
 
+		c.DiskIops = rawConfig.DiskIops
+	} else if c.DiskType == ec2.VolumeTypeGp3 && rawConfig.DiskIops != nil {
+		// gp3 disks start with 3000 IOPS by default, we _can_ pass better IOPS, but it is not a required field
+		iops := *rawConfig.DiskIops
```
ahmedwaleedmalik, line 430: Can we please move the null check for `rawConfig.DiskIops`  out of the if blocks altogether to remove redundancy?  
What you are doing is right too but I think instead of bypassing it with `else if c.DiskType == ec2.VolumeTypeGp3 && rawConfig.DiskIops != nil ` we should return an **explicit** error if it's nil since it's a **required field**. 

Similar to https://github.com/kubermatic/machine-controller/blob/184dd4bdde4d6378eb6979bcb94e4049afcc19ef/pkg/cloudprovider/provider/aws/provider.go#L419

Maybe something like this should do: 

```
if len(c.DiskType) > 0 && rawConfig.DiskIops == nil {
	return nil, nil, nil, errors.New("Missing required field `diskIops`")
}
```

embik: It's actually not a required field for `gp3`, that's why I added the comment. See "Iops" on https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html#API_CreateVolume_RequestParameters. Setting IOPS when creating a volume is only required for `io1` (and `io2`) disks.

So, for `gp3` disks, we only want to validate the field if it's set. If it's not, there's nothing to validate, and you get a `gp3` disk with 3000 IOPS.

ahmedwaleedmalik: Got it, that makes sense :+1:

ahmedwaleedmalik (approved):
/lgtm
/approve

PR #1133, opened 2021-12-13
New -node-containerd-registry-mirrors flag

kron4eg:
This is a companion/dependency PR for https://github.com/kubermatic/kubeone/pull/1674

**Optional Release Note**:

```release-note
New -node-containerd-registry-mirrors flag
```

### DIFF
--- a/cmd/machine-controller/custom_flags.go
+++ b/cmd/machine-controller/custom_flags.go
@@ -1,0 +1,60 @@
+/*
+Copyright 2019 The Machine Controller Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+type registryMirrorsFlags map[string][]string
+
+func (fl registryMirrorsFlags) Set(val string) error {
+	split := strings.SplitN(val, "=", 2)
+	if len(split) != 2 {
+		return fmt.Errorf("should have exactly 1 =")
+	}
+
+	key, value := split[0], split[1]
+	slice := fl[key]
+	slice = append(slice, value)
+	fl[key] = slice
+
+	return nil
+}
+
+func (fl registryMirrorsFlags) String() string {
+	var (
+		registryNames []string
+		result        []string
+	)
+
+	for registryName := range fl {
+		registryNames = append(registryNames, registryName)
+	}
+
+	sort.Strings(registryNames)
+
+	for _, registryName := range registryNames {
+		for _, mirror := range fl[registryName] {
+			result = append(result, fmt.Sprintf("%s=%s", registryName, mirror))
+		}
+	}
+
+	return fmt.Sprintf("%v", result)
+}
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -23,6 +23,8 @@ import (
 	"net"
 	"net/http"
 	"net/http/pprof"
+	"net/url"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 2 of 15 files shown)


cmd/machine-controller/main.go
```diff
@@ -253,10 +295,23 @@ func main() {
 	var registryMirrors []string
 	for _, mirror := range strings.Split(nodeRegistryMirrors, ",") {
 		if trimmedMirror := strings.TrimSpace(mirror); trimmedMirror != "" {
+			if !strings.HasPrefix(mirror, "http") {
+				trimmedMirror = "https://" + mirror
```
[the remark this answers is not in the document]
kron4eg: https://docs.docker.com/registry/insecure/

> With insecure registries enabled, Docker goes through the following steps:
> * First, try using HTTPS.
>   * If HTTPS is available but the certificate is invalid, ignore the error about the certificate.
>   * If HTTPS is not available, fall back to HTTP.

And so, by default (in insecure is not specified and no explicit http://) the registry is considered secure, hence is -- https://

cmd/machine-controller/main.go
```diff
@@ -83,12 +85,51 @@ var (
...
+	sort.Strings(registryNames)
+
+	for _, registryName := range registryNames {
+		for _, mirror := range fl[registryName] {
+			result = append(result, fmt.Sprintf("%s=%s", registryName, mirror))
+		}
+	}
+
+	return fmt.Sprintf("%v", result)
+}
```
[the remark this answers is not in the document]
kron4eg: :+1: :heavy_check_mark:

cmd/machine-controller/custom_flags.go
```diff
@@ -0,0 +1,60 @@
...
+	var (
+		registryNames []string
+		result        []string
+	)
+
+	for registryName := range fl {
+		registryNames = append(registryNames, registryName)
+	}
+
+	sort.Strings(registryNames)
```
xmudrii, line 51: Why is sorting required?

kron4eg: yes, to have stable output instead of randomized

cmd/machine-controller/main.go
```diff
@@ -253,10 +258,23 @@ func main() {
...
+				klog.Fatalf("incorrect mirror provided: %v", err)
+			}
+
 			registryMirrors = append(registryMirrors, trimmedMirror)
 		}
 	}
 
+	if len(registryMirrors) > 0 {
+		nodeContainerdRegistryMirrors["docker.io"] = registryMirrors
+	}
```
xmudrii, line 276: Should we have some validation to fail starting MC if both `registryMirrors` and `nodeContainerdRegistryMirrors["docker.io"]` are provided by the user?

kron4eg: No, if docker.io is empty it's OK for dockerd

cmd/machine-controller/main.go
```diff
@@ -163,6 +167,7 @@ func main() {
 	flag.StringVar(&nodePauseImage, "node-pause-image", "", "Image for the pause container including tag. If not set, the kubelet default will be used: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/")
 	flag.StringVar(&nodeKubeletRepository, "node-kubelet-repository", "quay.io/kubermatic/kubelet", "Repository for the kubelet container. Only has effect on Flatcar Linux.")
 	flag.StringVar(&nodeContainerRuntime, "node-container-runtime", "docker", "container-runtime to deploy")
+	flag.Var(&nodeContainerdRegistryMirrors, "node-containerd-registry-mirrors", "Configure registry mirrors endpoints. Can be used multiple times to specify multiple mirrors")
```
embik, line 170: It seems I can't add a comment to the line for `node-registry-mirrors` above this one, but should we update the description for `node-registry-mirrors` to either call it deprecated or add that it only is effective for `docker.io` images?

cmd/machine-controller/main.go
xmudrii: I'm not sure I get it, but isn't one flag going to override another one? For example, if you have both `--node-registry-mirrors` and `--node-containerd-registry-mirrors`, isn't only one going to have effect?

cmd/machine-controller/main.go
kron4eg: it's not depricated :)

cmd/machine-controller/main.go
kron4eg: Essentially it will, but I don't see it as a problem.

cmd/machine-controller/main.go
embik: Gotcha, wasn't sure what status we considered for it.

cmd/machine-controller/main.go
kron4eg: It's kind of backward compatibility, it's possible to configure docker and containerd at the same time (of course only 1 will be used, according to -node-container-runtime).

xmudrii: Okay, that makes sense, thanks for the explanation!

xmudrii (approved):

PR #1139, opened 2021-12-21
Fix evictionHard parsing

kron4eg:
Eviction hard [mapStringString flag](https://pkg.go.dev/k8s.io/component-base@v0.23.1/cli/flag#LangleSeparatedMapStringString) is using symbol `<` as a separator between pairs.

```release-note
Fix evictionHard parsing
```


moadqassem (approved):
/approve
/lgtm

PR #1143, opened 2022-01-03
update in kubervirt rawconfig

### LINKED ISSUE #1113: Allow creation of MD from Virtual Machine Presets
Machine Controller should have a possibility to create virtual machines from virtual machines presets with support for custom templating.

sankalp-r:

This PR updates kubeVirt's rawConfig and creation of VM using presets.

Fixes #1113 

**Optional Release Note**:

```release-note
None
```


ahmedwaleedmalik:
@sankalp-r IMO this changeset introduces breaking changes for the existing machine-deployments. I think we need to consider that as well, wdyt?

sankalp-r:
> @sankalp-r IMO this changeset introduces breaking changes for the existing kubevirt machine-deployments. I think we need to consider that as well, wdyt?

@ahmedwaleedmalik Yes these changes introduces breaking changes.
The idea is to fix machine-controller first, and then consume/update machine-controller version in KKP.

sankalp-r:
> @sankalp-r What is the status of this PR ?

As the UI for this change is not ready yet, can't proceed ahead.

This change needs UI changes as well, mentioned by Marcin:
https://github.com/kubermatic/machine-controller/issues/1113#issuecomment-1000767056

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -302,7 +332,7 @@ func (p *provider) MachineMetricsLabels(machine *clusterv1alpha1.Machine) (map[s
 	if err == nil {
 		labels["cpus"] = c.CPUs
 		labels["memoryMIB"] = c.Memory
-		labels["sourceURL"] = c.SourceURL
+		labels["sourceURL"] = c.OsImage
```
hdurand0710, line 335: We should also rename the metrics label from `sourceURL` to `OsImage`

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -346,102 +397,42 @@ func (p *provider) Create(machine *clusterv1alpha1.Machine, _ *cloudprovidertype
...
+	cpu := resourceRequirements.Requests[corev1.ResourceCPU]
 	virtualMachine := &kubevirtv1.VirtualMachine{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      machine.Name,
-			Namespace: c.Namespace,
+			Name:      c.VirtualMachineName,
+			Namespace: metav1.NamespaceDefault,
 			Labels: map[string]string{
-				"kubevirt.io/vm": machine.Name,
+				"kubevirt.io/vm": c.VirtualMachineName,
```
hdurand0710, line 406: We are missing the labels coming from the Preset here (from line 372).

sankalp-r: Preset labels are only required to be added in `VirtualMachineInstanceTemplateSpec` (line 414)

hdurand0710: perfect.

ahmedwaleedmalik:
/hold

IMO we need to consider the migration of old KubeVirt MachineDeployments to the new structure. 

CC @moadqassem (Since you are assigned as a reviewer)

ahmedwaleedmalik:
Thanks, for the clarification @hdurand0710 👍

moadqassem:
@sankalp-r Please let's have a discussions before releasing this. 

/hold

hdurand0710:
Discussion is added to the SIG-Virtualization meeting topics for today.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -72,16 +72,24 @@ func New(configVarResolver *providerconfig.ConfigVarResolver) cloudprovidertypes
...
-	Namespace        string
-	SourceURL        string
+	Kubeconfig         string
+	RestConfig         *rest.Config
+	DNSConfig          *corev1.PodDNSConfig
+	DNSPolicy          corev1.DNSPolicy
+	CPUs               string
+	Memory             string
+	Namespace          string
+	OsImage            string
```
moadqassem, line 82: Just as discussed earlier, I guess we need to make this a bit more generic type so that we can include importing images from url or importing an image via data volume cloning. Thus I would suggest that we convert this type from string to struct. Something like this: 
```go
// URL and DataVolumeName are mutually exclusive 
type OSImage struct{
// if not empty, importing image from an http endpoint 
URL                        string
// if not empty, importing image from a DV clone 
DataVolumeName string
}

File: pkg/cloudprovider/provider/kubevirt/types/types.go [hunk 4]
```diff
@@ -19,20 +19,57 @@ package types
...
+type Flavor struct {
+	Name    providerconfigtypes.ConfigVarString `json:"name,omitempty"`
+	Profile providerconfigtypes.ConfigVarString `json:"profile,omitempty"`
+}
+
+// Template
+type Template struct {
+	CPUs           providerconfigtypes.ConfigVarString `json:"cpus,omitempty"`
+	Memory         providerconfigtypes.ConfigVarString `json:"memory,omitempty"`
+	PrimaryDisk    PrimaryDisk                         `json:"primaryDisk,omitempty"`
```
moadqassem, line 54: Why not just having Disks, instead of Primary disk and secondary disks?

sankalp-r: Here is the rawConfig definition from Marcin: https://kubermatic.slack.com/archives/C01JHBHE82F/p1638872204002600

`Resources` inside `DataVolumeTemplateSpec` for root disk comes from PrimaryDisk, so having primary and secondary disks as separate helps in identifying the resources.

moadqassem:
/hold 

There are some issues with azure cloud will fix it and unhold the pr

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -72,16 +72,29 @@ func New(configVarResolver *providerconfig.ConfigVarResolver) cloudprovidertypes
...
-	Memory           string
-	Namespace        string
-	SourceURL        string
+	Kubeconfig         string
+	RestConfig         *rest.Config
+	DNSConfig          *corev1.PodDNSConfig
+	DNSPolicy          corev1.DNSPolicy
+	CPUs               string
+	Memory             string
+	Namespace          string
```
hdurand0710, line 81: As we will merge https://github.com/kubermatic/machine-controller/pull/1204 in the same release, we should now remove the `Namespace`.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -135,43 +148,53 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
...
+	config.Memory, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.VirtualMachine.Template.Memory)
 	if err != nil {
 		return nil, nil, fmt.Errorf(`failed to get value of "memory" field: %v`, err)
 	}
-	config.Namespace, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Namespace)
-	if err != nil {
-		return nil, nil, fmt.Errorf(`failed to get value of "namespace" field: %v`, err)
-	}
-	config.SourceURL, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.SourceURL)
+	config.Namespace = metav1.NamespaceDefault
```
hdurand0710, line 163: Due to https://github.com/kubermatic/machine-controller/pull/1204, we should now remove this hard-coded value and use the `getNamespace()` method.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -135,43 +153,53 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
...
+	config.Memory, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.VirtualMachine.Template.Memory)
 	if err != nil {
 		return nil, nil, fmt.Errorf(`failed to get value of "memory" field: %v`, err)
 	}
-	config.Namespace, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Namespace)
-	if err != nil {
-		return nil, nil, fmt.Errorf(`failed to get value of "namespace" field: %v`, err)
-	}
-	config.SourceURL, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.SourceURL)
+	config.Namespace = vmNamespace
```
hdurand0710, line 168: Why not just calling getNamespace( ) ?

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -46,13 +46,18 @@ import (
...
 )
 
+var vmNamespace string
+
 func init() {
 	if err := kubevirtv1.AddToScheme(scheme.Scheme); err != nil {
 		klog.Fatalf("failed to add kubevirtv1 to scheme: %v", err)
 	}
+
+	vmNamespace = getNamespace()
```
hdurand0710, line 60: Not a big fan of this variable here.

pkg/cloudprovider/provider/kubevirt/provider.go
hdurand0710: I find it confusing to keep the `Namespace` here as it's not a configurable option any more..

pkg/cloudprovider/provider/kubevirt/provider.go
sankalp-r: Instead of calling `getNamespace()` multiple times, we can call it once inside `init()` and use `vmNamespace` wherever required.

hdurand0710: But it's useless to store a variable. It's not a big object, but still....

sankalp-r: Yes, I agree with you. 
My previous experience with `os.Getenv()` is to better cache its value for reuse, as it affects the performance if it gets called multiple times.

hdurand0710: How to you prevent the value from being changed? Don't you think it's dangerous ?

sankalp-r: As in this case, performance is not the concern, I will use `getNamespace()` directly and remove `vmNamespace` variable.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -321,17 +375,38 @@ func (p *provider) Create(machine *clusterv1alpha1.Machine, _ *cloudprovidertype
 	// because its pod got deleted
 	// The secret has an ownerRef on the VMI so garbace collection will take care of cleaning up
 	terminationGracePeriodSeconds := int64(30)
-	userDataSecretName := fmt.Sprintf("userdata-%s-%s", machine.Name, strconv.Itoa(int(time.Now().Unix())))
-	requestsAndLimits, err := parseResources(c.CPUs, c.Memory)
+	userDataSecretName := fmt.Sprintf("userdata-%s-%s", c.VirtualMachineName, strconv.Itoa(int(time.Now().Unix())))
+
+	resourceRequirements := kubevirtv1.ResourceRequirements{}
+	labels := map[string]string{"kubevirt.io/vm": c.VirtualMachineName}
```
hdurand0710, line 381: Before we had `machine.Name` , which was unique for each `machine` (if replicas=3 in MD, we have 3 machines, each having a name).
Here `c.VirtualMachineName` is common to all machines, as comming from the MD.
Is it really what we want ?

It's a good idea to have a common label among all machines from the same MD (we would also need it for the kubermatic/kubermatic#8742), but the label key should contain `md` and not `vm`.
It's also a good idea to keep `kubevirt.io/vm` = `machine.Name`
having both makes sense.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -321,17 +375,38 @@ func (p *provider) Create(machine *clusterv1alpha1.Machine, _ *cloudprovidertype
...
 	}
+	ctx := context.Background()
 
-	var (
-		pvcRequest     = corev1.ResourceList{corev1.ResourceStorage: c.PVCSize}
-		dataVolumeName = machine.Name
+	// Add VMIPreset label if specified
+	if c.FlavorName != "" {
+		vmiPreset := kubevirtv1.VirtualMachineInstancePreset{}
+		if err := sigClient.Get(ctx, types.NamespacedName{Namespace: c.Namespace, Name: c.FlavorName}, &vmiPreset); err != nil {
```
hdurand0710, line 392: Until we have implemented https://app.zenhub.com/workspaces/work-in-progress---team-caffeine-overflow-6166fa570192a800141b6db9/issues/kubermatic/kubermatic/8490
We have a fetch the `Presets` from `default` namespace

pkg/cloudprovider/provider/kubevirt/provider.go
hdurand0710: I still think that it's a good idea to have the same labels in the VM and in the VMI.

pkg/cloudprovider/provider/kubevirt/types/types.go
```diff
@@ -19,20 +19,56 @@ package types
...
+	SecondaryDisks []SecondaryDisks                    `json:"secondaryDisks,omitempty"`
+}
+
+// PrimaryDisk
+type PrimaryDisk struct {
+	Disk
+	OsImage providerconfigtypes.ConfigVarString `json:"osImage,omitempty"`
```
hdurand0710, line 60: Do we want to have here the same struct as in provider.go ?

moadqassem (approved):
/approve
/lgtm

PR #1144, opened 2022-01-03
Containerd daemon level registry auth configuration

kron4eg:

xref https://github.com/kubermatic/kubeone/issues/1574

**Optional Release Note**:

```release-note
containerd daemon level registry auth configuration
```

### DIFF
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -28,7 +28,6 @@ import (
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
-	osmv1alpha1 "k8c.io/operating-system-manager/pkg/crd/osm/v1alpha1"
 
 	clusterv1alpha1 "github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
 	"github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1/migrations"
@@ -44,6 +43,7 @@ import (
 	machinesv1alpha1 "github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
 	"github.com/kubermatic/machine-controller/pkg/node"
 	"github.com/kubermatic/machine-controller/pkg/signals"
+	osmv1alpha1 "k8c.io/operating-system-manager/pkg/crd/osm/v1alpha1"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -69,20 +69,20 @@ var (
 	workerCount                      int
 	bootstrapTokenServiceAccountName string
 	skipEvictionAfter                time.Duration
-	nodeCSRApprover                  bool
 	caBundleFile                     string
 
 	useOSM bool
 
-	nodeHTTPProxy          string
-	nodeNoProxy            string
-	nodeInsecureRegistries string
-	nodeRegistryMirrors    string
-	nodePauseImage         string
-	nodeContainerRuntime   string
-	podCidr                string
-	nodePortRange          string
-
+	nodeCSRApprover               bool
+	nodeHTTPProxy                 string
+	nodeNoProxy                   string
+	nodeInsecureRegistries        string
+	nodeRegistryMirrors           string
+	nodePauseImage                string
+	nodeContainerRuntime          string
+	podCidr                       string
+	nodePortRange                 string
+	nodeRegistryCredentialsSecret string
 	nodeContainerdRegistryMirrors = registryMirrorsFlags{}
 )
 
@@ -169,7 +169,7 @@ func main() {
 	flag.BoolVar(&nodeCSRApprover, "node-csr-approver", true, "Enable NodeCSRApprover controller to automatically approve node serving certificate requests")
 	flag.StringVar(&podCidr, "pod-cidr", "172.25.0.0/16", "The network ranges from which POD networks are allocated")
 	flag.StringVar(&nodePortRange, "node-port-range", "30000-32767", "A port range to reserve for services with NodePort visibility")
-
+	flag.StringVar(&nodeRegistryCredentialsSecret, "node-registry-credentials-secret", "", "A Secret object reference, that containt auth info for image registry in namespace/secret-name form, example: kube-system/registry-credentials. Secret format: map[string]github.com/containerd/containerd/pkg/cri/config.AuthConfig")
 	flag.BoolVar(&useOSM, "use-osm", false, "use osm controller for node bootstrap")
 
 	flag.Parse()
@@ -273,10 +273,11 @@ func main() {
 		skipEvictionAfter:    skipEvictionAfter,
 		nodeCSRApprover:      nodeCSRApprover,
 		node: machinecontroller.NodeSettings{
-			ClusterDNSIPs: clusterDNSIPs,
-			HTTPProxy:     nodeHTTPProxy,
-			NoProxy:       nodeNoProxy,
-			PauseImage:    nodePauseImage,
+			ClusterDNSIPs:                clusterDNSIPs,
+			HTTPProxy:                    nodeHTTPProxy,
+			NoProxy:                      nodeNoProxy,
+			PauseImage:                   nodePauseImage,
+			RegistryCredentialsSecretRef: nodeRegistryCredentialsSecret,
 			ContainerRuntime: containerruntime.Get(
 				nodeContainerRuntime,
 				containerruntime.WithInsecureRegistries(insecureRegistries),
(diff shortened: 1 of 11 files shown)


cmd/machine-controller/main.go
```diff
@@ -169,7 +169,7 @@ func main() {
 	flag.BoolVar(&nodeCSRApprover, "node-csr-approver", true, "Enable NodeCSRApprover controller to automatically approve node serving certificate requests")
 	flag.StringVar(&podCidr, "pod-cidr", "172.25.0.0/16", "The network ranges from which POD networks are allocated")
 	flag.StringVar(&nodePortRange, "node-port-range", "30000-32767", "A port range to reserve for services with NodePort visibility")
-
+	flag.StringVar(&nodeRegistryCredentialsSecret, "node-registry-credentials-secret", "", "A Secret object reference, that containt auth info for image registry in namespace/secret-name form, example: kube-system/registry-credentials. Secret format: map[string]github.com/containerd/containerd/pkg/cri/config.AuthConfig")
```
ahmedwaleedmalik, line 172: This isn't intuitive enough and seems a bit off to me 
```
Secret format: map[string]github.com/containerd/containerd/pkg/cri/config.AuthConfig
```

Maybe it can be replaced with something like 
```
Secret format: map[string]AuthConfig with AuthConfig defined at https://github.com/kubermatic/machine-controller/blob/master/pkg/containerruntime/containerd.go#L223
```

pkg/containerruntime/containerd.go
```diff
@@ -172,3 +170,132 @@ systemctl enable --now containerd
...
+type containerdRegistryConfig struct {
+	TLS  *containerdRegistryTLSConfig `toml:"tls"`
+	Auth *AuthConfig                  `toml:"auth"`
+}
+
+type containerdRegistryTLSConfig struct {
+	InsecureSkipVerify bool `toml:"insecure_skip_verify"`
+}
+
+// AuthConfig is a COPY of github.com/containerd/containerd/pkg/cri/config.AuthConfig.
```
ahmedwaleedmalik, line 221: https://github.com/containerd/containerd/pkg/cri/config.AuthConfig returns 404 `Not Found`

Maybe we can just replace it with https://github.com/containerd/containerd/blob/v1.5.9/pkg/cri/config/config.go#L126

cmd/machine-controller/main.go
ahmedwaleedmalik: Just a nit, if you think it's clear enough then let's just go with it

pkg/controller/machine/machine_controller.go
```diff
@@ -721,6 +724,27 @@ func (r *Reconciler) ensureInstanceExistsForMachine(
...
+				}
+
+				for registry, data := range credsSecret.Data {
+					var regCred containerruntime.AuthConfig
+					if err := json.Unmarshal(data, &regCred); err != nil {
+						return nil, fmt.Errorf("failed to unmarshal registry credentials: %w", err)
+					}
+					registryCredentials[registry] = regCred
+				}
+			}
```
ahmedwaleedmalik, line 743: Would be nice to have an error or at the very least a warning `if len(secRef) != 2` because at the moment it would just get skipped past

ahmedwaleedmalik (changes requested):
This looks great, just a few minor changes required.

pkg/containerruntime/containerd.go
kron4eg: because `github.com/containerd/containerd/pkg/cri/config.AuthConfig` is Go type identity, not an URL like https://github.com/containerd/containerd/blob/v1.5.9/pkg/cri/config/config.go#L126

ahmedwaleedmalik: Yes, I'm completely aware of that but IMO having an accessible URL here can be better since it's documentation. But fine either way :)

docs/registry-authentication.md
```diff
@@ -0,0 +1,26 @@
+# Registry Authentication
+
+Machine-controller supports configuring container-runtime with authentication
```
xmudrii, line 3: Tiny nit:

```suggestion
machine-controller supports configuring container runtime with authentication
```

docs/registry-authentication.md
```diff
@@ -0,0 +1,26 @@
...
+information. Flag `-node-registry-credentials-secret` can take a secret
+reference where authentication info will be stored. During VM creation this info
+will be used to configure container-runtime.
```
xmudrii, line 6:
```suggestion
will be used to configure container runtime.
```

docs/registry-authentication.md
```diff
@@ -0,0 +1,26 @@
+# Registry Authentication
+
+Machine-controller supports configuring container-runtime with authentication
+information. Flag `-node-registry-credentials-secret` can take a secret
```
xmudrii, line 4: Please describe what value this flag expects (i.e. namespace/secret-name).

docs/registry-authentication.md
```diff
@@ -0,0 +1,26 @@
...
+reference where authentication info will be stored. During VM creation this info
+will be used to configure container-runtime.
+
+Secret format is serialized `map[string]github.com/containerd/containerd/pkg/cri/config.AuthConfig`
```
xmudrii: Similar to @ahmedwaleedmalik’s comment, this is not really intuitive, so we should probably replace it with a link to the struct.

docs/registry-authentication.md
```diff
@@ -0,0 +1,26 @@
...
+
+example:
```
xmudrii, line 29:
```suggestion
Example:
```

docs/registry-authentication.md
```diff
@@ -0,0 +1,26 @@
...
+  name: my-registries
+  namespace: kube-system
+data:
+  gcr.io: |
+    eyJ1c2VybmFtZSI6ImwwZzFuIiwicGFzc3dvcmQiOiJjMDBscDQ1NXc
+    wcmQiLCJhdXRoIjoiIiwiaWRlbnRpdHl0b2tlbiI6IiJ9Cg==
+
+```
+
+Now having this saved in kubernetes API, launch machine-controller with
```
xmudrii, line 44:
```suggestion
Now having this saved in the Kubernetes API, launch machine-controller with
```

pkg/containerruntime/containerd.go
```diff
@@ -172,3 +170,132 @@ systemctl enable --now containerd
...
+		criPlugin.Registry.Configs[registry] = containerdRegistryConfig{
+			TLS: &containerdRegistryTLSConfig{
+				InsecureSkipVerify: true,
+			},
+		}
+	}
+
+	for registry, auth := range eng.registryCredentials {
+		regConfig := criPlugin.Registry.Configs[registry]
+		auth := auth
```
xmudrii, line 278: What’s the point of this line? To avoid lint failures?

pkg/containerruntime/containerd.go
```diff
@@ -172,3 +170,132 @@ systemctl enable --now containerd
...
+		regConfig.Auth = &auth
+		criPlugin.Registry.Configs[registry] = regConfig
```
xmudrii, line 280: I’m worried this can throw NPEs if the Configs map doesn’t have the target registry.

pkg/controller/machine/machine_controller.go
kron4eg: changed validation of the value on the fatal error on the start up

pkg/controller/machine/machine_controller.go
```diff
@@ -721,6 +724,27 @@ func (r *Reconciler) ensureInstanceExistsForMachine(
 				externalCloudProvider, _ = strconv.ParseBool(val)
 			}
 
+			registryCredentials := map[string]containerruntime.AuthConfig{}
+
+			if secRef := strings.SplitN(r.nodeSettings.RegistryCredentialsSecretRef, "/", 2); len(secRef) == 2 {
+				var credsSecret corev1.Secret
+				err := r.client.Get(ctx, types.NamespacedName{Namespace: secRef[0], Name: secRef[1]}, &credsSecret)
```
xmudrii, line 731: This will require changes to RBAC so machine-controller can access the Secret. We need to somehow document that.

cmd/machine-controller/main.go
kron4eg: Added link to the doc where format and structure is explained

pkg/containerruntime/containerd.go
kron4eg: to make a copy, because original auth is an iteration variable, and it's being reused every iteration. We need a copy because we take pointer of this var inside the loop.

pkg/containerruntime/containerd.go
kron4eg: If map has no such key, empty value will be initialized and I'm exploiting this fact.

kron4eg: https://go.dev/play/p/l-5vNYpL6qx

xmudrii (approved):
Besides the RBAC comment, looks good to me.

/lgtm
/approve

kron4eg:
I think this clusterRole lets machine-controller to read all the secrets in the Cluster https://github.com/kubermatic/machine-controller/blob/master/examples/machine-controller.yaml#L532-L537

ahmedwaleedmalik (approved):
/approve
/assign @xrstf due to the vsphere issue

ahmedwaleedmalik:
/assign @xrstf
Required due to the vsphere issue

PR #1145, opened 2022-01-04
Add E2E test case for Nutanix cloud provider

### LINKED ISSUE #1138: E2E tests for Nutanix cloud provider
**User Story**
https://github.com/kubermatic/machine-controller/pull/1132 introduces support for Nutanix to machine-controller. We want to implement e2e tests on top of our internal Nutanix setup to test this provider.

**Acceptance Criteria**
An E2E test for Nutanix is added to the prow configuration of machine-controller.

embik:

Adding an e2e test case for Nutanix as cloud provider.

To overcome the limitations of the private Nutanix setup, the changes in the ci scripts tunnel the HTTP proxy that is available to the CI infrastructure into the VM via SSH. To bind the forwarded port to a secure IP that is available from a Pod (but not the internet), the Terraform is expanded to include a private network on Hetzner.

The "migrateUID" test scenario is excluded, as the provider implementation is a noop. It does not work well with the HTTP proxy as the given proxy IP is only available inside the VM, not inside the prow job.

Fixes #1138

**Optional Release Note**:

```release-note
NONE
```


.prow.yaml
```diff
@@ -566,6 +566,29 @@ presubmits:
               memory: 1Gi
               cpu: 500m
 
+  - name: pull-machine-controller-e2e-nutanix
+    optional: true
+    always_run: true
```
moadqassem, line 571: I would remove this line here

hack/ci-e2e-test.sh
```diff
@@ -95,6 +98,11 @@ export E2E_SSH_PUBKEY="$(cat ~/.ssh/id_rsa.pub)"
 ./test/tools/integration/provision_master.sh
 
 echo "Running e2e tests..."
+if [[ ! -z "${NUTANIX_E2E_PROXY_HOST:-}" ]]; then
```
moadqassem, line 101: Ok, so this is a bit hacky and might be error prune, however let's go with it for now. The next step would be, introducing connectivity on a higher level. Like running the control plane stack on Nutanix as well.

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -880,6 +881,45 @@ func TestScalewayProvisioningE2E(t *testing.T) {
...
+	project := os.Getenv("NUTANIX_E2E_PROJECT_NAME")
+	subnet := os.Getenv("NUTANIX_E2E_SUBNET_NAME")
+	endpoint := os.Getenv("NUTANIX_E2E_ENDPOINT")
+
+	if password == "" || username == "" || endpoint == "" || cluster == "" || project == "" || subnet == "" {
+		t.Fatal("unable to run the test suite, NUTANIX_E2E_PASSWORD, NUTANIX_E2E_USERNAME, NUTANIX_E2E_CLUSTER_NAME, " +
+			"NUTANIX_E2E_ENDPOINT, NUTANIX_E2E_PROJECT_NAME or NUTANIX_E2E_SUBNET_NAME environment variables cannot be empty")
+	}
+
+	proxyURL := os.Getenv("NUTANIX_E2E_PROXY_URL")
```
moadqassem, line 898: Maybe a bit of documentation about this variable?

test/e2e/provisioning/testdata/machinedeployment-nutanix.yaml
```diff
@@ -0,0 +1,45 @@
...
+            subnetName: '<< NUTANIX_SUBNET >>'
+            imageName: 'machine-controller-e2e-<< OS_NAME >>'
+            cpus: 2
+            memoryMB: 2048
+            diskSize: 20
+          operatingSystem: "<< OS_NAME >>"
+          operatingSystemSpec:
+            distUpgradeOnBoot: false
+            disableAutoUpdate: true
+            attachSubscription: false
```
moadqassem, line 43: Why do you need this here? This is for RHEL thus I would remove it

moadqassem (commented):
It looks good, just left some minor changes

.prow.yaml
embik: All the other private providers like kubevirt and vSphere are set to `always_run`. Shouldn't we keep testing Nutanix as much as possible? Running the test is basically "free" for us.

test/e2e/provisioning/testdata/machinedeployment-nutanix.yaml
embik: Copied from another `MachineDeployment` testdata YAML. I'll remove it.

test/e2e/provisioning/all_e2e_test.go
embik: Sounds reasonable. Where would you expect the documentation? Inline with the code? Somewhere else?

hack/ci-e2e-test.sh
embik: That's a good suggestion, thanks. The alternative would be to move the machine-controller e2e tests to the same model as the KKP E2E tests and run the control plane in kind, eliminating the VM completely. But I agree this should be seen more as a temporary hack than a permanent solution.

PR #1136, opened 2021-12-17
Add validation for machinedeployments when using OSM

ahmedwaleedmalik:
**What this PR does / why we need it**:
Adds validations for `MachineDeployment` in machine-controller webhooks, when OSM is enabled.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Optional Release Note**:

```release-note
NONE
```

PR #1147, opened 2022-01-04
Use default scheme and pass request context to ConfigVarResolver

### LINKED ISSUE #1142: secretKeyRef for `token` missing for Hetzner
While deploying the example yaml for Hetzner the following is thrown:

```secret/machine-controller-hetzner configured
Error from server: error when creating "hetzner-machinedeployment.yaml": admission webhook "machinedeployments.machine-controller.kubermatic.io" denied the request: validation failed: token is missing
```

Machine controller version: v1.40.1

embik:
This PR fixes two problems:

1. #1136 introduced usage of an empty/fresh `Scheme` to generate a Kubernetes client for the webhook. Unfortunately, that broke reading machine spec values from secret or configmap references (#1142), as the client did not recognise `v1.Secret` or `v1.ConfigMap` resources anymore. We're using the default scheme now and only add OSM APIs to it.

Because https://github.com/kubermatic/machine-controller/blob/86b1621658bb2e4a38ce3ff3c7c8d0c621aff1b0/pkg/providerconfig/types.go#L109-L117 does not throw or log the error from `cvr.GetConfigVarStringValue`, I found this only by running a debugger and reading the `err` variable. We might want to adjust that as well, but this PR focuses on fixing the webhook.

2. The client request (which was broken as per above) needs a context. But the passed context was always `nil`, because it came from an `admissionData` struct, for which the `ctx` field was never set. I honestly have no idea how the webhook worked at all before #1136 broke it. But this PR passes the `http.Request`'s context down to the `ConfigVarResolver`, which means that canceling the HTTP request for the admission request that is being processed and is using `ConfigVarResolver` will cancel downstream Kubernetes client calls too.

This problem was only reported for the Hetzner token, but I expect it to impact other providers too, where an unexpected empty value did not necessarily fail validation but validated the spec in a different way than the controller would use it.

Fixes #1142

**Optional Release Note**:

```release-note
Fix unexpected empty values when reading Secret or ConfigMap references in the webhook
```

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -25,6 +25,7 @@ import (
 	userdatamanager "github.com/kubermatic/machine-controller/pkg/userdata/manager"
 	osmv1alpha1 "k8c.io/operating-system-manager/pkg/crd/osm/v1alpha1"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog"
@@ -77,6 +78,11 @@ func main() {
 	}
 
 	scheme := runtime.NewScheme()
+
+	if err := corev1.AddToScheme(scheme); err != nil {
+		klog.Fatalf("failed to add corev1 api to scheme: %v", err)
+	}
+
 	if err := osmv1alpha1.AddToScheme(scheme); err != nil {
 		klog.Fatalf("failed to add osmv1alpha1 api to scheme: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


ahmedwaleedmalik (approved):
Thank you for taking care of this.
/approve
/lgtm

kron4eg (approved):
/approve

ahmedwaleedmalik:
> I want to explore a suggestion by @kron4eg to not use a fresh scheme at all.

Okay, what advantage does that approach provide?

embik:
> > I want to explore a suggestion by @kron4eg to not use a fresh scheme at all.
> 
> Okay, what advantage does that approach provide?

It always gives us the schemes that Kubernetes comes with by default, afaik. It's basically a good baseline to work with, so we don't end up forgetting resource schemes we need (like with the v1 scheme in this example).

embik:
/hold

`pull-machine-controller-e2e-aws-spot-instance` is likely not succeeding any time soon. I've opened #1150 to move to a different AZ.

PR #1149, opened 2022-01-05
Add mutating logic to populate OSP annotation based on OS flavor

ahmedwaleedmalik:

This PR provides us a way to populate default OSP annotation in case it's missing. This is a requirement since OSM will only watch and cater resources that have this particular annotation.

Fixes #

Webhook is the best place to facilitate this mutation. Since in the future, we'll have an option to specify a custom value from the front-end/API but for defaulting webhook will be used.

**Optional Release Note**:

```release-note
NONE
```


xrstf (approved):
/approve

xrstf:

vCenter is currently down.

PR #1153, opened 2022-01-07
Split clients into worker and seed(default) clients for webhooks

ahmedwaleedmalik:

In KKP, these webhooks are deployed in a seed/master and worker(user) cluster architecture. Beforehand we were assuming that the only k8s client it needs is for the worker cluster since there was no need to interact with seed/master cluster. 

Although, after the addition of validations for OSM. There is a need to access resources, specifically OperatingSystemProfile, from the seed/master cluster.

To facilitate that we are splitting clients into:
1. Client - For seed/master, can use the in-cluster config
2. Worker Client - For worker cluster, needs out-of-cluster config

In case, the flag `worker-cluster-kubeconfig` is not specified then we assume that our subject is a single cluster and we use the same client as worker client.

Fixes #

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -31,58 +31,84 @@ import (
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-var (
-	masterURL              string
-	kubeconfig             string
-	admissionListenAddress string
-	admissionTLSCertPath   string
-	admissionTLSKeyPath    string
-	caBundleFile           string
-	useOSM                 bool
-	namespace              string
-)
+type options struct {
+	masterURL               string
+	kubeconfig              string
+	admissionListenAddress  string
+	admissionTLSCertPath    string
+	admissionTLSKeyPath     string
+	caBundleFile            string
+	useOSM                  bool
+	namespace               string
+	workerClusterKubeconfig string
+}
 
 func main() {
 	nodeFlags := node.NewFlags(flag.CommandLine)
+	opt := &options{}
 
 	klog.InitFlags(nil)
 	if flag.Lookup("kubeconfig") == nil {
-		flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+		flag.StringVar(&opt.kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
 	}
 	if flag.Lookup("master") == nil {
-		flag.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
+		flag.StringVar(&opt.masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	}
-	flag.StringVar(&admissionListenAddress, "listen-address", ":9876", "The address on which the MutatingWebhook will listen on")
-	flag.StringVar(&admissionTLSCertPath, "tls-cert-path", "/tmp/cert/cert.pem", "The path of the TLS cert for the MutatingWebhook")
-	flag.StringVar(&admissionTLSKeyPath, "tls-key-path", "/tmp/cert/key.pem", "The path of the TLS key for the MutatingWebhook")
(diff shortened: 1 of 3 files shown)


cmd/webhook/main.go
```diff
@@ -31,78 +31,104 @@ import (
...
-	flag.StringVar(&caBundleFile, "ca-bundle", "", "path to a file containing all PEM-encoded CA certificates (will be used instead of the host's certificates if set)")
-	flag.StringVar(&namespace, "namespace", "kubermatic", "The namespace where the webhooks will run")
+	flag.StringVar(&opt.admissionListenAddress, "listen-address", ":9876", "The address on which the MutatingWebhook will listen on")
+	flag.StringVar(&opt.admissionTLSCertPath, "tls-cert-path", "/tmp/cert/cert.pem", "The path of the TLS cert for the MutatingWebhook")
+	flag.StringVar(&opt.admissionTLSKeyPath, "tls-key-path", "/tmp/cert/key.pem", "The path of the TLS key for the MutatingWebhook")
+	flag.StringVar(&opt.caBundleFile, "ca-bundle", "", "path to a file containing all PEM-encoded CA certificates (will be used instead of the host's certificates if set)")
+	flag.StringVar(&opt.namespace, "namespace", "kubermatic", "The namespace where the webhooks will run")
+	flag.StringVar(&opt.workerClusterKubeconfig, "worker-cluster-kubeconfig", "", "Path to kubeconfig of worker/user cluster where machines and machinedeployments exist")
```
embik, line 62: Can we please add a note to the description saying that if not passed, will use the `--kubeconfig` kubeconfig? Basically, what is said in the PR description. I think otherwise  the output of `--help` is confusing (which kubeconfig do I need to set if I want to pass only one?)

cmd/webhook/main.go
```diff
@@ -31,78 +31,104 @@ import (
...
 	}
 
-	client, err := ctrlruntimeclient.New(cfg, ctrlruntimeclient.Options{})
+	cfg, err := clientcmd.BuildConfigFromFlags(opt.masterURL, opt.kubeconfig)
+	if err != nil {
+		klog.Fatalf("error building kubeconfig: %v", err)
+	}
+
+	client, err := ctrlruntimeclient.New(cfg, ctrlruntimeclient.Options{
+		Scheme: scheme.Scheme,
```
embik, line 88: This isn't needed, just leave `ctrlruntimeclient.Options{}` empty as it was before.

cmd/webhook/main.go
```diff
@@ -31,78 +31,104 @@ import (
...
+		workerClusterConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			&clientcmd.ClientConfigLoadingRules{ExplicitPath: opt.workerClusterKubeconfig},
+			&clientcmd.ConfigOverrides{}).ClientConfig()
+		if err != nil {
+			klog.Fatal(err)
+		}
+
+		// Build dedicated client for worker cluster
+		workerClient, err = ctrlruntimeclient.New(workerClusterConfig, ctrlruntimeclient.Options{
+			Scheme: scheme.Scheme,
```
embik, line 107: Same here, passing the standard scheme isn't necessary.

cmd/webhook/main.go
```diff
@@ -31,78 +31,104 @@ import (
...
 		klog.Fatalf("failed to build client: %v", err)
 	}
 
+	// Start with assuming that current cluster will be used as worker cluster
+	workerClient := client
+	// Handing for worker client
+	if opt.workerClusterKubeconfig != "" {
+		workerClusterConfig, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+			&clientcmd.ClientConfigLoadingRules{ExplicitPath: opt.workerClusterKubeconfig},
+			&clientcmd.ConfigOverrides{}).ClientConfig()
```
embik, line 100: Why is this needed? Is it because we want to prevent the code to load the kubeconfig from a standard location?

embik (changes requested):

cmd/webhook/main.go
ahmedwaleedmalik: Yes

cmd/webhook/main.go
ahmedwaleedmalik: Good point, updated

embik (approved):
/lgtm

PR #1154, opened 2022-01-10
Actually read environment variable in GetConfigVarBoolValueOrEnv

embik:
`GetConfigVarBoolValueOrEnv` never read the environment variable it was passed, as `stringVal` was always `"false"` when no value was passed (`configVar.Value` is a bool and defaults to `false` if not set), but the condition to read the env variable was `stringVal == ""`. This must be a pretty old bug as this code was introduced 4 years ago, but it was only ever used in `VSPHERE_ALLOW_INSECURE`, which is not in use in KKP for whatever reason.

This PR reworks the `ConfigVarBool` functions to make sure we can differentiate between an unset value and `false`.

This is needed as it also breaks `NUTANIX_ALLOW_INSECURE`, which is used in https://github.com/kubermatic/kubermatic/pull/8448.

Fixes #

**Optional Release Note**:

```release-note
Fix not reading bool configuration values from environment (VSPHERE_ALLOW_INSECURE and NUTANIX_INSECURE)
```

### DIFF
--- a/pkg/providerconfig/types.go
+++ b/pkg/providerconfig/types.go
@@ -135,13 +135,16 @@ func (cvr *ConfigVarResolver) GetConfigVarBoolValueOrEnv(configVar providerconfi
 	if err != nil {
 		return false, err
 	}
-	if stringVal == "" {
+
+	// if stringVar is "false", it might have been not set and we can attempt to read from environment.
+	// if it is "true", we know it was configured via value or secret reference.
+	if stringVal == "" || stringVal == strconv.FormatBool(false) {
 		envVal, envValFound := os.LookupEnv(envVarName)
-		if !envValFound {
-			return false, fmt.Errorf("all mechanisms(value, secret, configMap) of getting the value failed, including reading from environment variable = %s which was not set", envVarName)
+		if envValFound {
+			stringVal = envVal
 		}
-		stringVal = envVal
 	}
+
 	boolVal, err := strconv.ParseBool(stringVal)
 	if err != nil {
 		return false, err
--- a/pkg/providerconfig/types.go
+++ b/pkg/providerconfig/types.go
@@ -147,11 +147,11 @@ func (cvr *ConfigVarResolver) GetConfigVarBoolValue(configVar providerconfigtype
 		return false, false, fmt.Errorf("configmap '%s' in namespace '%s' has no key '%s'", configVar.ConfigMapKeyRef.Name, configVar.ConfigMapKeyRef.Namespace, configVar.ConfigMapKeyRef.Key)
 	}
 
-	if !configVar.Valid {
+	if configVar.Value == nil {
 		return false, false, nil
 	}
 
-	return configVar.Value, true, nil
+	return configVar.Value != nil && *configVar.Value, true, nil
 }
 
 func (cvr *ConfigVarResolver) GetConfigVarBoolValueOrEnv(configVar providerconfigtypes.ConfigVarBool, envVarName string) (bool, error) {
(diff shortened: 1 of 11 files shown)


pkg/providerconfig/types.go
```diff
@@ -135,13 +135,16 @@ func (cvr *ConfigVarResolver) GetConfigVarBoolValueOrEnv(configVar providerconfi
 	if err != nil {
 		return false, err
 	}
-	if stringVal == "" {
+
+	// if stringVar is "false", it might have been not set and we can attempt to read from environment.
+	// if it is "true", we know it was configured via value or secret reference.
+	if stringVal == "" || stringVal == strconv.FormatBool(false) {
```
xmudrii, line 141: Shouldn't we use and here? If someone puts `false` in the ProviderSpec, and they have env var set to true, I think we should respect what's in the ProviderSpec. With this `if`, we would ignore `false` value and still try to read from env, which could override it.

kron4eg: Problems like this are usually solved with additional bool field which will clearly indicate ether the value has need set explicitly or just empty.
Example: https://pkg.go.dev/database/sql#NullBool

```
type NullBool struct {
	Bool  bool
	Valid bool // Valid is true if Bool is not NULL
}
```

`.Bool` field is the value, and .Valid is indicator if it was set/present/not-null.

We can do the same trick with `ConfigVarBool`, since it's employ custom json unmarshall method anyway.

embik: That sounds sensible, I'll look into it. Thanks for the suggestion!

embik: Done, please review the new implementation.

pkg/providerconfig/types.go
```diff
@@ -116,37 +116,60 @@ func (cvr *ConfigVarResolver) GetConfigVarStringValueOrEnv(configVar providercon
...
 }
 
-func (cvr *ConfigVarResolver) GetConfigVarBoolValue(configVar providerconfigtypes.ConfigVarBool) (bool, error) {
-	cvs := providerconfigtypes.ConfigVarString{Value: strconv.FormatBool(configVar.Value), SecretKeyRef: configVar.SecretKeyRef}
-	stringVal, err := cvr.GetConfigVarStringValue(cvs)
-	if err != nil {
-		return false, err
+// GetConfigVarBoolValue returns a boolean from a ConfigVarBool. If there is no valid source for the boolean,
+// the second bool returned will be false (to be able to differentiate between "false" and "unset")
+func (cvr *ConfigVarResolver) GetConfigVarBoolValue(configVar providerconfigtypes.ConfigVarBool) (bool, bool, error) {
```
xmudrii, line 121: While the comment describes what's each return value, I think it might also make sense to name those return values. For example (might not be fully-correct):

```suggestion
func (cvr *ConfigVarResolver) GetConfigVarBoolValue(configVar providerconfigtypes.ConfigVarBool) (val bool, valid bool, err error) {
```

xmudrii (approved):
Other than the return value names, LGTM. This can also be fixed later.

/lgtm
/approve

ahmedwaleedmalik (approved):
/approve

pkg/providerconfig/types/types.go
```diff
@@ -221,7 +220,7 @@ func (configVarString *ConfigVarString) UnmarshalJSON(b []byte) error {
 }
 
 type ConfigVarBool struct {
-	Value           bool                       `json:"value,omitempty"`
+	Value           *bool                      `json:"value,omitempty"`
```
kron4eg, line 223: I personally really don't like this approach. It lead to this clunky (bool, bool, error) function signature which then the middle boolean is proceeded to be ignored.

The way it's done in the stdlib (and not only there) is much more pleasant, to my taste.

embik: I tried with a `Valid bool` approach, but that didn't work out well with the marshalling and unmarshalling tests. Maybe I misunderstood your suggestion.

IMHO, the function signature isn't _that_ clunky. It's additional information that can be used if needed, it just happens to be not needed in most calls (since the provider config parsing doesn't care if the value is `false` because it's set or unset; `false` is what it defaults to anyway).

PR #1161, opened 2022-01-13
Install and enable iscsid on Nutanix machines

xmudrii:

This PR changes the userdata so the following tasks are done on Nutanix machines:

* the `open-iscsi` package is installed on Ubuntu
* the `iscsi-initiator-utils` package is installed on CentOS and RHEL
* the `iscsid` service is enabled on Ubuntu/CentOS/RHEL

This is required for the Nutanix CSI driver to attach volumes.

**Optional Release Note**:
```release-note
Install and enable iscsid on Nutanix machines
```

/assign @embik @moadqassem @ahmedwaleedmalik 
/hold
to test it manually

### DIFF
--- a/pkg/userdata/centos/provider_test.go
+++ b/pkg/userdata/centos/provider_test.go
@@ -193,6 +193,16 @@ func TestUserDataGeneration(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "kubelet-v1.21-nutanix",
+			spec: clusterv1alpha1.MachineSpec{
+				ObjectMeta: metav1.ObjectMeta{Name: "node1"},
+				Versions: clusterv1alpha1.MachineVersionInfo{
+					Kubelet: "1.21.5",
+				},
+			},
+			cloudProviderName: stringPtr("nutanix"),
+		},
 	}
 
 	defaultCloudProvider := &fakeCloudConfigProvider{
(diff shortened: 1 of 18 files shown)


pkg/userdata/centos/provider_test.go
```diff
@@ -193,6 +193,16 @@ func TestUserDataGeneration(t *testing.T) {
...
 		},
+		{
+			name: "kubelet-v1.21-nutanix",
+			spec: clusterv1alpha1.MachineSpec{
+				ObjectMeta: metav1.ObjectMeta{Name: "node1"},
+				Versions: clusterv1alpha1.MachineVersionInfo{
+					Kubelet: "1.21.5",
+				},
+			},
+			cloudProviderName: stringPtr("nutanix"),
```
embik, line 204: Nit: should/could we use https://github.com/kubermatic/machine-controller/blob/17474051fe67702a33a453ff1883b6008983988f/pkg/providerconfig/types/types.go#L56 here instead?

embik: Seems to be done that way for other test cases too. Resolving.

embik (approved):

ahmedwaleedmalik (approved):
/approve

embik (approved):

PR #1162, opened 2022-01-13
Differentiate CloudProviderName and KubeletCloudProviderName in templates

xmudrii:

We currently use the `CloudProviderName` field in templates for:

* the `--cloud-provider` flag in the Kubelet service
* running commands only for some provider(s)

However, `CloudProviderName` contains the value returned by `GetCloudConfig` function. This function returns the cloud provider name used by Kubelet, and only if the provider has an in-tree cloud provider implementation (in-tree CCM). In other words, `CloudProviderName` will be empty on providers such as Nutanix, DigitalOcean, Hetzner, etc...

We encountered the issue when we tried to install some packages only on Nutanix machines in #1161. Because Nutanix doesn't have in-tree CCM, `.CloudProviderName` is empty, and the Nutanix-specific logic is not executed.  Therefore, this PR is a prerequisite for #1161.

This issue is fixed by differentiating `CloudProviderName` and `KubeletCloudProviderName`:

* `CloudProviderName` is `providerConfig.CloudProvider`. It's used in templates when we want to run cloud-provider-specific logic (e.g. install some packages only for one provider)
* `KubeletCloudProviderName` is from `GetCloudConfig). It's passed to Kubelet via `--cloud-provider` flag

**Optional Release Note**:
```release-note
Differentiate CloudProviderName and KubeletCloudProviderName in templates
```

/assign @embik @moadqassem @ahmedwaleedmalik

### DIFF
--- a/pkg/userdata/flatcar/provider_test.go
+++ b/pkg/userdata/flatcar/provider_test.go
@@ -421,16 +421,17 @@ func TestUserDataGeneration(t *testing.T) {
 			}
 
 			req := plugin.UserDataRequest{
-				MachineSpec:           test.spec,
-				Kubeconfig:            kubeconfig,
-				CloudConfig:           cloudConfig,
-				CloudProviderName:     cloudProviderName,
-				DNSIPs:                test.DNSIPs,
-				ExternalCloudProvider: test.externalCloudProvider,
-				HTTPProxy:             test.httpProxy,
-				NoProxy:               test.noProxy,
-				PauseImage:            test.pauseImage,
-				KubeletFeatureGates:   kubeletFeatureGates,
+				MachineSpec:              test.spec,
+				Kubeconfig:               kubeconfig,
+				CloudConfig:              cloudConfig,
+				CloudProviderName:        cloudProviderName,
+				KubeletCloudProviderName: cloudProviderName,
+				DNSIPs:                   test.DNSIPs,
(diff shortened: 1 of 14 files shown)


pkg/userdata/flatcar/provider_test.go
```diff
@@ -102,7 +101,7 @@ func (p *fakeCloudConfigProvider) GetCloudConfig(spec clusterv1alpha1.MachineSpe
 type userDataTestCase struct {
 	name                  string
 	spec                  clusterv1alpha1.MachineSpec
-	ccProvider            cloud.ConfigProvider
+	ccProvider            *fakeCloudConfigProvider
```
embik, line 104: I'm not sure I'm a fan of this. Doesn't this tell us that the `ConfigProvider` interface is missing something, like `func (p *ConfigProvider) Name() string`?

xmudrii: This is correct — having the `Name()` function in the `CloudProvider` interface would be helpful. However, that requires changing all cloud providers, and I don't have capacity for that right now. This is handled in a similar way for other providers where we just provide a string pointer and then based on it create `fakeCloudConfigProvider`.

xmudrii: Turns out this change was not needed at all. I reverted it.

ahmedwaleedmalik (approved):

PR #1169, opened 2022-01-18
Configure server group of the openstack VM

### LINKED ISSUE #1167: Add servergroup for Openstack Nodes
Openstack put the (anti-)affinity feature in the server group and not in the VM resource.

In our setup the Scheduler returns a valid Hypervisor to any new machine and we end up to often with more then one Worker on one Hypervisor, which is bad for reliability.

The idea would be to create a Servergroup for every machine deployment and add a anti-affinity(affinity is the default) option to the cloudprovider spec.

That way we should have a better balancing of the machines and can keep the default, for older installations

kron4eg:
Fixes #1167

**Optional Release Note**:

```release-note
Configure server group of the openstack VM
```


ahmedwaleedmalik (approved):
/lgtm
/approve

xrstf:

vCenter is offline right now.

PR #1174, opened 2022-01-25
Improve generation for container runtime configuration

ahmedwaleedmalik:

Currently, the configuration seems a bit scattered. This PR aims to bring parsing, validation, and manipulation of `flags` used for container runtime configuration into a single more visible place.

Another major reason for this refactoring is that we can now consume these methods directly in OSM instead of duplicating everything there as well.

Fixes #

**Optional Release Note**:

```release-note
NONE
```


pkg/userdata/ubuntu/testdata/containerd.yaml
```diff
@@ -363,8 +363,6 @@ write_files:
     [plugins."io.containerd.grpc.v1.cri".registry]
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."docker.io"]
-    endpoint = ["https://registry-1.docker.io"]
-    [plugins."io.containerd.grpc.v1.cri".registry.mirrors."k8s.gcr.io"]
```
kron4eg, line 367: no, this is incorrect

ahmedwaleedmalik: Fixed :)

cmd/machine-controller/main.go
```diff
@@ -237,37 +236,17 @@ func main() {
...
-
-	if len(registryMirrors) > 0 {
-		nodeContainerdRegistryMirrors["docker.io"] = registryMirrors
-	}
-
-	if nodeRegistryCredentialsSecret != "" {
-		if secRef := strings.Split(nodeRegistryCredentialsSecret, "/"); len(secRef) != 2 {
-			klog.Fatalf("-node-registry-credentials-secret is in incorrect format %q, should be in 'namespace/secretname'", nodeRegistryCredentialsSecret)
-		}
+	containerRuntimeConfig, err := containerruntime.GenerateContainerRuntimeConfig(containerRuntimeOpts)
```
kron4eg, line 247: usually functions like `GenerateContainerRuntimeConfig` are called BuildSomething. in our case it can be `BuildConfig()`. and the word `containerdruntime` is already part of the type identity since it's in the `containerruntime` package, meaning we can skip repeating ourselves with additional `ContainerRuntime`.

kron4eg (commented):
Looks very good!

Small naming adjustments would be good.

cmd/machine-controller/main.go
ahmedwaleedmalik: Great point, let me fix that right away.

kron4eg (approved):
:shipit:

PR #1176, opened 2022-01-26
Enable SeccompDefault setting if kubelet feature gate is passed

embik:

[SeccompDefault](https://kubernetes.io/docs/tutorials/security/seccomp/#enable-the-use-of-runtimedefault-as-the-default-seccomp-profile-for-all-workloads) is a useful kubelet feature gate to isolate every container with seccomp by default. Since it's an alpha feature, it needs a kubelet feature gate _and_ a kubelet config value. This PR sets the config value if the feature gate is passed.

Fixes #

**Optional Release Note**:

```release-note
Enable SeccompDefault setting if kubelet feature gate `SeccompDefault` is passed
```

### DIFF
--- a/pkg/userdata/helper/kubelet.go
+++ b/pkg/userdata/helper/kubelet.go
@@ -235,6 +235,10 @@ func kubeletConfiguration(clusterDomain string, clusterDNS []net.IP, featureGate
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/lgtm
/approve

PR #1184, opened 2022-02-04
Introduce Hetzner placement groups into machinedeployments

### LINKED ISSUE #1120: Hetzner pass placement group in cloudProviderSpec
Add support for [placement groups](https://docs.hetzner.cloud/#placement-groups) ID passthrough via providerSpec on `machine.spec.providerConfig.cloudProviderSpec.placementGroup`.

ol-iver:
Signed-off-by: ol-iver <scarface@mywoh.de>

This PR enables machine-controller to use [Hetzner's placement groups](https://docs.hetzner.com/cloud/placement-groups/overview) for machine controller.
The documentation suggests that there is a maximum number of 10 virtual machines per placement group, so we should ensure that this limit does not prevent machine-controller from creating new worker nodes.
A manual definition of placement groups looks to cumbersome and error-prone to me, especially in an autoscaling case.
That's why I decided to define a placement group prefix and letting machine-controller create the placement groups rather than maintaining concrete groups manually.

These are the key facts:
- there is a new option `placementGroupPrefix` in Hetzner provider config
- it does not need to be verified, because you can enter pretty much everything as a placement group name, including things like leading `-`
- the placement groups naming is looking like this `<PLACEMENT_GROUP_PREFIX>-<random 5 letter string>` 
- when new servers are created, machine controller always try to populate existing placement groups before creating new one
- empty placement groups (after scaling down) are deleted automatically

Fixes #1120 

Follow up of my Hetzner Terraform [PR@kubeone](https://github.com/kubermatic/kubeone/pull/1762)

**Optional Release Note**:

```release-note
Machine-deployment for Hetzner now includes the option to use the placement group functionality to ensure that virtual machines are spread accross different physical machines.
It can be enabled by setting "placementGroupPrefix" option in Hetzner's provider config. The placement groups are created automatically by machine-controller.
```


ol-iver:
I was not too happy with the idea of leaving orphan placement groups behind. 
That's why I added a cleanup function which deletes placement groups which are going to be empty when a virtual machine is deleted.
The naming convention of the placement groups is different now, because of the changes: `<PLACEMENT_GROUP_PREFIX>-<random 5 letter string>` 

Btw. I tested it on my Hetzner account. I built a [docker image](https://github.com/users/ol-iver/packages/container/package/machine-controller%2Fmachine-controller) you could use to replace your default images of `machine-controller` and `machine-controller-webhook`, if you would like to run a quick test
`ghcr.io/ol-iver/machine-controller/machine-controller:dev`

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -267,6 +275,41 @@ func (p *provider) Create(machine *v1alpha1.Machine, _ *cloudprovidertypes.Provi
...
+				Type:   hcloud.PlacementGroupTypeSpread,
+			})
+			if err != nil {
+				return nil, hzErrorToTerminalError(err, "failed to create placement group")
+			}
+			n = p.PlacementGroup
+		}
+		serverCreateOpts.PlacementGroup = n
+	}
+
```
kron4eg, line 312: There are bunch of one letter variables with places of use far then 1-3 lines after declaration. Please give them proper names, it's hard to read (I know, we use them too and I'm not proud of it, but it's legacy, at least let's not increase that problem in the new code)

pkg/cloudprovider/provider/hetzner/provider.go
```diff
@@ -376,6 +419,26 @@ func (p *provider) Cleanup(machine *v1alpha1.Machine, data *cloudprovidertypes.P
...
 	}
+
+	if instance.(*hetznerServer).server.PlacementGroup != nil {
+		pg, _, err := client.PlacementGroup.Get(ctx, instance.(*hetznerServer).server.PlacementGroup.Name)
+		if err != nil {
+			return false, hzErrorToTerminalError(err, "failed to get placement group")
+		}
+		count := 0
+		for _, s := range pg.Servers {
+			if s != instance.(*hetznerServer).server.ID {
```
kron4eg, line 430: so many `instance.(*hetznerServer)` :) you can save it into the var btw.

```go
hzServer := instance.(*hetznerServer)
hzServer.this...
hzServer.that...
passItDownTo(hzServer)
```

kron4eg (changes requested):
Please fix the linter complains

pkg/cloudprovider/provider/hetzner/provider.go
ol-iver: hehe ok, I thought the short variables are your coding style 😄

pkg/cloudprovider/provider/hetzner/provider.go
ol-iver: done

ol-iver:
Btw. the /retest function is not working as @kubermatic-bot suggests

ol-iver:
Ok, I moved the determination of the placement group in an own method, but even then the complexity was too high (31)
Thus, I removed some unneeded `if len(xyz) > 0` around loops and made it consistant through the file.

make lint is now running successfully on my machine.

kron4eg (approved):

kron4eg:
/assign @moadqassem 
Please take a look

PR #1185, opened 2022-02-09
Sets kubelet ContainerLog settings when docker is used

WeirdMachine:
Sets kubelet ContainerLog settings when docker is used, although it has no effect.
This will prevent the default settings to be applied to kubelet where docker has different settings.
Ensures consistency to KubeOne which does the same for control-plane nodes.

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve
/lgtm

PR #1187, opened 2022-02-12
Remove cluster field from vSphere provider

moadqassem:
The cluster id field for vSphere cloud provider, isn't used anywhere except in the labels of machines. It used to be part of the cloud-config however, it seems that it is not the case anymore. Thus we will remove it since it is not used by the provider. 
Fixes #

**Optional Release Note**:

```release-note
Remove cluster-id from vsphere cloud provider 
```


xmudrii:
/hold
for discussion

xmudrii (approved):

PR #1188, opened 2022-02-16
fix: returned vSphere RawConfig.Cluster

kron4eg:
Removal of this field is a backward compatibility issue and will break all the vSphere machineDeployment objects that had .Cluster in their ProviderSpec.

**Optional Release Note**:

```release-note
fix: returned vSphere RawConfig.Cluster
```


moadqassem (approved):
/approve
/lgtm

PR #1190, opened 2022-02-16
Waiting for volumeAttachments deletion

### LINKED ISSUE #1189: Wait volumeAttachments deletion before deleting nodes
When using vSphere as a cloud provider with the external CSI driver, the node rollout leads to the following problem: the CSI driver doesn't have enough time to delete the existing `volumeAttachments` related to the old node before it is deleted; they hang forever and are not collected anymore by the CSI driver because the old node doesn't exist anymore. To solve this issue, the machine controller should wait for the existing `volumeAttachments` deletion before deleting the node.

mlavacca:
When a machine is deleted, the machine controller now waits for the volumeAttachments deletion before deleting the node.

Fixes #1189 

**Optional Release Note**:

```release-note
NONE
```


moadqassem:
@mlavacca So I am not sure I completely follow here but I have two questions:
1- It seems that the issue is related to the vSphere not the other cloud providers, thus why you are doing this on a global level(machine level) ?
2- If the node got deleted before the volume what is the error that you get actually? Because this is quite strange and an issue in the CSI driver itself not machine controller. Usually when a node doesn't exist, CCM/CSI or whatever controller should delete its resources. IIRC, it was possible to delete the volumeAttachment objects and then the PVs are released, PVCs get deleted, via a ticker service.

In general I don't think it is the job for machine controller to remove volumes, that's why we use CSI driver :-) . If this is indeed an issue in the driver, I would strongly recommend to leave a finalizer on the machine object, during the cleanup, check for the node name under the machine status and remove the volumeAttachment. and then remove the finalizer off the machine.

moadqassem:
@mlavacca This seems quite strange and I have two questions:
1- It seems that the issue is related to the vSphere not the other cloud providers, thus why you are doing this on a global level(machine level) ?
2- If the node got deleted before the volume what is the error that you get actually? Because this is quite strange and an issue in the CSI driver itself not machine controller. Usually when a node doesn't exist, CCM/CSI or whatever controller should delete its resources. IIRC, it was possible to delete the volumeAttachment objects and then the PVs are released, PVCs get deleted, via a ticker service.

In general I don't think it is the job for machine controller to remove volumes, that's why we use CSI driver :-) . If this is indeed an issue in the driver, I would strongly recommend to leave a finalizer on the machine object, during the cleanup, check for the node name under the machine status and remove the volumeAttachment. and then remove the finalizer off the machine.

mlavacca:
@moadqassem 
> 2- If the node got deleted before the volume what is the error that you get actually? Because this is quite strange and an issue in the CSI driver itself not machine controller.

Yes, there is an issue in the CSI driver for vSphere, and the aiming of this PR is to mitigate that. If the node is deleted before that CSI driver has time to delete the `volumeAttachmentes`, they will hang forever, and the new pods that will try to mount the associated volume won't be able because of the existence of the outdated `volumeAttachments`.

> In general I don't think it is the job for machine controller to remove volumes, that's why we use CSI driver :-) . If this is indeed an issue in the driver, I would strongly recommend to leave a finalizer on the machine object, during the cleanup, check for the node name under the machine status and remove the volumeAttachment. and then remove the finalizer off the machine.

I'm not removing any `volumeAttachment` manually, this code introduces await to be sure that all the volumeAttachments are correctly deleted by the CSI driver. For this reason, I do not see why it should be a problem having this behavior for all the cloud providers.

BTW, something didn't work as expected during the e2e tests, I'm going to investigate and debug them

moadqassem:
> @moadqassem
> 
> > 2- If the node got deleted before the volume what is the error that you get actually? Because this is quite strange and an issue in the CSI driver itself not machine controller.
> 
> Yes, there is an issue in the CSI driver for vSphere, and the aiming of this PR is to mitigate that. If the node is deleted before that CSI driver has time to delete the `volumeAttachmentes`, they will hang forever, and the new pods that will try to mount the associated volume won't be able because of the existence of the outdated `volumeAttachments`.
> 
> > In general I don't think it is the job for machine controller to remove volumes, that's why we use CSI driver :-) . If this is indeed an issue in the driver, I would strongly recommend to leave a finalizer on the machine object, during the cleanup, check for the node name under the machine status and remove the volumeAttachment. and then remove the finalizer off the machine.
> 
> I'm not removing any `volumeAttachment`, this code introduces a wait to be sure that all the volumeAttachments are correctly deleted by the CSI driver. For this reason, I do not see why it should be a problem having this behavior for all the cloud providers.
> 
> BTW, something didn't work as expected during the e2e tests, I'm going to investigate and debug them

> I'm not removing any volumeAttachment
That's the thing. I mean if we. can just simply remove those as part of the cleanup process. So in other words, when cleanup is called, check out the volume attachment, and remove them based on the machine referenced node. This way you will not block the node draining/termination and you keep the mitigation locally where it should be until the vmware folks fix the issue is resolved:

https://github.com/kubernetes-sigs/vsphere-csi-driver/issues/359

P.S: it was created almost 2 years ago, but was active again 3 weeks ago so 🤞

mlavacca:
> > I'm not removing any volumeAttachment
> 
> That's the thing. I mean if we. can just simply remove those as part of the cleanup process. So in other words, when cleanup is called, check out the volume attachment, and remove them based on the machine referenced node. This way you will not block the node draining/termination and you keep the mitigation locally where it should be until the vmware folks fix the issue is resolved:
> 
> [kubernetes-sigs/vsphere-csi-driver#359](https://github.com/kubernetes-sigs/vsphere-csi-driver/issues/359)
> 
> P.S: it was created almost 2 years ago, but was active again 3 weeks ago so 🤞

Problem is that the VM attachments are managed by the CSI driver; if you delete them, they are automatically recreated. Furthermore, they have a finalizer (`external-attacher/csi-vsphere-vmware-com`) that is added/removed by the CSI driver, therefore I don't see a correct way to manually clean them up.

pkg/controller/machine/machine_controller.go
```diff
@@ -623,39 +624,53 @@ func (r *Reconciler) deleteCloudProviderInstance(prov cloudprovidertypes.Provide
 	})
 }
 
-func (r *Reconciler) deleteNodeForMachine(ctx context.Context, machine *clusterv1alpha1.Machine) error {
+func (r *Reconciler) deleteNodeForMachine(ctx context.Context, machine *clusterv1alpha1.Machine) (*reconcile.Result, error) {
+	// List all the volumeAttachments in the cluster; we must be sure that all
+	// of them will be deleted before deleting the node
+	volumeAttachments := &storagev1.VolumeAttachmentList{}
```
moadqassem, line 630: Can we just abstract this behaviour in a method and only run it if the cloud provider is vSphere.

pkg/controller/machine/machine_controller.go
```diff
@@ -550,7 +551,7 @@ func (r *Reconciler) deleteMachine(ctx context.Context, prov cloudprovidertypes.
 		return nil, nil
 	}
 
-	return nil, r.deleteNodeForMachine(ctx, machine)
+	return r.deleteNodeForMachine(ctx, machine)
```
moadqassem, line 554: Just call this method here when the cloud provider is vsphere and the Volume attachment are gone.

mlavacca:
@moadqassem I implemented your suggested solution, PTAL

mlavacca:
Code has been improved to handle both the node rollout and the node deletion on a single node cluster. @moadqassem PTAL

pkg/controller/machine/machine_controller.go
```diff
@@ -116,6 +117,7 @@ type Reconciler struct {
 	satelliteSubscriptionManager     rhsm.SatelliteSubscriptionManager
 
 	useOSM        bool
+	externalCSI   bool
```
moadqassem, line 120: Why this needed? If you would like to know if the external cluster is used then you can check the annotation on the machine: 
```v1.kubelet-flags.machine-controller.kubermatic.io/ExternalCloudProvider: "true"```

File: pkg/node/poddeletion/pod_deletion.go [hunk 4]
```diff
@@ -0,0 +1,267 @@
...
+	"k8s.io/client-go/util/retry"
+	"k8s.io/klog"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	ErrorQueueLen = 1000
+)
+
+type NodeVolumeAttachmentsCleanup struct {
```
moadqassem, line 40: I thought that we are not gonna need this and we will just change how the node is drained. For example if there is a volume still attached then don't remove the csi driver pod and once no volumeattachment is there then remove the pod.

mlavacca: Yes, and this is an implementation of that behavior. If there are volumeAttachments:
1. Cordon the old node
2. delete all pods using volumes attached to the old node
3. wait for the CSI-driver to collect the volumeAttachments
4. drain the old node.

pkg/controller/machine/machine_controller.go
mlavacca: I didn't consider the existence of that annotation, I'll use it.

pkg/controller/machine/machine_controller.go
```diff
@@ -521,22 +544,34 @@ func (r *Reconciler) shouldEvict(ctx context.Context, machine *clusterv1alpha1.M
 }
 
 // deleteMachine makes sure that an instance has gone in a series of steps.
-func (r *Reconciler) deleteMachine(ctx context.Context, prov cloudprovidertypes.Provider, machine *clusterv1alpha1.Machine) (*reconcile.Result, error) {
+func (r *Reconciler) deleteMachine(ctx context.Context, prov cloudprovidertypes.Provider, providerName providerconfigtypes.CloudProvider, machine *clusterv1alpha1.Machine) (*reconcile.Result, error) {
```
moadqassem, line 547: `providerName providerconfigtypes.CloudProvider` is not used anywhere

pkg/node/poddeletion/pod_deletion.go
moadqassem: Sure, but can't we just fix this in the NodeEviction.Run method? why don't we handle the case over there, instead of having it here. Eventually the code is kinda similar, only difference is, the deletion criteria.

moadqassem (approved):
/approve 
/lgtm

PR #1193, opened 2022-02-17
Improve nil pointer checks and consider VPC overlay networks for Nutanix

### LINKED ISSUE #1192: Support advanced network types in Nutanix implementation
Right now, our Nutanix implementation seems to assume that we use a VLAN-based network. Using a different network type (e.g. VPC overlay networking) makes the webhook fail with a nil pointer dereference. We should:

1.) Make sure that this no longer happens and that all pointers are checked correctly.
2.) Support other network types. It's not clear how complicated that is or if it's simple to implement once we figure out the incorrect assumption that the code is making.

embik:

Our Nutanix provider implementation assumed that all networks would have a cluster reference. That does not seem to be the case, as e.g. VPC overlay networks do not have one (also see https://github.com/nutanix/terraform-provider-nutanix/pull/328).

This PR includes that as a possible scenario when checking the cluster reference on a discovered subnet.

Fixes #1192

**Optional Release Note**:

```release-note
Support VPC overlay network types in Nutanix
```

### DIFF
--- a/pkg/cloudprovider/provider/nutanix/client.go
+++ b/pkg/cloudprovider/provider/nutanix/client.go
@@ -232,8 +232,11 @@ func getSubnetByName(client *ClientSet, name, clusterID string) (*nutanixv3.Subn
 	}
 
 	for _, subnet := range subnets.Entities {
-		if *subnet.Status.Name == name && *subnet.Status.ClusterReference.UUID == clusterID {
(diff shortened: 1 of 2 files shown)


embik:
/hold

I'd like to test this, but we do not have a VPC overlay network yet.

Issue #1197, opened 2022-02-25
Validate spot instances required specs 


moadqassem:
**User Story**
At the moment, users can create aws spot instances without setting a max price in the dashboard, api and machine deployment spec object. This would lead to errors coming from the aws api stating that, it is not possible to create aws spot instances without setting a max price.

**Acceptance Criteria**
- Machine controller webhook validates the machine deployment spot instance spec when spot instance is used 
- If the max price is not set for the spot instance return an error and don't create an MD.

PR #1198, opened 2022-02-25
Validate max price in spot instance request

moadqassem:
Makes sure that we validate the max price field of the machine deployment spec when the aws instance is marked as spot instance. 

Fixes kubermatic/machine-controller#1197

**Optional Release Note**:

```release-note
None
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -633,6 +633,12 @@ func (p *provider) Validate(spec clusterv1alpha1.MachineSpec) error {
 		return fmt.Errorf("failed to validate instance profile: %v", err)
 	}
 
+	if config.IsSpotInstance != nil && *config.IsSpotInstance {
+		if config.SpotMaxPrice == nil {
+			return errors.New("failed to validate max price for the spot instance: max price cannot be empty when spot instance ")
+		}
+	}
+
 	return nil
 }
 


pkg/cloudprovider/provider/aws/provider.go
```diff
@@ -633,6 +633,12 @@ func (p *provider) Validate(spec clusterv1alpha1.MachineSpec) error {
 		return fmt.Errorf("failed to validate instance profile: %v", err)
 	}
 
+	if config.IsSpotInstance != nil && *config.IsSpotInstance {
+		if config.SpotMaxPrice == nil {
```
ahmedwaleedmalik, line 637: I'm not sure about this change. In general, I'd prefer if we allow a `nil` value here. Since by default if `SpotMaxPrice` is not specified the current on demand price would be used.

moadqassem: AWS api doesn't allow it as you saw during the crd migration tests ;)

ahmedwaleedmalik: It didn't allow this because `undefined` was being passed as a value from the front-end.

kron4eg: yeah, what' the point of having nil SpotMaxPrice if the instance IsSpotInstance

kron4eg (approved):

pkg/cloudprovider/provider/aws/provider.go
moadqassem: Exactly, the reason why you need to set this up is, the max price will be the same as the on-demand machine price which makes kinda no sense.

moadqassem:
/unhold 

cc @kron4eg

PR #1200, opened 2022-03-01
Fix wrong CPU config

hdurand0710:

I got why changing in the MD ProviderSpec in the KKP Kubevirt e2e tests the CPU to 100m/1 did not work.
https://kubevirt.io/user-guide/virtual_machines/dedicated_cpu_resources/
Several issues:
- It has to be a whole number integer ([1-9]+)
- Users should not specify both spec.domain.cpu and spec.domain.resources.[requests/limits].cpu, which we did.
- We did hard-code `spec.domain.cpu=2`
- `Number of vCPUs is counted as sockets * cores * threads or if spec.domain.cpu is empty then it takes value from spec.domain.resources.requests.cpu or spec.domain.resources.limits.cpu`. As we did set `spec.domain.cpu` the value we did put in `spec.domain.resources.requests.cpu` was useless.

I removed the hard-coded  `spec.domain.cpu=2` to keep only the `spec.domain.resources.requests.cpu` (same value as `spec.domain.resources.limits.cpu`). 

Fixes #

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -365,9 +365,6 @@ func (p *provider) Create(machine *clusterv1alpha1.Machine, _ *cloudprovidertype
 				},
 				Spec: kubevirtv1.VirtualMachineInstanceSpec{
 					Domain: kubevirtv1.DomainSpec{
-						CPU: &kubevirtv1.CPU{
-							Cores: 2,
-						},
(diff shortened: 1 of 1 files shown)


moadqassem (approved):
/approve 
/lgtm 

@hdurand0710 I guess we need to consider CPU pinning in the future, because I believe there are some scenarios, where we would like to do that. For the time. being we can just  ignore this and let the k8s resources take over.

PR #1204, opened 2022-03-08
Create VM in dedicated namespace

hdurand0710:

Last piece needed in machine-controller to implement https://app.zenhub.com/workspaces/work-in-progress---team-caffeine-overflow-6166fa570192a800141b6db9/issues/kubermatic/kubermatic/8489.
After this PR, a last PR in KKP will be done.

Fixes kubermatic/kubermatic#9095

**Optional Release Note**:

```release-note
KubeVirt VMs are created in a dedicated namespace <cluster-id>
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -188,6 +189,19 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
 	return &config, pconfig, nil
 }
 
+// getNamespace returns the namespace where the VM is created.
+// VM is created in a dedicated namespace <cluster-id>
+// which is the namespace where the machine-controller pod is running.
+// Defaults to `kube-system`.
+func getNamespace() string {
+	ns := os.Getenv("POD_NAMESPACE")
+	if ns == "" {
+		// Useful especially for ci tests.
+		ns = metav1.NamespaceSystem
+	}
+	return ns
+}
+
 func (p *provider) Get(machine *clusterv1alpha1.Machine, _ *cloudprovidertypes.ProviderData) (instance.Instance, error) {
 	c, _, err := p.getConfig(machine.Spec.ProviderSpec)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -188,6 +189,19 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
...
 
+// getNamespace returns the namespace where the VM is created.
+// VM is created in a dedicated namespace <cluster-id>
+// which is the namespace where the machine-controller pod is running.
+// Defaults to `kube-system`.
+func getNamespace() string {
+	ns := os.Getenv("POD_NAMESPACE")
+	if ns == "" {
+		// Useful especially for ci tests.
+		ns = metav1.NamespaceSystem
```
xrstf, line 200: Why is this special case a special case? Can't the CI tests set a proper POD_NAMESPACE env?

hdurand0710: yes, they can.
I did hesitate between this version and setting the POD_NAMESPACE env.

My reasoning: if we add it in the ci, we also need to ask everybody to set this variable if they run the tests locally.
It's only a 'technical' variable, nothing to do with KubeVirt credentials.
Is it a good idea to ask everybody to add this env variable in their local env the running the tests ?
What do you think ?

xrstf: I see. I am still a bit unsure about how and where this code runs. Isn't the machine-controller running (in KKP) in the cluster namespace (`cluster-xxxxyyyy`)? Is the cluster-id just the `xxxxyyyy` part or the entire `cluster-xxxxyyyy`?

hdurand0710: The machine-controller runs in `cluster-xxxyy` namespace. So here POD_NAMESPACE is the `machine-controller` pod namespace (in machine-controller manifest env variable) (`cluster-xxxyy`):
```
    - name: POD_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
```

So, in the real life of the `machine-controller`, value of env variable POD_NAMESPACE would never be empty.
It's true that I'm here coding some default value that would never be used in real, but only used for local/ci tests, but avoiding people to set POD_NAMESPACE as a required env variable in their local env.

xrstf: 👍

hdurand0710:
Covered and included in https://github.com/kubermatic/machine-controller/pull/1143, closing this one.

PR #1207, opened 2022-03-11
consider pod cidr information during instance creation

PratikDeoghare:

Adds `pod-cidrs` command line flag to accept pod-cidrs information from command line.
Changes `provider.Create` method to use the information during machine instance creation. 

Implements part of [#9294](https://github.com/kubermatic/kubermatic/issues/9294)

**Optional Release Note**:
```release-note
NONE
```

### DIFF
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -166,7 +166,7 @@ func main() {
 	flag.Var(&nodeContainerdRegistryMirrors, "node-containerd-registry-mirrors", "Configure registry mirrors endpoints. Can be used multiple times to specify multiple mirrors")
 	flag.StringVar(&caBundleFile, "ca-bundle", "", "path to a file containing all PEM-encoded CA certificates (will be used instead of the host's certificates if set)")
 	flag.BoolVar(&nodeCSRApprover, "node-csr-approver", true, "Enable NodeCSRApprover controller to automatically approve node serving certificate requests")
-	flag.StringVar(&podCIDRs, "pod-cidrs", "172.25.0.0/16,fd00::/104", "The network ranges from which POD networks are allocated")
+	flag.StringVar(&podCIDRs, "pod-cidrs", "172.25.0.0/16", "The network ranges from which POD networks are allocated")
 	flag.StringVar(&nodePortRange, "node-port-range", "30000-32767", "A port range to reserve for services with NodePort visibility")
 	flag.StringVar(&nodeRegistryCredentialsSecret, "node-registry-credentials-secret", "", "A Secret object reference, that containt auth info for image registry in namespace/secret-name form, example: kube-system/registry-credentials. See doc at https://github.com/kubermaric/machine-controller/blob/master/docs/registry-authentication.md")
 	flag.BoolVar(&useOSM, "use-osm", false, "use osm controller for node bootstrap")
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -79,7 +79,7 @@ var (
 	nodeRegistryMirrors           string
 	nodePauseImage                string
 	nodeContainerRuntime          string
-	podCIDRs                      string
+	podCIDRs                      listFlag
 	nodePortRange                 string
 	nodeRegistryCredentialsSecret string
 	nodeContainerdRegistryMirrors = containerruntime.RegistryMirrorsFlags{}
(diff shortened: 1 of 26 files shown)


cmd/machine-controller/main.go
```diff
@@ -166,7 +166,7 @@ func main() {
 	flag.Var(&nodeContainerdRegistryMirrors, "node-containerd-registry-mirrors", "Configure registry mirrors endpoints. Can be used multiple times to specify multiple mirrors")
 	flag.StringVar(&caBundleFile, "ca-bundle", "", "path to a file containing all PEM-encoded CA certificates (will be used instead of the host's certificates if set)")
 	flag.BoolVar(&nodeCSRApprover, "node-csr-approver", true, "Enable NodeCSRApprover controller to automatically approve node serving certificate requests")
-	flag.StringVar(&podCidr, "pod-cidr", "172.25.0.0/16", "The network ranges from which POD networks are allocated")
+	flag.StringVar(&podCIDRs, "pod-cidrs", "172.25.0.0/16,fd00::/104", "The network ranges from which POD networks are allocated")
```
rastislavs, line 169: I would not change the default value here.

pkg/cloudprovider/types/types.go
```diff
@@ -76,6 +76,12 @@ type Provider interface {
 	SetMetricsForMachines(machines clusterv1alpha1.MachineList) error
 }
 
+// NetworkConfig holds information about cluster networking.
+// PodCIDRs fields is used to choose IPv4, IPv6 or dual-stack modes.
```
rastislavs, line 80: Move the `PodCIDRs` comment from here to one line above the `PodCIDRs` struct member.

cmd/machine-controller/main.go
```diff
@@ -166,7 +166,7 @@ func main() {
 	flag.Var(&nodeContainerdRegistryMirrors, "node-containerd-registry-mirrors", "Configure registry mirrors endpoints. Can be used multiple times to specify multiple mirrors")
 	flag.StringVar(&caBundleFile, "ca-bundle", "", "path to a file containing all PEM-encoded CA certificates (will be used instead of the host's certificates if set)")
 	flag.BoolVar(&nodeCSRApprover, "node-csr-approver", true, "Enable NodeCSRApprover controller to automatically approve node serving certificate requests")
-	flag.StringVar(&podCidr, "pod-cidr", "172.25.0.0/16", "The network ranges from which POD networks are allocated")
+	flag.StringVar(&podCIDRs, "pod-cidrs", "172.25.0.0/16", "The network ranges from which POD networks are allocated")
```
kron4eg, line 169: This is breaking change to flags, is it _really_ justified?

kron4eg: The way around is to have both, `-pod-cidr` and `-pod-cidrs` flags, and then in runtime merge their values if anything been provided.

kron4eg: OK, you're right, nether kubermatic nor kubeone uses it, so it's OK from my PoV.

PratikDeoghare: 👍

we can also pass same flag multiple times to get multiple values if it breaks anything at all

PratikDeoghare:
Weird. Bad rebase :grimacing: Fixing...

moadqassem (commented):
Please check the dependencies are imported correctly

pkg/cloudprovider/types/types.go
```diff
@@ -76,6 +76,11 @@ type Provider interface {
 	SetMetricsForMachines(machines clusterv1alpha1.MachineList) error
 }
 
+// NetworkConfig holds information about cluster networking.
+type NetworkConfig struct {
```
moadqassem, line 80: I am not sure why do we need this change here? I mean why do we need to specify this via command line? eventually what needs to be done is, deriving this from the machine deployment from here: 
https://github.com/kubermatic/machine-controller/blob/master/pkg/providerconfig/types/types.go#L108-L112

moadqassem (commented):
So in general, I guess it possible to specify who the CIDR should look like directly in the  machine created by KKP for example, as this is quite static when you add it directly in machine controller. Thus I would suggest that this could be on machine level.

pkg/cloudprovider/types/types.go
PratikDeoghare: [This]( https://github.com/kubermatic/machine-controller/blob/master/pkg/providerconfig/types/types.go#L108-L112) holds machine specific static network configuration. 

Machine doesn't need to know about pod cidrs. It only needs to have IPv6 address so that it can host pods with IPv6 addresses. So we decided not to put it in there.

moadqassem (approved):
/approve 
/lgtm

PR #1213, opened 2022-03-16
Add tag support for vSphere provider

### LINKED ISSUE #1209: Extend vSphere provider for tag creation
Validate and create vSphere tags

zreigz:
**What this PR does / why we need it**: Add tag support for vSphere provider

Fixes #1209

```release-note
Add tag support for vSphere provider
```

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/client.go
+++ b/pkg/cloudprovider/provider/vsphere/client.go
@@ -25,6 +25,7 @@ import (
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/session"
+	"github.com/vmware/govmomi/vapi/rest"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/soap"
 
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/provider/vsphere/client.go
```diff
@@ -85,3 +83,41 @@ func (s *Session) Logout() {
 		utilruntime.HandleError(fmt.Errorf("vsphere client failed to logout: %s", err))
 	}
 }
+
+func NewRESTSession(ctx context.Context, config *Config) (*RESTSession, error) {
```
moadqassem, line 87: So it is not possible to create and manage tags using the current client that we have? I thought they have refactored this and you can use the soap client directly without the rest client.

pkg/cloudprovider/provider/vsphere/helper.go
```diff
@@ -455,3 +456,47 @@ func resolveResourcePoolRef(ctx context.Context, config *Config, session *Sessio
...
+	defer restAPISession.Logout(ctx)
+	tagManager := tags.NewManager(restAPISession.Client)
+	klog.V(3).Info("Creating tags")
+	for _, tag := range config.Tags {
+		tagID, err := tagManager.CreateTag(ctx, &tag)
+		if err != nil {
+			return fmt.Errorf("failed to create tag: %v %v", tag, err)
+		}
+
+		if err := tagManager.AttachTag(ctx, tagID, vm.Reference()); err != nil {
```
moadqassem, line 474: Would it make sense to delete the tag if the attaching failed? I am looking at the way you delete the tags and what you do is you fetch all vm-referenced tags and then clean them, so if you failed to attach a tag here, then we will have orphan tags. 

I would suggest either those tags deleted here, or at the deletion stage, fetch the tags based on the machine not the vm and then clean them.

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -203,6 +213,28 @@ func (p *provider) Validate(spec clusterv1alpha1.MachineSpec) error {
...
+		restAPISession, err := NewRESTSession(ctx, config)
+		if err != nil {
+			return fmt.Errorf("failed to create REST API session: %v", err)
+		}
+		defer restAPISession.Logout(ctx)
+		tagManager := tags.NewManager(restAPISession.Client)
+		klog.V(3).Info("Found tags")
+		for _, tag := range config.Tags {
+			if tag.Name == "" {
+				return fmt.Errorf("one of the tag name is empty")
```
moadqassem, line 226: nit:
```suggestion
				return fmt.Errorf("one of the tags name is empty")
```

moadqassem (commented):
@zreigz Looks really cool, I left minor comments.

pkg/cloudprovider/provider/vsphere/provider.go
zreigz: done

pkg/cloudprovider/provider/vsphere/client.go
zreigz: All examples and tag occurrences were with the rest-client. I couldn't find anything with soap client

moadqassem (approved):
/approve
/lgtm

PR #1214, opened 2022-03-16
increase number of inotify instances per user

LittleFox94:

We often run into the problem of not being able to `kubectl logs -f` because of all allowed inotify instances being in use. This increases the instance limit (from a default of 128 we saw) to 8192, giving some more breathing space.

Corresponds to #471, but increases the maximum number of instances instead of watches, as was recommended in https://github.com/kubernetes/kubernetes/issues/10421 roughly a month after #471 was merged.

Instance limit of 8192 works great for us, but I have no calculation for how to choose this value.

**Optional Release Note**:

```release-note
Increase the inotify instance limit to 8192
```


LittleFox94:
> @LittleFox94 please run `go test ./pkg/userdata/... -update` and commit the results.

which do you like better: in a new commit or squashed into the original one?

kron4eg:
@LittleFox94 doesn't matter, in the end all commits will be squashed by the @kubermatic-bot (which is pushing merges).

LittleFox94:
I only see timeouts in the failed tests, is there anything I have to do @kron4eg ?

ahmedwaleedmalik (approved):
/approve
/lgtm

LittleFox94:
just randomly saw this is mentioned in the v1.46 release as being for our provider only - if anyone stumbles over this, this is, in fact, a provider-independent change

PR #1217, opened 2022-03-21
Kubevirt Add common label to all VMs linked to the same MachineDeployment

hdurand0710:

This PR adds a common label `machine: md-<machine-id>` to all VM spawned by the same MachineDeployment.
We already have in all `machine` this common label. Re-using it to add it to the `VirtualMachines`.
Also taking this PR opportunity to update the `example` for the Kubevirt MachineDeployment specification.

Fixes #

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -373,6 +374,10 @@ func (p *provider) Create(machine *clusterv1alpha1.Machine, data *cloudprovidert
 
 	resourceRequirements := kubevirtv1.ResourceRequirements{}
 	labels := map[string]string{"kubevirt.io/vm": machine.Name}
+	// Add a common label to all VirtualMachines spawned by the same MachineDeployment
+	if mdLabel, ok := machine.Labels["machine"]; ok {
+		labels["machine"] = mdLabel
```
moadqassem, line 379: What is the point here? I mean why you are adding this label on the vm ?

hdurand0710: To have a common label among VMs to all VMs created from the same MachineDeployment.
Needed for example for the affinity we discussed yesterday.

moadqassem: Yeah but the label that you are adding, isn't the machine deployment name :)

hdurand0710: @moath, as discussed, updated with the MachineDeployment name, rather than the `machine` label using the helper function.
Please review

moadqassem (approved):
/approve
/lgtm
/hold cancel

PR #1219, opened 2022-03-22
Flatcar stop disabled units

HaveFun83:

Fixes #

When `FlatcarConfig.DisableUpdateEngine` is set to true the update-engine.service is not stopped on new provisioned machines, which lead to an unwanted auto-updates to the latest flatcar version with a followed reboot. 
Only after this first reboot the service is then disabled.
The stop command stop the systemd service instantly to suppress the first auto-update on new provisioned machines.

**Optional Release Note**:

```release-note
fix disable auto-update when FlatcarConfig.DisableUpdateEngine is set to true
```

### DIFF
--- a/pkg/userdata/flatcar/provider.go
+++ b/pkg/userdata/flatcar/provider.go
@@ -450,10 +450,12 @@ coreos:
 {{- end }}
 {{- if .FlatcarConfig.DisableUpdateEngine }}
   - name: update-engine.service
+    command: stop
     mask: true
 {{- end }}
 {{- if .FlatcarConfig.DisableLocksmithD }}
   - name: locksmithd.service
+    command: stop
     mask: true
 {{- end }}
 {{- if .HTTPProxy }}
--- a/pkg/userdata/flatcar/provider.go
+++ b/pkg/userdata/flatcar/provider.go
@@ -184,12 +184,10 @@ systemd:
   units:
 {{- if .FlatcarConfig.DisableUpdateEngine }}
     - name: update-engine.service
-      command: stop
       mask: true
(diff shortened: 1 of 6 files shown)


HaveFun83:
@ahmedwaleedmalik is there a documented way to edit the flatcar/testdata/ignition* files? The stop command is missing here. I can edit but then the files are json beautified.

moadqassem:
> @ahmedwaleedmalik @moadqassem is there a documented way to edit the flatcar/testdata/ignition* files? The stop command is missing here. I can edit but then the files are json beautified.

@HaveFun83 Unfortunately not, but what you can do is, go to the ignition test and add the field `DisableUpdateEngine` here https://github.com/kubermatic/machine-controller/blob/master/pkg/userdata/flatcar/provider_test.go#L150. However, I am looking at the tests that are failing and I have the impression that, the tests are failing because of this newly introduced field. Are you sure that the field `command` is a valid ignition v1.20.1 field ?

ahmedwaleedmalik:
@HaveFun83 this doesn't seem to be a valid field for the ignition config https://github.com/coreos/ignition/blob/spec2x/config/v2_2/types/schema.go#L221

HaveFun83:
i removed the unsupported `command: stop` from the userDataIgnitionTemplate.

ahmedwaleedmalik:
This PR only provides a partial solution i.e. it only fixes it for `cloud-init` and not `ignition` provisioning. We'll go ahead with this and then work on fixing this for ignition later.

ahmedwaleedmalik (approved):
/approve

PR #1226, opened 2022-03-25
Disable firewalld on supported OSes

kron4eg:
Out of context (i.e. not kubernetes aware iptables operations) ip filtering is fundamentally incompatible with the Kubernetes. Network resources should be protect by the cloud security groups or network policies (or both!).

XRef: https://github.com/kubermatic/kubermatic/issues/8768

```release-note
Disable firewalld on supported OSes
```


ahmedwaleedmalik (approved):
/approve

PR #1241, opened 2022-04-05
Support rocky linux in azure cloud provider 

moadqassem:
Rocky linux support in azure
Fixes #

**Optional Release Note**:

```release-note
Support rockylinux in azure 
```


pkg/userdata/ubuntu/testdata/kubelet-version-without-v-prefix.yaml
```diff
@@ -191,6 +188,16 @@ write_files:
...
 
+- path: "/opt/disable-swap.sh"
+  permissions: "0755"
+  content: |
+    # Make sure we always disable swap - Otherwise the kubelet won't start as for some cloud
+    # providers swap gets enabled on reboot or after the setup script has finished executing.
+    cp /etc/fstab /etc/fstab.orig
+    cat /etc/fstab.orig | awk '$3 ~ /^swap$/ && $1 !~ /^#/ {$0="# commented out by cloudinit\n#"$0} 1' > /etc/fstab.noswap
+    mv /etc/fstab.noswap /etc/fstab
+    swapoff -a
```
xmudrii, line 199: Any specific reason why we changed this only for Ubuntu? Previously, it was only:

```
sed -i.orig '/.*swap.*/d' /etc/fstab
swapoff -a
```

moadqassem: That's actually wrong. This is a copy and paste mistake, will put it back to what it was.

xmudrii (approved):

PR #1246, opened 2022-04-07
KubeVirt restart kubelet at reboot

### LINKED ISSUE #1202: Kubevirt: user cluster node not re-joining the cluster after VM pod's deletion.
Issue opened after https://support.kubermatic.com/helpdesk/tickets/3654 investigation.
Scenario on dev with StorageClass=`local-path`

- Create a KubeVirt with several nodes, for example 3.
- Wait the the 3 nodes join the user cluster.
- Delete in the infra KubeVirt Cluster one of the VM pod.

What should happen:
- a new VM pod is spawned.
- the VM should join the cluster.

In the ticket raised by the customer, the VMs is spawned on a different infra node, which might also have other issues with the storage -> to investigate (having multiple infra nodes would be required to test).
On our infra cluster, there is only 1 node. Already with the VM re-scheduled on the same infra node, the VM does not re-join the cluster.

- User cluser
```
% k get node
NAME                                          STATUS     ROLES    AGE     VERSION
awesome-bose-worker-8b9rhh-85d98758c8-2td24   Ready      <none>   179m    v1.21.8
awesome-bose-worker-8b9rhh-85d98758c8-tcxld   Ready      <none>   3h14m   v1.21.8
awesome-bose-worker-8b9rhh-85d98758c8-xl9hm   NotReady   <none>   93m     v1.21.8
```

- Infra cluster:
```
% k get po -o wide
NAME                                                              READY   STATUS    RESTARTS   AGE     IP             NODE                           NOMINATED NODE   READINESS GATES
virt-launcher-awesome-bose-worker-8b9rhh-85d98758c8-2td24-nfz8c   1/1     Running   0          3h1m    10.244.0.202   ubuntu-2004-focal-64-minimal   <none>           <none>
virt-launcher-awesome-bose-worker-8b9rhh-85d98758c8-tcxld-rws9h   1/1     Running   0          3h15m   10.244.0.198   ubuntu-2004-focal-64-minimal   <none>           <none>
virt-launcher-awesome-bose-worker-8b9rhh-85d98758c8-xl9hm-vt5xk   1/1     Running   0          72m     10.244.0.237   ubuntu-2004-focal-64-minimal   <none>           <none>
virt-launcher-infallible-feynman-worker-qjdv4t-696854f455-tjf9w   1/1     Running   0          119m    10.244.0.225   ubuntu-2004-focal-64-minimal
```

- In the KCCM logs:
```
E0304 09:36:34.861185       1 node_controller.go:354] Specified Node IP not found in cloudprovider for node "awesome-bose-worker-8b9rhh-85d98758c8-xl9hm"
```

The user cluster node:
```
 k get node awesome-bose-worker-8b9rhh-85d98758c8-xl9hm -o yaml
apiVersion: v1
kind: Node
metadata:
  annotations:
    alpha.kubernetes.io/provided-node-ip: 10.244.0.236
```

`alpha.kubernetes.io/provided-node-ip: 10.244.0.236` is the old VM pod IP, not the new one.

Need to investigate why.

hdurand0710:

Part 2 to fix #1202. We read the configuration and restart the kubelet at each VM reboot. 

Fixes #1202

**Optional Release Note**:

```release-note
Kubevirt: reload the kubelet configuration and restart it at each VM reboot.
```


ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik:
/retest

Failed due to CPU quota limits on Azure

pkg/userdata/centos/provider.go
```diff
@@ -334,6 +337,42 @@ write_files:
...
+    set -xeuo pipefail
+
+    # This helps us avoid an unnecessary restart for kubelet on the first boot
+    if [ -f /etc/kubelet_needs_restart ]; then
+      # restart kubelet since it's not the first boot
+      systemctl daemon-reload
+      systemctl restart kubelet.service
+    else
+      touch /etc/kubelet_needs_restart
+    fi
```
embik, line 357: I'm just randomly stumbling into this PR but I'm curious, how is this script different from the systemd service `kubelet.service` being enabled in line 247 (https://github.com/kubermatic/machine-controller/pull/1246/files#diff-ad511f10cb3024c5aa13e897590c6bc85d1ba948341c05da6d3c6ac9dde196e0R247)? Shouldn't systemd start the service?

embik: My point is that the `kubelet` service isn't started, it's set up via `systemctl enable --now`:

https://github.com/kubermatic/machine-controller/blob/b6567d9a57c5d88984401367c2e952db828ac9e3/pkg/userdata/centos/provider.go#L247

That both starts and _enables_ the service. In systemd, "enabled" means it will start it at system startup/boot.

embik: Is it because the service fails after a reboot? That might mean we have a dependency issue in the service definition.

ahmedwaleedmalik: @hdurand0710 ^^

hdurand0710: it did not fail at reboot but was started with the wrong configuration. We needed to reload the configuration and restart.

embik: Sorry for being such a PITA, I just don't think I understand it yet. How can it run with the wrong configuration after the system just freshly booted up? Is there a systemd service that refreshes a config file that kubelet relies on? If yes, couldn't we just add a dependency for the kubelet service on that service when on kubevirt?

embik: The script reload systemd which makes me think that the kubelet.service systemd unit itself changed, but what is doing that?

hdurand0710: The configuration is updated in /opt/bin/setup_net_env.sh.
After reboot, this file was correct : /etc/systemd/system/kubelet.service.d/nodeip.conf
Nut the kubelet was running with the previous configuration.
So the idea was to reload the configuration and restart the kubelet.
If you think there a better way, let's go for it.

embik: Yeah okay, dealing with updated systemd units is painful. Thanks for explaining it to me, I really didn't get it. Sorry for wasting your time :-)

hdurand0710: Any feedback is never a waste of time. Always good! Thanks for taking the time to read this `systems` thingy. :-)

PR #1252, opened 2022-04-14
Expose google provisioning model to be able to acquire spot instances

tlamr:
Signed-off-by: Tomas Lamr <tomas@lamr.org>

Hello, we wanted to expose node selection API from google (SPOT/STANDARD) so that we can create a node pool of spot instances as in was.

Fixes #

**Optional Release Note**:

```release-note
Support for specifying provisioning model for instances on GCE
```

### DIFF
--- a/pkg/cloudprovider/provider/gce/config.go
+++ b/pkg/cloudprovider/provider/gce/config.go
@@ -101,7 +101,7 @@ type config struct {
 	network               string
 	subnetwork            string
 	preemptible           bool
-	automaticRestart      bool
+	automaticRestart      *bool
 	provisioningModel     string
 	labels                map[string]string
 	tags                  []string
@@ -169,15 +169,16 @@ func newConfig(resolver *providerconfig.ConfigVarResolver, spec v1alpha1.Provide
 		return nil, fmt.Errorf("cannot retrieve preemptible: %v", err)
 	}
 
-	cfg.automaticRestart = !cfg.preemptible
+	cfg.automaticRestart = nil
 	if cpSpec.AutomaticRestart != nil {
-		cfg.automaticRestart, _, err = resolver.GetConfigVarBoolValue(*cpSpec.AutomaticRestart)
+		var automaticRestart, _, err = resolver.GetConfigVarBoolValue(*cpSpec.AutomaticRestart)
 		if err != nil {
 			return nil, fmt.Errorf("cannot retrieve automaticRestart: %v", err)
 		}
+		cfg.automaticRestart = &automaticRestart
 	}
 
-	cfg.provisioningModel = "STANDARD"
+	cfg.provisioningModel = ""
 	if cpSpec.ProvisioningModel != nil {
 		cfg.provisioningModel, err = resolver.GetConfigVarStringValue(*cpSpec.ProvisioningModel)
 		if err != nil {
--- a/pkg/cloudprovider/provider/gce/config.go
+++ b/pkg/cloudprovider/provider/gce/config.go
@@ -102,7 +102,7 @@ type config struct {
 	subnetwork            string
 	preemptible           bool
 	automaticRestart      *bool
-	provisioningModel     string
+	provisioningModel     *string
 	labels                map[string]string
 	tags                  []string
 	jwtConfig             *jwt.Config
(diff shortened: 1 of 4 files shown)


tlamr:
Difference between preemptible and spot according to them:

However, Spot VMs provide new features that are not supported for preemptible VMs. For example, preemptible VMs can only run for up to 24 hours at a time, but Spot VMs do not have a maximum runtime. This page describes preemptible virtual machine (VM) instances.

ahmedwaleedmalik:
@tlamr can you please look into the failing CI jobs?

tlamr:
Hello @ahmedwaleedmalik, I'm not an expert but I can see 20x this error and then it seems to give up.
`Error: network limit reached (resource_limit_exceeded)

  with hcloud_network.net,
  on hetzner.tf line 10, in resource "hcloud_network" "net":
  10: resource "hcloud_network" "net" {
...
Creating cloud provider env failed!
Cleaning up machines.`

ahmedwaleedmalik:
> Error: network limit reached (resource_limit_exceeded)

This is unrelated to your PR. I'll take care of this issue. 

Meanwhile can you check the logs for https://public-prow.loodse.com/view/gs/prow-dev-public-data/pr-logs/pull/kubermatic_machine-controller/1252/pull-machine-controller-lint/1514635654829445120

pkg/cloudprovider/provider/gce/types/types.go
```diff
@@ -38,6 +38,8 @@ type CloudProviderSpec struct {
 	Network               providerconfigtypes.ConfigVarString `json:"network"`
 	Subnetwork            providerconfigtypes.ConfigVarString `json:"subnetwork"`
 	Preemptible           providerconfigtypes.ConfigVarBool   `json:"preemptible"`
+	AutomaticRestart      providerconfigtypes.ConfigVarBool   `json:"automaticRestart,omitempty"`
+	ProvisioningModel     providerconfigtypes.ConfigVarString `json:"provisioningModel,omitempty"`
```
ahmedwaleedmalik, line 42:
```suggestion
	AutomaticRestart      *providerconfigtypes.ConfigVarBool   `json:"automaticRestart,omitempty"`
	ProvisioningModel     *providerconfigtypes.ConfigVarString `json:"provisioningModel,omitempty"`
```

These should be optional values

ahmedwaleedmalik (changes requested):

ahmedwaleedmalik:
@tlamr I have updated the description of your PR and made it compliant with the required template.

tlamr:
@ahmedwaleedmalik Awesome, thank you! :)

pkg/cloudprovider/provider/gce/provider.go
```diff
@@ -265,7 +265,9 @@ func (p *Provider) Create(machine *clusterv1alpha1.Machine, data *cloudprovidert
 		Disks:             disks,
 		Labels:            labels,
 		Scheduling: &compute.Scheduling{
-			Preemptible: cfg.preemptible,
+			Preemptible:       cfg.preemptible,
+			AutomaticRestart:  &cfg.automaticRestart,
+			ProvisioningModel: cfg.provisioningModel,
```
ahmedwaleedmalik, line 270: Need to check for nil before assigning these values.

ahmedwaleedmalik, line 270: Check for nil before assigning these values

pkg/cloudprovider/provider/gce/config.go
```diff
@@ -101,6 +101,8 @@ type config struct {
 	network               string
 	subnetwork            string
 	preemptible           bool
+	automaticRestart      bool
+	provisioningModel     string
```
ahmedwaleedmalik, line 105: Preferably you should use pointers here

pkg/cloudprovider/provider/gce/provider.go
```diff
@@ -265,7 +265,9 @@ func (p *Provider) Create(machine *clusterv1alpha1.Machine, data *cloudprovidert
 		Disks:             disks,
 		Labels:            labels,
 		Scheduling: &compute.Scheduling{
-			Preemptible: cfg.preemptible,
+			Preemptible:       cfg.preemptible,
+			AutomaticRestart:  &cfg.automaticRestart,
```
ahmedwaleedmalik, line 269: This is an incomplete implementation as you are not consuming values from `ProviderSpec` for the new fields i.e. AutomaticRestart and ProvisioningModel.

Please go through the handling of `preemptible` in `newConfig` as an example:

https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/gce/config.go#L165

ahmedwaleedmalik (changes requested):

pkg/cloudprovider/provider/gce/config.go
tlamr: would it work to have some sensible defaults?

pkg/cloudprovider/provider/gce/provider.go
tlamr: what if we always provide an ok default?

pkg/cloudprovider/provider/gce/provider.go
tlamr: You are totally right, sorry for that!

pkg/cloudprovider/provider/gce/provider.go
ahmedwaleedmalik: It's already going to be defaulted in the upstream API(google) if you don't provide a value. In that case I think it makes complete sense to use pointers and just rely on their defaults.

pkg/cloudprovider/provider/gce/config.go
ahmedwaleedmalik: It's already going to be defaulted in the upstream API(google) if you don't provide a value. In that case I think it makes complete sense to use pointers and just rely on their defaults.

tlamr: ok, did my best, hopefully should be ok now. Would using "" as empty value for string work in go? Using pointer there would mean I'm unable to write pass it nicely in that struct the gcp uses.

Thanks @ahmedwaleedmalik for all the patience with my beginner go issues :)

ahmedwaleedmalik: > Thanks @ahmedwaleedmalik for all the patience with my beginner go issues :)

Happy to help ✌️

ahmedwaleedmalik:
@tlamr Thank you for the awesome work, it's a very valuable addition to the GCE config.

I have made some amendments with https://github.com/kubermatic/machine-controller/pull/1252/commits/1383ada7c828c12a4be73d3168edbc4c40d516e6. Can you please take a look? This is what I was trying to suggest earlier. Do let me know if that makes sense for you.

ahmedwaleedmalik (approved):
/approve 🚀

tlamr:
> 1383ada

ah, ok, thanks a lot! :)

PR #1254, opened 2022-04-19
rename NetworkFamily to AddressFamily

PratikDeoghare:

Rename NetworkFamily to AddressFamily. 

**Optional Release Note**:
```release-note
NONE
```


pkg/cloudprovider/provider/gce/service.go
```diff
@@ -102,7 +102,7 @@ func (svc *service) networkInterfaces(cfg *config) ([]*compute.NetworkInterface,
 				},
 			}
 		} else {
-			klog.Infof("network family doesn't specify dual stack: %s", cfg.providerConfig.Network.GetNetworkFamily())
+			klog.Infof("network family doesn't specify dual stack: %s", cfg.providerConfig.Network.GetAddressFamily())
```
rastislavs, line 105:
```suggestion
			klog.Infof("address family doesn't specify dual stack: %s", cfg.providerConfig.Network.GetAddressFamily())
```

PR #1263, opened 2022-04-26
define static ip config

PratikDeoghare:
Without  this change IP config  is static or not is decided just by checking if `.Network` is `nil`. 
All the fields in the `NetworkConfig` are not for static IP configuration anymore due to addition  of `IPFamily` field. 
This PR checks if the fields that specify static IP config are set or not and updates the code accordingly.  

PR also removes leftover refs to podCIDRs command  line flag. It was just wired through the code but never actually used.

Part of https://github.com/kubermatic/kubermatic/issues/9617

**Optional Release Note**:
```release-note
NONE
```

### DIFF
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -79,7 +79,7 @@ var (
 	nodeRegistryMirrors           string
 	nodePauseImage                string
 	nodeContainerRuntime          string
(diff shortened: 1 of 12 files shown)


ahmedwaleedmalik (approved):
/approve

PR #1272, opened 2022-05-04
Add support for Kubernetes 1.24 and containerd 1.5

embik:
Kubernetes 1.24 removed a couple of flags on `kubelet` and also requires containerd 1.5 minimum, so this PR accounts for both when creating 1.24 machines.

Amazon Linux 2 does not seem to have containerd 1.5 (yet?), so the PR sets that to still use 1.4. Might be worth it to follow https://github.com/awslabs/amazon-eks-ami/issues/800, although that is an EKS image and not directly Amazon Linux 2.

Fixes #

**Optional Release Note**:

```release-note
Add support for Kubernetes 1.24 (uses containerd 1.5 as container runtime except on Amazon Linux 2, which remains on 1.4)
```


pkg/containerruntime/docker.go
```diff
@@ -26,8 +26,9 @@ import (
 )
 
 const (
-	DefaultDockerVersion = "19.03"
-	LegacyDockerVersion  = "18.09"
+	DefaultDockerContainerdVersion = "1.4"
+	DefaultDockerVersion           = "19.03"
```
kron4eg, line 30: too old, let's make it 20.10

embik: I agree we should update this, but since 1.24 is not going to use docker anymore I'm hesitant to include this change in this PR. Would you be okay with us updating it in a separate PR?

ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik (commented):
/lgtm

PR #1276, opened 2022-05-05
set kubelet --node-ip based on IPFamily

PratikDeoghare:
Pass `--node-ip` values based  on IPFamily. 

Part of https://github.com/kubermatic/kubermatic/issues/9617

**Optional Release Note**:

```release-note
NONE
```


PratikDeoghare:
Ref https://github.com/kubernetes/kubernetes/blob/e8ac6d8a8e27305e52f2a611d0f3df28fde20761/cmd/kubelet/app/server.go#L1108

moadqassem (commented):
Just a nit, also you need to reflect this change on osm

PR #1277, opened 2022-05-06
Update containerd to 1.5 for all Kubernetes versions

embik:
As discussed, this PR changes logic introduced with #1272 to only use containerd 1.5 in Kubernetes 1.24+. containerd 1.5 is now used for all Kubernetes versions still supported by `machine-controller`. Amazon Linux 2 is the exception as it does not have containerd 1.5 available.

Note: containerd installed with docker remains on 1.4 right now due to docker 19.03 only mentioning containerd 1.3 in its release notes. Maybe 1.5 works fine too, but let's investigate that in a separate PR which also bumps docker to 20.10.

Fixes #

**Optional Release Note**:

```release-note
Update containerd to 1.5 when using it as container runtime (Amazon Linux 2 remains on 1.4 until newer versions are available)
```

### DIFF
--- a/pkg/containerruntime/containerd.go
+++ b/pkg/containerruntime/containerd.go
@@ -63,13 +63,10 @@ func (eng *Containerd) ScriptFor(os types.OperatingSystem) (string, error) {
 		args.ContainerdVersion = eng.version
 	}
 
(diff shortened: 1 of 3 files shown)


ahmedwaleedmalik (approved):
/approve ⚡

PR #1282, opened 2022-05-12
Remove use-osm flag from run-machine-controller script

ahmedwaleedmalik:

OSM is not GA yet, it's confusing for people who are trying to run machine-controller locally that it's enabled by default. Since you need to run OSM in parallel with machine-controller for everything to work in this case. We should disable it for now.

Fixes #

**Optional Release Note**:

```release-note
NONE
```


embik (approved):
🦭

PR #1283, opened 2022-05-13
remove check for dualstack unsupported zones in GCP

PratikDeoghare:
We are removing this check because GCP now supports dual stack in all  the regions. 

This is not directly documented. But the docs that had listed restrictions are gone and 
@rastislavs was able to create dual stack subnets anywhere. 

Partial fulfillment of https://github.com/kubermatic/kubermatic/pull/9751

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/gce/provider.go
+++ b/pkg/cloudprovider/provider/gce/provider.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
-	"strings"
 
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (approved):
/approve ⚡

ahmedwaleedmalik:
@xmudrii this test won't work till we have https://github.com/kubermatic/machine-controller/pull/1284 in

PR #1286, opened 2022-05-16
Fix cleanup for machines on azure

### LINKED ISSUE #1285: Cluster deletion stucks when there is a no node due to quota limit
### What happened?
- Created a user cluster on Azure with public IP Enabled
- Azure public IP quota got reached
- Cluster has a machine but no nodes due to no network interface created and which was due to quota limit. 
```
 Warning  ReconcilingError  2m14s (x20 over 14m)  machine_controller  failed to create machine at cloudprovider, due to failed to create public IP: failed to create public IP address: network.PublicIPAddressesClient#CreateOrUpdate: │
│  Failure sending request: StatusCode=400 -- Original Error: Code="PublicIPCountLimitReached" Message="Cannot create more than 25 public IP addresses for this subscription in this region." Details=[]
```
- Try deleting a cluster
- The cluster was not deleted even after a long wait
- Error in controller Manager
```
controller-manager {"level":"error","time":"2022-03-16T13:59:47.590Z","logger":"kubermatic_kubernetes_controller","caller":"kubernetes/cluster_controller.go:259","msg":"Reconciling failed","worker-name":"","request":"/8786dh5t6v","cluster":"8786dh5t6v","error":"Cluster.kubermatic.k8s.io \"8786dh5t6v\" is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{\"kubermatic.io/cleanup-usersshkeys-cluster-ids\"}","errorCauses":[{"error":"Cluster.kubermatic.k8s.io \"8786dh5t6v\" is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{\"kubermatic.io/cleanup-usersshkeys-cluster-ids\"}"}]}
```
### What should happen?
Cluster deletion should work irrespective of the status of a nodes.
### How to reproduce?
Follow steps mentioned in _What happened_ section with public IP quota limit.
### Environment
- KKP Version: 2.18.7
- Shared or separate master/seed clusters?: separate
- Master/seed cluster provider: ANY (in our case, it was Azure as master and vsphere as seed)

ahmedwaleedmalik:

Fix cleanup for machines on azure.

Fixes #1285

I don't think we need a release note for updating `github.com/Azure/azure-sdk-for-go` to v64.1.0 or do we?

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.18
 require (
 	cloud.google.com/go/logging v1.1.2
 	cloud.google.com/go/monitoring v1.4.0
-	github.com/Azure/azure-sdk-for-go v62.0.0+incompatible
+	github.com/Azure/azure-sdk-for-go v64.1.0+incompatible
 	github.com/Azure/go-autorest/autorest/azure/auth v0.5.5
 	github.com/Azure/go-autorest/autorest/to v0.4.0
 	github.com/BurntSushi/toml v0.3.1
--- a/go.sum
+++ b/go.sum
@@ -56,8 +56,8 @@ cloud.google.com/go/storage v1.6.0/go.mod h1:N7U0C8pVQ/+NIKOBQyamJIeKQKkZ+mxpohl
 cloud.google.com/go/storage v1.8.0/go.mod h1:Wv1Oy7z6Yz3DshWRJFhqM/UCfaWIRTdp0RXyy7KQOVs=
 cloud.google.com/go/storage v1.10.0/go.mod h1:FLPqc6j+Ki4BU591ie1oL6qBQGu2Bl/tZ9ullr3+Kg0=
 dmitri.shuralyov.com/gpu/mtl v0.0.0-20190408044501-666a987793e9/go.mod h1:H6x//7gZCb22OMCxBHrMx7a5I7Hp++hsVxbQ4BYO7hU=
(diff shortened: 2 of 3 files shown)


embik:
This makes sense to me personally, but I wonder why it was done differently in the first place. How does it fix the bug in question by the way?

ahmedwaleedmalik:
> How does it fix the bug in question by the way?

Basically, you get left with a finalizer for `public IP` on the machine but it's never removed. It tries to clean up and returns when the instance is not found https://github.com/kubermatic/machine-controller/blob/azure-cleanup/pkg/cloudprovider/provider/azure/provider.go#L739

Cluster cleanup fails as a result.

I have ideas for further improvements in this particular case. But for now this PR should be enough for fixing the customer issue.

ahmedwaleedmalik:
> https://github.com/kubermatic/machine-controller/pull/288#issuecomment-408425491 this is where the decision was made to add the finalizers beforehand. Essentially it's done so machine-controller does not forget to clean up a resource in case adding the finalizer fails. I can see where that is coming from.

We `delete` the resource BEFORE removing or checking for the finalizer. So I don't think that statement is valid anymore.

embik (approved):
/approve

PR #1287, opened 2022-05-16
Only attempt resource deletion in azure provider if finalizer is set

embik:
We set finalizers for different resources created by the azure provider, but the cleanup logic effectively ignored their presence (except for removing them) and always tried to delete all resources. When cleanup is only partially successful, subsequent attempts should be faster because there is no reason to try deleting the VM again if we were able to delete it before.

Also replaces a bunch of `ListAll` calls for VMs because we know the resource group to query, so let's not list _all_ VMs.

Fixes #

**Optional Release Note**:

```release-note
Azure resources are ignored during machine cleanup if resource-specific finalizers are not set
```


ahmedwaleedmalik (commented):
/approve ❤️

PR #1298, opened 2022-05-19
Support node registry credentials with docker

### LINKED ISSUE #1279: Add support to set node wide registry credentials for container runtime docker
Currently the registry credentials passed via flag mentioned in [docs/registry-authentication.md](https://github.com/kubermatic/machine-controller/blob/master/docs/registry-authentication.md) is only used with containerd as container runtime. For https://github.com/kubermatic/kubermatic/pull/9260 it would be nice if this flag was container runtime agnostic and also work if docker is used.

dermorz:
This PR introduces the possibility to set node-level registry authentication when using docker as CRI. Before this has only been possible when using containerd. Although docker as CRI is being phased out, not everyone who wants this feature has made the switch to containerd.

It's now also possible to reference secrets with the type [`kubernetes.io/dockerconfigjson`](https://kubernetes.io/docs/concepts/configuration/secret/#docker-config-secrets).

This is needed for the implementation of [Proposal: Container Registry Credentials per User Cluster](https://github.com/kubermatic/kubermatic/blob/master/docs/proposals/container-registry-credentials-per-usercluster.md).

Fixes #1279 

Regarding the addition of `User=root` to the kubelet systemd unit file check: https://github.com/kubernetes/kubernetes/issues/45487

**Optional Release Note**:

```release-note
Add support for setting node level registry credentials with CRI docker
```


pkg/userdata/amzn2/provider.go
```diff
@@ -302,6 +311,14 @@ write_files:
   content: |
 {{ .ContainerRuntimeConfig | indent 4 }}
 
+{{- if and (eq .ContainerRuntimeName "docker") (ne .ContainerRuntimeAuthConfig "{}") }}
```
ahmedwaleedmalik, line 314: `(ne .ContainerRuntimeAuthConfig "{}")` should return an empty string instead of `{}` if no credentials were provided

ahmedwaleedmalik (changes requested):

ahmedwaleedmalik (approved):
Please implement the same for OSM 😁 
/approve ⚡

ahmedwaleedmalik:
/hold

This is not flaking out.

ahmedwaleedmalik (approved):
/approve ✌️ 
/unhold

PR #1300, opened 2022-05-19
Add support for VMware Cloud Director as cloud provider

### LINKED ISSUE #1262: Support VMware vCloud Director as a Cloud Provider
- [x] Support in machine-controller
- [x] E2E tests
- [x] Documentation

ahmedwaleedmalik:

Add support for VMware Cloud Director as cloud provider. 🥳 

Fixes #1262

I have intentionally only covered `ubuntu` as an operating system for now. The plan is to integrate other operating systems in separate PRs once we have the templates ready.

**Optional Release Note**:

```release-note
Support for VMware Cloud Director as cloud provider
```


examples/vmware-cloud-director-machinedeployment.yaml
```diff
@@ -0,0 +1,87 @@
...
+          operatingSystemSpec:
+            distUpgradeOnBoot: false
+            disableAutoUpdate: true
+            # 'rhelSubscriptionManagerUser' is only used for rhel os and can be set via env var `RHEL_SUBSCRIPTION_MANAGER_USER`
+            rhelSubscriptionManagerUser: "<< RHEL_SUBSCRIPTION_MANAGER_USER >>"
+            # 'rhelSubscriptionManagerPassword' is only used for rhel os and can be set via env var `RHEL_SUBSCRIPTION_MANAGER_PASSWORD`
+            rhelSubscriptionManagerPassword: "<< RHEL_SUBSCRIPTION_MANAGER_PASSWORD >>"
+            # 'rhsmOfflineToken' if it was provided red hat systems subscriptions will be removed upon machines deletions, and if wasn't
+            # provided the rhsm will be disabled and any created subscription won't be removed automatically
+            rhsmOfflineToken: "<< REDHAT_SUBSCRIPTIONS_OFFLINE_TOKEN >>"
```
kron4eg, line 85: ubuntu's Config structure doesn't have those fields

https://github.com/kubermatic/machine-controller/blob/ebbd117869da8a35867e74193510ef6241adb4e1/pkg/userdata/ubuntu/ubuntu.go#L25-L28

ahmedwaleedmalik: True but does it matter? I think we add all of this by default in all of our examples and testdata so that switching between different OS is easier. It will get ignored since this is not strictly unmarshalled.

kron4eg: Yes, let's not confuse people with non-existing fields.

ahmedwaleedmalik: Updated PTAL 👍

pkg/cloudprovider/provider/vmware-cloud-director/provider.go
```diff
@@ -0,0 +1,531 @@
...
+		return nil, nil, nil, err
+	}
+
+	c.IPAllocationMode = rawConfig.IPAllocationMode
+	c.CPUs = rawConfig.CPUs
+	c.CPUCores = rawConfig.CPUCores
+	c.MemoryMB = rawConfig.MemoryMB
+	c.DiskSizeGB = rawConfig.DiskSizeGB
+	c.DiskBusType = rawConfig.DiskBusType
+	c.DiskIOPS = rawConfig.DiskIOPS
```
xmudrii, line 377: Are there any minimum and maximum values for those fields? For example, what if someone provides 0 IOPS? Or even something like negative IOPS or size?

pkg/cloudprovider/provider/vmware-cloud-director/provider.go
```diff
@@ -0,0 +1,531 @@
...
+				return fmt.Errorf("sizing policy '%s' doesn't exist", *c.SizingPolicy)
+			}
+		}
+
+		if c.PlacementPolicy != nil && *c.PlacementPolicy != "" {
+			placementPolicy := getComputePolicy(*c.PlacementPolicy, allPolicies)
+			if placementPolicy == nil {
+				return fmt.Errorf("placement policy '%s' doesn't exist", *c.SizingPolicy)
+			}
+		}
```
xmudrii, line 496: As far as I see, we only validate if it is a valid/existing compute policy. What if someone provides a placement policy as a sizing policy (and vice-versa)? Does it matter at all?

examples/vmware-cloud-director-machinedeployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+            # Can also be set via the env var 'VCD_ALLOW_UNVERIFIED_SSL' on the machine-controller
+            allowInsecure: false
+            # vApp to associate the VM with. This should be created before the machine is created
+            vapp: "<< VCD_VAPP >>"
+            # Name of catalog where the VM template is located
+            catalog: "<< VCD_CATALOG >>"
+            # Name of OS template to be used for the VM
+            template: "<< VCD_TEMPLATE >>"
+            # Direct or routed network that can be used for the VM
+            network: "<< VCD_NETWORK >>"
```
xmudrii, line 64: I don't see that we define those environment variables in `provider.go`. Is that as expected?

pkg/cloudprovider/provider/vmware-cloud-director/client.go
```diff
@@ -0,0 +1,173 @@
...
+	return vcd, err
+}
+
+func (c *Client) GetVMByName(vappName, vmName string) (*govcd.VM, error) {
+	_, _, vapp, err := c.GetOrganizationVDCAndVapp(vappName)
+	if err != nil {
+		return nil, err
+	}
+
+	// We don't need ID here since we explicitly set the name field when creating the resource.
```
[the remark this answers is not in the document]
ahmedwaleedmalik: Yes, but not sure that there is much that we can do about it. In this context, since the VM is created using `machine-controller` we assume that we can just use the name.

pkg/cloudprovider/provider/vmware-cloud-director/helper.go
```diff
@@ -0,0 +1,286 @@
...
+
+	vAppTemplate, err := catalogItem.GetVAppTemplate()
+	if err != nil {
+		return fmt.Errorf("failed to get vApp template '%s' in catalog '%s': %w", c.Template, c.Catalog, err)
+	}
+
+	templateHref := vAppTemplate.VAppTemplate.HREF
+	if vAppTemplate.VAppTemplate.Children != nil && len(vAppTemplate.VAppTemplate.Children.VM) != 0 {
+		templateHref = vAppTemplate.VAppTemplate.Children.VM[0].HREF
+	}
```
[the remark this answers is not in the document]
ahmedwaleedmalik: Yes, it's possible if we generate a template from a vApp with more than 1 VMs. In that case we just assume that the first VM is the one that we want to clone.

pkg/cloudprovider/provider/vmware-cloud-director/helper.go
```diff
@@ -0,0 +1,286 @@
...
+
+func recomposeComputeAndDisk(config *Config, vm *govcd.VM) (*govcd.VM, error) {
+	needsComputeRecomposition := false
+	needsDiskRecomposition := false
+	// Perform compute recomposition if SizingPolicy was not specified.
+	vmSpecSection := vm.VM.VmSpecSection
+	if config.SizingPolicy == nil || *config.SizingPolicy == "" {
+		vmSpecSection.MemoryResourceMb.Configured = config.MemoryMB
+		vmSpecSection.NumCpus = pointer.Int(int(config.CPUs))
+		vmSpecSection.NumCoresPerSocket = pointer.Int(int(config.CPUCores))
```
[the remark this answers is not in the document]
ahmedwaleedmalik: `SizingPolicy` is optional and these values will only be used if `SizingPolicy` is not specified. `nil` value in this case would mean that we just use the value from the template and we won't override it.

pkg/cloudprovider/provider/vmware-cloud-director/provider.go
ahmedwaleedmalik: 0 IOPS is allowed and the default value. Yeah maybe we can check for negative values :)

pkg/cloudprovider/provider/vmware-cloud-director/helper.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	return vm, nil
+}
+
+func setUserData(userdata string, vm *govcd.VM, providerConfig *providerconfigtypes.Config) error {
+	userdataBase64 := base64.StdEncoding.EncodeToString([]byte(userdata))
+	props := map[string]string{
+		"user-data":       userdataBase64,
+		"disk.enableUUID": "1",
+		"instance-id":     vm.VM.Name,
```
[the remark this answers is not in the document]
ahmedwaleedmalik: it's not really used anywhere

pkg/cloudprovider/provider/vmware-cloud-director/provider.go
ahmedwaleedmalik: Unfortunately, there is no way to detect the type at the moment. That's why we just check for their existence. 

Also, it's eventually just a `ComputePolicy` so it doesn't really matter.


https://github.com/kubermatic/machine-controller/blob/vmware-cloud-director/pkg/cloudprovider/provider/vmware-cloud-director/helper.go#L89

examples/vmware-cloud-director-machinedeployment.yaml
xmudrii: My bad -- for some reason I thought those are environment variable names.

pkg/cloudprovider/provider/vmware-cloud-director/client.go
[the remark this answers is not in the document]
xmudrii: What if we have more than one Machine with the same name? Which one will be picked by the client? Why don't we use the ID instead?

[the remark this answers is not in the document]
ahmedwaleedmalik: We can't set the `ID` when creating a VM. It's auto-assigned.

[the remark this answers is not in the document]
xmudrii: We decided to leave this for a follow-up.

xmudrii (approved):
/lgtm
/approve
/hold
if anyone else wants to take a look

PR #1304, opened 2022-05-20
Use kind for E2E tests

ahmedwaleedmalik:
Signed-off-by: Waleed Malik <ahmedwaleedmalik@gmail.com>

**What this PR does / why we need it**:
Alt Text

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Optional Release Note**:

```release-note
NONE
```

PR #1307, opened 2022-05-24
Switch to Ubuntu 22.04 LTS for e2e tests

ahmedwaleedmalik:

We were previously using `kubic` packages from SUSE for the installation of buildah. Although they have been discontinued.

The alternative that I chose is to upgrade to `Ubuntu 22.04` in this case. Although eventually, we'll remove all of this in favor of KIND cluster and docker for building image with https://github.com/kubermatic/machine-controller/pull/1304

```
"NOTE: Kubic packages have been discontinued for Ubuntu 22.04 LTS. Current users of the Kubic repos for Ubuntu are highly recommended to uninstall the packages from the Kubic repos before upgrading to Ubuntu 22.04 LTS."
```

Fixes #

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 ARG GO_VERSION=1.18.2
-FROM golang:${GO_VERSION} AS builder
+FROM docker.io/golang:${GO_VERSION} AS builder
 WORKDIR /go/src/github.com/kubermatic/machine-controller
 COPY . .
 RUN make all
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 ARG GO_VERSION=1.18.2
-FROM golang:${GO_VERSION} AS builder
```
LittleFox94, line 16: I'd be interested why this is needed, can you tell me more @ahmedwaleedmalik?

PR #1312, opened 2022-05-25
Block setting spec.ConfigSource for Kubernetes 1.24+

embik:

We already removed the feature flag in kubelet, but in my opinion we should also block this from being set in a `MachineSpec`, as it mirrors to the `Node` object and I would expect that the `NodeSpec` fields are going to go away eventually as well.

Fixes #

**Optional Release Note**:

```release-note
Machine specs for Kubernetes 1.24 or higher are not allowed to set spec.ConfigSource (as the dynamic kubelet configuration feature has been removed)
```

### DIFF
--- a/pkg/admission/machines.go
+++ b/pkg/admission/machines.go
@@ -139,6 +139,19 @@ func (ad *admissionData) defaultAndValidateMachineSpec(ctx context.Context, spec
 		return fmt.Errorf("kubernetes version constraint didn't allow %q kubelet version", kubeletVer)
 	}
 
+	// Do not allow 1.24+ to use config source (dynamic kubelet configuration)
+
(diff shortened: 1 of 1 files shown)


pkg/admission/machines.go
```diff
@@ -139,6 +139,19 @@ func (ad *admissionData) defaultAndValidateMachineSpec(ctx context.Context, spec
 		return fmt.Errorf("kubernetes version constraint didn't allow %q kubelet version", kubeletVer)
 	}
 
+	// Do not allow 1.24+ to use config source (dynamic kubelet configuration)
+
```
ahmedwaleedmalik, line 143: Please remove the extra newline

ahmedwaleedmalik (approved):
/approve

PR #1322, opened 2022-06-07
Reduce instance profile validation to existence check

### LINKED ISSUE #787: Support running MC in private subnet in AWS without disabling webhook validation
When machine-controller is running in a private subnet in AWS without NAT gateway, it is possible to rely on [amazon api gateway](https://aws.amazon.com/blogs/compute/introducing-amazon-api-gateway-private-endpoints/) in order to access EC2 API. The same is not possible for IAM API, and this makes the `machine` and `machinedeployment` validation fail because the instance profile existence check relies on it (see https://github.com/kubermatic/machine-controller/blob/56a7f0e/pkg/cloudprovider/provider/aws/provider.go#L446).

The goal of this task is to define and implement a strategy to support this use-case without disabling the validation.

Some possible options:

- Introduce a flag to disable instance profile check
- Make it best effort e.g. in case of connection timeout when contacting IAM API don't consider as a failure
- Reduce scope of the check and only verify that is present

dermorz:
In offline setups it's not possible to reach the IAM API. To make the machine controller function in a private VPC with no internet access, the validation of the instance profile needs to be reduced to a simple check, if it was set.

Fixes #787

Reference: https://stackoverflow.com/questions/72307375

**Optional Release Note**:

```release-note
(AWS) Instance profile validation is no longer calling the IAM API
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -31,7 +31,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
-	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/aws/aws-sdk-go/service/sts"
 	gocache "github.com/patrickmn/go-cache"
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve 👍 
/hold for @kron4eg to review this as well.

PR #1323, opened 2022-06-07
add dual stack support for DigitalOcean

PratikDeoghare:

Adds dual stack support for DigitalOcean.

Part of https://github.com/kubermatic/kubermatic/issues/10005

**Optional Release Note**:

```release-note
Added dual stack support for DigitalOcean.
```

### DIFF
--- a/pkg/cloudprovider/provider/digitalocean/provider.go
+++ b/pkg/cloudprovider/provider/digitalocean/provider.go
@@ -310,7 +310,7 @@ func (p *provider) Create(ctx context.Context, machine *clusterv1alpha1.Machine,
 		Name:              machine.Spec.Name,
 		Region:            c.Region,
 		Size:              c.Size,
-		IPv6:              c.IPv6 || pc.Network.GetIPFamily() == util.DualStack, // since c.IPv6 is deprecated
+		IPv6:              c.IPv6 || pc.Network.GetIPFamily() == util.DualStack,
 		PrivateNetworking: c.PrivateNetworking,
 		Backups:           c.Backups,
 		Monitoring:        c.Monitoring,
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/digitalocean/provider.go
```diff
@@ -188,6 +189,17 @@ func (p *provider) Validate(ctx context.Context, spec clusterv1alpha1.MachineSpe
...
+	switch f := pc.Network.GetIPFamily(); f {
+	case util.Unspecified, util.IPv4:
+	// noop
+	case util.IPv6:
+		return fmt.Errorf(util.ErrIPv6OnlyUnsupported)
+	case util.DualStack:
+		// noop
+	default:
+		return fmt.Errorf(util.ErrUnknownNetworkFamily, f)
+	}
```
ahmedwaleedmalik, line 201: Do you really need `switch` here? All I could garner is that you need:

```
if pc.Network.GetIPFamily() == util.IPv6 {
     return fmt.Errorf(util.ErrIPv6OnlyUnsupported)
}
```

We'll never reach the `default` condition imo.

PratikDeoghare: If you mistype `IPv6` as `Ipv6` you can reach the default condition.

ahmedwaleedmalik:
```
// IPFamily IPv4 | IPv6 | IPv4+IPv6.
type IPFamily string

const (
	Unspecified IPFamily = "" // interpreted as IPv4
	IPv4        IPFamily = "IPv4"
	IPv6        IPFamily = "IPv6"
	DualStack   IPFamily = "IPv4+IPv6"
)
```

I assumed that marshaling would have errored out since it's not an acceptable value for the type. But I guess it doesn't.

Can we add a validation directly in the webhook for this? Just to check if IPFamily has a valid value, irrespective of the cloud provider. That'd make our life so much easier.

ahmedwaleedmalik (approved):
/approve ⚡

PR #1336, opened 2022-06-24
set --node-ip depending on external CCM availaibilty in dualstack 

PratikDeoghare:

`--node-ip` and `--cloud-provider` kubelet flags conflict in the dualstack setup.

 In general, it is not expected to need to use `--node-ip` with external CCMs, as the the cloud provider is expected to know the correct IPs to return.

This PR

```
|--------------+------------------------------+----------------------------|
|              | KKP support for external CCM | --node-ip set in dualstack |
|--------------+------------------------------+----------------------------|
| DigitalOcean | No                           | Yes                        |
| OpenStack    | Yes                          | No                         |
|--------------+------------------------------+----------------------------|
```

Fixes 
https://github.com/kubermatic/kubermatic/issues/10152 
https://github.com/kubermatic/kubermatic/issues/9959

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/pkg/userdata/helper/helper.go
+++ b/pkg/userdata/helper/helper.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/kubermatic/machine-controller/pkg/cloudprovider/util"
+
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
@@ -146,14 +148,31 @@ NO_PROXY=%s
 no_proxy=%s`, proxy, proxy, proxy, proxy, noProxy, noProxy)
 }
 
-func SetupNodeIPEnvScript() string {
+func SetupNodeIPEnvScript(ipFamily util.IPFamily) string {
+	const defaultIfcIPv4 = `DEFAULT_IFC_IP=$(ip -o  route get 1 | grep -oP "src \K\S+")`
+
+	var defaultIfcIP string
+	switch ipFamily {
+	case util.Unspecified, util.IPv4:
+		defaultIfcIP = defaultIfcIPv4
+	case util.IPv6:
+		defaultIfcIP = `DEFAULT_IFC_IP=$(ip -o -6 route get  1:: | grep -oP "src \K\S+")`
+	case util.DualStack:
+		defaultIfcIP = `
+		DEFAULT_IFC_IP=$(ip -o route get  1 | grep -oP "src \K\S+")
+		DEFAULT_IFC_IP6=$(ip -o -6 route get  1:: | grep -oP "src \K\S+")
+		DEFAULT_IFC_IP=$DEFAULT_IFC_IP,$DEFAULT_IFC_IP6
+		`
+	default:
+		defaultIfcIP = defaultIfcIPv4
+	}
 	return `#!/usr/bin/env bash
 echodate() {
   echo "[$(date -Is)]" "$@"
 }
 
 # get the default interface IP address
-DEFAULT_IFC_IP=$(ip -o  route get 1 | grep -oP "src \K\S+")
+` + defaultIfcIP + `
 
 # get the full hostname
 FULL_HOSTNAME=$(hostname -f)
--- a/pkg/userdata/helper/helper.go
+++ b/pkg/userdata/helper/helper.go
@@ -158,11 +158,9 @@ func SetupNodeIPEnvScript(ipFamily util.IPFamily) string {
 	case util.IPv6:
 		defaultIfcIP = `DEFAULT_IFC_IP=$(ip -o -6 route get  1:: | grep -oP "src \K\S+")`
 	case util.DualStack:
-		defaultIfcIP = `
-		DEFAULT_IFC_IP=$(ip -o route get  1 | grep -oP "src \K\S+")
-		DEFAULT_IFC_IP6=$(ip -o -6 route get  1:: | grep -oP "src \K\S+")
-		DEFAULT_IFC_IP=$DEFAULT_IFC_IP,$DEFAULT_IFC_IP6
-		`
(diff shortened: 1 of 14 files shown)


pkg/userdata/helper/helper.go
```diff
@@ -157,14 +159,29 @@ NO_PROXY=%s
 no_proxy=%s`, proxy, proxy, proxy, proxy, noProxy, noProxy)
 }
 
-func SetupNodeIPEnvScript() string {
+func SetupNodeIPEnvScript(ipFamily util.IPFamily) string {
+	const defaultIfcIPv4 = `DEFAULT_IFC_IP=$(ip -o  route get 1 | grep -oP "src \K\S+")`
+
+	var defaultIfcIP string
+	switch ipFamily {
+	case util.Unspecified, util.IPv4:
```
moadqassem, line 167: If you have a default case already, which should cover IPv4 why do you need this case here?

pkg/userdata/helper/kubelet.go
```diff
@@ -39,15 +41,20 @@ const (
 	defaultKubeletContainerLogMaxSize = "100Mi"
 )
 
-const (
-	kubeletFlagsTpl = `--bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf \
+func kubeletFlagsTpl(withNodeIP, withCloudProvider bool) string {
+	x := `--bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf \
```
moadqassem, line 45: Nit: can you please a better name for this var instead if `x`

pkg/userdata/helper/kubelet.go
```diff
@@ -39,15 +41,20 @@ const (
...
 --kubeconfig=/var/lib/kubelet/kubeconfig \
 --config=/etc/kubernetes/kubelet.conf \
---cert-dir=/etc/kubernetes/pki \
+--cert-dir=/etc/kubernetes/pki \`
+
+	if withCloudProvider {
```
moadqassem, line 50: What does that mean? Can you please elaborate what this `withCloudProvider` parameter is used for?

pkg/userdata/helper/kubelet.go
```diff
@@ -279,8 +297,33 @@ func kubeletConfiguration(clusterDomain string, clusterDNS []net.IP, featureGate
 }
 
 // KubeletFlags returns the kubelet flags.
-func KubeletFlags(version, cloudProvider, hostname string, dnsIPs []net.IP, external bool, pauseImage string, initialTaints []corev1.Taint, extraKubeletFlags []string) (string, error) {
-	tmpl, err := template.New("kubelet-flags").Funcs(TxtFuncMap()).Parse(kubeletFlagsTpl)
+func KubeletFlags(version, cloudProvider, hostname string, dnsIPs []net.IP, external bool, ipFamily util.IPFamily, pauseImage string, initialTaints []corev1.Taint, extraKubeletFlags []string) (string, error) {
+	// --node-ip and --cloud-provider kubelet flags conflict in the dualstack setup.
```
moadqassem, line 301: Probably this documentation should be written on top of the func since you are describing the effect of parameter `ipFamily`

pkg/userdata/helper/kubelet.go
```diff
@@ -279,8 +297,33 @@ func kubeletConfiguration(clusterDomain string, clusterDNS []net.IP, featureGate
...
+	// In general, it is not expected to need to use --node-ip with external CCMs,
+	// as the cloud provider is expected to know the correct IPs to return.
+
+	// For details read kubernetes/sig-networking channel discussion
+	// https://kubernetes.slack.com/archives/C09QYUH5W/p1654003958331739
+
+	withCloudProvider := true
```
moadqassem, line 308: Ok so this parameter means, whether a CCM is supported or not, can you please have a better name for it? as `withCloudProvider` is super vague and doesn't really have an obvious meaning in this context. maybe something like `ccmSupported` or whatever naming that reflects the functionality of this guy :)

moadqassem: Oh and small nit, in go usually default values for bools is false, so that you can have early exit of the code

pkg/userdata/helper/helper.go
PratikDeoghare: `util.Unspecified` covers when nothing was set assuming everything will be IPv4. 

`default` covers typing mistakes `IPv5` for example.

pkg/userdata/helper/kubelet.go
PratikDeoghare: It is about kubelet's `--cloud-provider` flag. Changed name to make that clear.

PratikDeoghare: > Oh and small nit, in go usually default values for bools is false...

Here the value will always be true except in one case. So makes sense to start with true.

pkg/userdata/helper/kubelet.go
PratikDeoghare: Makes sense. Will add to the function doc.

pkg/userdata/helper/helper.go
PratikDeoghare: Ok. Will keep the default case only.

moadqassem (approved):
/approve
/lgtm

ahmedwaleedmalik:
@PratikDeoghare It doesn't seem like these retries would work. You can get this overriden.

```
W0713 15:41:49.240722   22185 verify.go:157] Creation of "ubuntu-1-22-7-vcd-1547242389494566912" failed, retrying: admission webhook "machinedeployments.machine-controller.kubermatic.io" denied the request: validation failed: failed to get organization 'ORG0087': [ENF] entity not found
```

P.s. It's always better to look at logs for the reason of failures instead of just `/retry` :P

PR #1339, opened 2022-06-27
Add metro support for equinixmetal

moadqassem:
Adding support for equinixmetal metro field in machine controller. This is needed to enable machine controller equinixmetal cloud provider to pick any server available in a whole metro(group of facilities) to avoid availability problems when deploying multiple replica. 

Fixes #

**Optional Release Note**:

```release-note
Equinixmetal metro support in machine controller 
```

### DIFF
--- a/pkg/cloudprovider/provider/equinixmetal/types/types.go
+++ b/pkg/cloudprovider/provider/equinixmetal/types/types.go
@@ -26,7 +26,8 @@ type RawConfig struct {
 	ProjectID    providerconfigtypes.ConfigVarString   `json:"projectID,omitempty"`
 	BillingCycle providerconfigtypes.ConfigVarString   `json:"billingCycle"`
 	InstanceType providerconfigtypes.ConfigVarString   `json:"instanceType"`
-	Facilities   []providerconfigtypes.ConfigVarString `json:"facilities"`
+	Metro        providerconfigtypes.ConfigVarString   `json:"metro,omitempty"`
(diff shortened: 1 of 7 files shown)


pkg/cloudprovider/provider/equinixmetal/types/types.go
```diff
@@ -26,6 +26,7 @@ type RawConfig struct {
 	ProjectID    providerconfigtypes.ConfigVarString   `json:"projectID,omitempty"`
 	BillingCycle providerconfigtypes.ConfigVarString   `json:"billingCycle"`
 	InstanceType providerconfigtypes.ConfigVarString   `json:"instanceType"`
+	Metro        providerconfigtypes.ConfigVarString   `json:"metro"`
```
ahmedwaleedmalik, line 29: Shouldn't this be an optional field? Basically, both Facilities and metro should be made optional and then we can check in `Validate` that at least one of them is specified. WDYT?

moadqassem: I do this already https://github.com/kubermatic/machine-controller/pull/1339/files#diff-86dce1d2d7e8fb9dfddab19175ce5b62bb771587033f0bb6a3b83961b9318e75R195

moadqassem: Oh actually that's true, will adjust that now

PR #1342, opened 2022-06-30
Upgrade to alpine 3.16 for machine-controller's image

ahmedwaleedmalik:

Upgrade to alpine 3.16 for machine-controller's image.

Fixes #

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ WORKDIR /go/src/github.com/kubermatic/machine-controller
 COPY . .
 RUN make all
 
-FROM alpine:3.12
+FROM alpine:3.16
 
 RUN apk add --no-cache ca-certificates cdrkit
 


Dockerfile
```diff
@@ -18,7 +18,7 @@ WORKDIR /go/src/github.com/kubermatic/machine-controller
 COPY . .
 RUN make all
 
-FROM alpine:3.12
+FROM alpine:3.16
```
kron4eg, line 21: just saying: I'd recommend distroless images

ahmedwaleedmalik: That'd be my personal preference as well. @xrstf WDYT? I agree with @kron4eg.

xrstf: I personally don't think the security gains / size savings of distroless are worth the hassle of losing the ability to debug a container quickly. But I also cannot remember the last time I exec'ed into a machine-controller pod, so... whatever the maintainers of the machine-controller think is right.

PR #1348, opened 2022-07-05
vcd: ipAllocationMode should be an optional field

ahmedwaleedmalik:

For VMware Cloud Director, `ipAllocationMode` should be an optional field since we are already [defaulting it to DHCP](https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/vmwareclouddirector/provider.go#L139). 

Fixes #

**Optional Release Note**:

```release-note
VMware Cloud Director: `ipAllocationMode` is an marked as optional field
```

### DIFF
--- a/pkg/cloudprovider/provider/vmwareclouddirector/types/types.go
+++ b/pkg/cloudprovider/provider/vmwareclouddirector/types/types.go
@@ -46,7 +46,7 @@ type RawConfig struct {
 
 	// Network configuration.
 	Network          providerconfigtypes.ConfigVarString `json:"network"`
-	IPAllocationMode IPAllocationMode                    `json:"ipAllocationMode"`
+	IPAllocationMode IPAllocationMode                    `json:"ipAllocationMode,omitempty"`
 
 	// Compute configuration.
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/vmwareclouddirector/types/types.go
```diff
@@ -46,7 +46,7 @@ type RawConfig struct {
 
 	// Network configuration.
 	Network          providerconfigtypes.ConfigVarString `json:"network"`
-	IPAllocationMode IPAllocationMode                    `json:"ipAllocationMode"`
+	IPAllocationMode IPAllocationMode                    `json:"ipAllocationMode,omitempty"`
```
kron4eg, line 49: Just saying, this change has no effect on anything...

ahmedwaleedmalik: 😱

ahmedwaleedmalik: Wouldn't it generate an error for a missing field because we strictly unmarshall this?

PR #1351, opened 2022-07-08
Fix logic for finding node by ProviderID

### LINKED ISSUE #1353: Getting Node by providerID doesn't work for all providers
We're getting Node objects in two ways:

* by providerID
* by IP addresses (if getting Node by providerID wasn't successful, e.g. there's no CCM)

https://github.com/kubermatic/machine-controller/blob/5c8d3b44f270b6d39c3cbda991b9c4f39e37b77a/pkg/controller/machine/machine_controller.go#L1059-L1113

However, getting Node by providerID is broken for some providers. We should check if getting Node by providerID is working for all supported providers.

* [ ] Anexia
* [ ] AWS
* [ ] Azure
* [ ] DigitalOcean
* [ ] Equinix Metal
* [ ] GCE
* [ ] Hetzner
* [ ] KubeVirt
* [ ] Nutanix
* [ ] OpenStack
* [ ] vSphere

(providers listed here have CCM support either in KKP or KubeOne)

xmudrii:

`machine_controller` is matching Machine object with respective Node object by:

* matching expected ProviderID with actual ProviderID of the Node object
* matching IP addresses of Machine object with IP addresses of the Node object

Matching Machines and Nodes by ProviderID currently works only for some providers such as AWS and Azure. It doesn't work for GCP, vSphere, and some other providers. This is because we expect all providers to format ProviderID like `<provider-name>:///<instance-id>` but that's not the case. Instead:

* some providers use double-slash instead of triple-slash, e.g. `<provider-name>://<instance-id>`
* some providers use something other than `<instance-id>`, e.g. `<provider-name>:///<availability-zone>/<instance-id>`

That said, the issue is that we're generating incorrect expected ProviderID for some providers. This PR fixes this problem by extending the `Instance` interface with the `ProviderID()` function which returns the expected ProviderID based on how CCM generates ProviderID for that provider. We use this function in place of the old expected ProviderID generation.

The reason for extending the `Instance` interface instead of going with another approach is:

* we anyways need to implement the `ProviderID()` function for each provider because formats are different
* this approach doesn't require modifying signature of the `getNode()` function which might be more complicated
* in the majority of cases, the information needed to generate the expected ProviderID is private and not available to `machine_controller` (e.g. availability zone, project ID, UUID...)

Fixes #1353

**Optional Release Note**:
```release-note
Fix finding nodes by providerID for clusters with CCM or in-tree cloud provider
```

### DIFF
--- a/pkg/cloudprovider/provider/anexia/instance.go
+++ b/pkg/cloudprovider/provider/anexia/instance.go
@@ -45,6 +45,10 @@ func (ai *anexiaInstance) ID() string {
 	return ai.info.Identifier
 }
 
+func (ai *anexiaInstance) ProviderID() string {
+	return ""
+}
+
 func (ai *anexiaInstance) Addresses() map[string]v1.NodeAddressType {
 	addresses := map[string]v1.NodeAddressType{}
 
--- a/pkg/cloudprovider/provider/anexia/instance.go
+++ b/pkg/cloudprovider/provider/anexia/instance.go
@@ -45,6 +45,7 @@ func (ai *anexiaInstance) ID() string {
 	return ai.info.Identifier
 }
 
+// TODO(xmudrii): Implement this.
 func (ai *anexiaInstance) ProviderID() string {
 	return ""
 }
--- a/pkg/cloudprovider/provider/equinixmetal/provider.go
+++ b/pkg/cloudprovider/provider/equinixmetal/provider.go
@@ -400,7 +400,7 @@ func (s *metalDevice) ID() string {
 }
 
 func (s *metalDevice) ProviderID() string {
-	return ""
+	return "equinixmetal://" + s.device.ID
 }
 
 func (s *metalDevice) Addresses() map[string]v1.NodeAddressType {
--- a/pkg/cloudprovider/provider/equinixmetal/provider.go
+++ b/pkg/cloudprovider/provider/equinixmetal/provider.go
@@ -399,6 +399,10 @@ func (s *metalDevice) ID() string {
 	return s.device.ID
 }
 
+func (s *metalDevice) ProviderID() string {
+	return ""
+}
+
 func (s *metalDevice) Addresses() map[string]v1.NodeAddressType {
 	// returns addresses in CIDR format
 	addresses := map[string]v1.NodeAddressType{}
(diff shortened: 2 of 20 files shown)


pkg/controller/machine/machine_controller.go
```diff
@@ -1065,16 +1065,21 @@ func (r *Reconciler) getNode(ctx context.Context, instance instance.Instance, pr
...
 				return node.DeepCopy(), true, nil
 			}
 		} else {
-			if node.Spec.ProviderID == providerID {
+			// We trim leading slashes in raw ID, since we always want three slashes in full ID
+			// Some provider IDs have three slashes (e.g. AWS), but some have two (e.g. GCP, vSphere)
+			// To mitigate this, we split on "://", and then compare provider part (the first part),
+			// and the ID part (the second part) specially
+			pid := strings.Split(node.Spec.ProviderID, "://")
+			if len(pid) == 2 && pid[0] == string(provider) && strings.TrimLeft(pid[1], "/") == strings.TrimLeft(instance.ID(), "/") {
```
pkprzekwas, line 1082: nit: consider moving to function and adding some unittest with some realistic cloud provider ID examples.

xmudrii: Added some unit tests, however, it's hard to test ProviderID generation for real providers because we don't have unit tests for provider implementations.

xmudrii:
/hold
I want to manually test this for some providers once again.

pkg/cloudprovider/provider/anexia/instance.go
```diff
@@ -45,6 +45,11 @@ func (ai *anexiaInstance) ID() string {
 	return ai.info.Identifier
 }
 
+// TODO(xmudrii): Implement this.
+func (ai *anexiaInstance) ProviderID() string {
+	return ""
+}
```
xmudrii, line 51: I wasn't able to get Anexia cluster working, so this is left as a TODO once we can provision Anexia clusters.

pkg/cloudprovider/provider/equinixmetal/provider.go
```diff
@@ -399,6 +399,11 @@ func (s *metalDevice) ID() string {
 	return s.device.ID
 }
 
+// TODO(xmudrii): Implement this.
+func (s *metalDevice) ProviderID() string {
+	return ""
+}
```
xmudrii, line 405: Equinix Metal integrations for KKP and KubeOne are currently broken, so I wasn't able to create an Equinix Metal cluster.

xmudrii:
/retest
This is a flake, the VM got stuck in the Error state.

moadqassem (approved):
/approve 
/lgtm

PR #1363, opened 2022-07-14
Equinix: use m3.small.x86 for CentOS E2E tests

xmudrii:

`c3.small.x86` is not available in AM metro, so let's use `m3.small.x86` for CentOS E2E tests. We already use that instance size for Ubuntu and Rocky Linux E2E tests.

**Optional Release Note**:
```release-note
NONE
```

/assign @moadqassem


moadqassem (approved):
/approve
/lgtm

PR #1367, opened 2022-07-21
Fix unmarshalling for `configvarstring` in cloudProviderSpec

ahmedwaleedmalik:

To keep strings safe to embed inside HTML `json.Marshal()` encodes them using [HTMLEscape](https://github.com/golang/go/blob/go1.18.4/src/encoding/json/encode.go#L161). It replaces invalid bytes with the Unicode replacement rune. For example "<", ">" will be converted to "\u003c","\u003e", respectively. `json.Unmarshal` takes care of replacing these escape characters with the actual value. 

In general, this works fine for us but in the case of `ConfigVarString` we override the unmarshalling and marshalling behavior. And Unmarshalling [doesn't replace the escape characters](https://github.com/kubermatic/machine-controller/blob/master/pkg/providerconfig/types/types.go#L164).

This PR reworks the `ConfigVarString` unmashalling to rectify this issue.

Special thanks to @xmudrii for coming up with the fix. ⚡ 

Fixes #

**Optional Release Note**:

```release-note
Fix unmarshalling for `configvarstring` in cloudProviderSpec.
```

### DIFF
--- a/pkg/providerconfig/types/types.go
+++ b/pkg/providerconfig/types/types.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strconv"
 
 	clusterv1alpha1 "github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
 	"github.com/kubermatic/machine-controller/pkg/cloudprovider/util"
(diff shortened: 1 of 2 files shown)


xmudrii (approved):
/lgtm
/approve
/hold
If anyone else wants to take a look.

pkg/providerconfig/types/types.go
```diff
@@ -226,7 +227,11 @@ func (configVarString *ConfigVarString) UnmarshalJSON(b []byte) error {
 	if !bytes.HasPrefix(b, []byte("{")) {
 		b = bytes.TrimPrefix(b, []byte(`"`))
 		b = bytes.TrimSuffix(b, []byte(`"`))
-		configVarString.Value = string(b)
+		s, err := strconv.Unquote("\"" + string(b) + "\"")
```
xmudrii, line 230: We should leave a comment that we intentionally remove prefix and suffix quotes and then add them for `Unquote`. That's because we don't know are there any quotes before calling `TrimPrefix` and `TrimSuffix`, and if the original byte slice doesn't have them, `Unquote` will fail. This is ensuring that this doesn't happen, i.e. we explicitly quote string, so `Unquote` will never fail.

ahmedwaleedmalik: Updated

xmudrii (approved):
Looks great! :zap:

PR #1385, opened 2022-07-29
Implement ProviderID() for Anexia

LittleFox94:

#1351 changed the mapping from Machine to Node, but left the `ProviderID()` method unimplemented for Anexia, this adds the implementation.

Our CCM currently does not set any prefix, which I guess could make problems when having a single cluster spanning multiple providers. At least they are fairly long and random, so chance for collision isn't too high .. still, if you need us to change that, I'll happily find a way for that.

Probably would make the most sense if @xmudrii takes a look at this, as I took that named TODO ^^'

**Optional Release Note**:

```release-note
Implement ProviderID function for Anexia
```

### DIFF
--- a/pkg/cloudprovider/provider/anexia/instance.go
+++ b/pkg/cloudprovider/provider/anexia/instance.go
@@ -45,9 +45,8 @@ func (ai *anexiaInstance) ID() string {
 	return ai.info.Identifier
 }
 
-// TODO(xmudrii): Implement this.
 func (ai *anexiaInstance) ProviderID() string {
-	return ""
+	return ai.ID()
 }
 
 func (ai *anexiaInstance) Addresses() map[string]v1.NodeAddressType {


pkg/cloudprovider/provider/anexia/instance.go
```diff
@@ -45,9 +45,8 @@ func (ai *anexiaInstance) ID() string {
 	return ai.info.Identifier
 }
 
-// TODO(xmudrii): Implement this.
 func (ai *anexiaInstance) ProviderID() string {
-	return ""
+	return ai.info.Identifier
```
LittleFox94, line 49: hm, `ID()` above checks if `ai.info` is nil .. not sure if this is always called late enough in lifecycle to not be a problem

xmudrii: I guess we could also have that check here to be on the safe side.

LittleFox94: changed `ProviderID()` to just call `ID()`

xmudrii (approved):
/lgtm
/approve

moadqassem (approved):
/approve
/lgtm

LittleFox94:
Thanks for fast approval!

You can always ping me for Anexia stuff btw :)

PR #1389, opened 2022-08-01
fix kubevirt secondary disk naming

### LINKED ISSUE #1378: KubeVirt: Conflict with Secondary Disk naming
Secondary Disk for kubevirt VM has a naming pattern like `secondarydisk0`, `secondarydisk1`.
After scaling the Machine deployment, all the VMs will try to create secondary disks with the same naming pattern resulting in a conflict and some VMs doesn't work.

sankalp-r:
This PR adds 'machine-name' prefix to secondary disk's name for kubevirt provider to avoid naming conflict.

Fixes #1378 

**Optional Release Note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -140,6 +140,7 @@ type NodeAffinityPreset struct {
 }
 
 type SecondaryDisks struct {
+	Name             string
 	Size             resource.Quantity
 	StorageClassName string
 }
@@ -282,7 +283,7 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
 		config.DNSConfig = rawConfig.VirtualMachine.DNSConfig
 	}
 	config.SecondaryDisks = make([]SecondaryDisks, 0, len(rawConfig.VirtualMachine.Template.SecondaryDisks))
-	for _, sd := range rawConfig.VirtualMachine.Template.SecondaryDisks {
+	for i, sd := range rawConfig.VirtualMachine.Template.SecondaryDisks {
 		sdSizeString, err := p.configVarResolver.GetConfigVarStringValue(sd.Size)
 		if err != nil {
 			return nil, nil, fmt.Errorf(`failed to parse "secondaryDisks.size" field: %w`, err)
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -535,8 +534,8 @@ func (p *provider) Create(ctx context.Context, machine *clusterv1alpha1.Machine,
 
 	// we need this check until this issue is resolved:
 	// https://github.com/kubevirt/containerized-data-importer/issues/895
-	if len(dataVolumeName) > 63 {
-		return nil, fmt.Errorf("dataVolumeName size %v, is bigger than 63 characters", len(dataVolumeName))
+	if len(c.DataVolumeName) > 63 {
+		return nil, fmt.Errorf("dataVolumeName size %v, is bigger than 63 characters", len(c.DataVolumeName))
```
ahmedwaleedmalik, line 538: I wonder if we really need this check since the name for resources in Kubernetes is already limited to 63 characters. So `machine.name` will never exceed this limit.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -101,6 +101,7 @@ type Config struct {
 	PodAffinityPreset     AffinityType
 	PodAntiAffinityPreset AffinityType
 	NodeAffinityPreset    NodeAffinityPreset
+	DataVolumeName        string
```
mfranczy, line 104: why do we need this `DataVolumeName`? This is too vague.

mfranczy (changes requested):

pkg/cloudprovider/provider/kubevirt/provider.go
mfranczy: better is to just extend the secondary disks struct with name and add the naming when parse config

mfranczy: https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/kubevirt/provider.go#L299-L302

sankalp-r: > https://github.com/kubermatic/machine-controller/blob/master/pkg/cloudprovider/provider/kubevirt/provider.go#L299-L302

I thought of a similar way, but inside `getConfig`, we don't have access to `machineName`.

mfranczy (approved):
I think we have to re-work the whole MC KubeVirt provider and add tests for it. I will open a dedicated ticket for this. I am ok to merge that because it's a hot-fix.

/lgtm
/approve

ahmedwaleedmalik (approved):
/approve

PR #1390, opened 2022-08-01
Fix IPv6 ExternalIP in machine object to align with hetzner ccm

sachintiptur:
Signed-off-by: Sachin Tiptur <sachin@kubermatic.com>

In KKP, "kubectl logs" fail on dual-stack user cluster with Hetzner cloud provider with Konnectivity enabled (see #[10534](https://github.com/kubermatic/kubermatic/issues/10534)).
The External IPv6 address format in the machine object is not aligned properly according to the Hetzner node. Due to this
csr approval fails in machine-controller and in turn certificate validation fails when kubectl logs command is executed.
This PR fixes this by assigning the IPv6 address in proper format to align with Hetzner ccm.

Partially: #[10534](https://github.com/kubermatic/kubermatic/issues/10534)

This fixes only the Konnectivity enabled case of the issue. 
Konnectivity disabled case needs a fix in hcloud-cloud-controller-manager

**Optional Release Note**:

```release-note
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #1393, opened 2022-08-03
set IPv6AcceptRA=true for Flatcar

PratikDeoghare:
Fixes network config for Flatcar to work with dual-stack. 

Fixes https://github.com/flatcar-linux/Flatcar/issues/384, https://github.com/kubermatic/kubermatic/issues/9798

**Optional Release Note**:

```release-note
NONE
```


pkg/userdata/flatcar/provider.go
```diff
@@ -374,6 +376,13 @@ storage:
         inline: |
 {{ .NodeIPScript | indent 10 }}
 
+    - path: "/etc/systemd/network/zz-default.network.d/ipv6-fix.conf"
+      filesystem: root
+      mode: 0755
+      contents:
+        inline: |
+{{ .IPv6RouterRA | indent 10 }}
```
ahmedwaleedmalik, line 384: Do we really need a method/template for this? Seems like there is nothing dynamic and we can just simply hardcode those values here.

ahmedwaleedmalik, line 384: Do we want to apply this unconditionally? Can we use [static-nic.network]( https://github.com/kubermatic/machine-controller/blob/master/pkg/userdata/flatcar/provider.go#L176) for this?

pkg/userdata/flatcar/provider.go
```diff
@@ -374,6 +376,13 @@ storage:
         inline: |
 {{ .NodeIPScript | indent 10 }}
 
+    - path: "/etc/systemd/network/zz-default.network.d/ipv6-fix.conf"
```
ahmedwaleedmalik, line 379: Please add the link of the flatcar issue https://github.com/flatcar-linux/Flatcar/issues/384 for context. Also please add a one-liner here regarding why this is required.

pkg/userdata/flatcar/provider.go
PratikDeoghare: Good idea.

PratikDeoghare: > Do we want to apply this unconditionally?

Yes. Because it harms nothing. We could apply it only in IPv6 or IPv4+IPv6 mode but not worth adding complexity to the code I think.

> Can we use [static-nic.network](https://github.com/kubermatic/machine-controller/blob/master/pkg/userdata/flatcar/provider.go#L176) for this?

Afaik it should work even if we put it there.

It will apply only to the interfaces with names matching `en*`.   So I put it here where it applies to any interface.

ahmedwaleedmalik: Okay makes sense 👍

ahmedwaleedmalik (approved):
/approve

PR #1396, opened 2022-08-04
Add `publicIPSKU` field to Azure provider

### LINKED ISSUE #1395: Allow passing IP address SKU for Azure
At the moment, the SKU for public IP addresses is determined based on whether dualstack is enabled or not:

https://github.com/kubermatic/machine-controller/blob/a7237249d5dc671f96719d97529da56caee022ba/pkg/cloudprovider/provider/azure/provider.go#L585-L595

Unfortunately, that is not the only scenario in which the SKU should be set. When using the `standard` load balancer SKU for your cluster's load balancing, enabling public IP addresses results in unusable load balancers (see https://github.com/kubermatic/kubermatic/issues/10673). That's because SKUs on load balancer and public IP addresses don't match.

As such, we should expose the public IP address SKU as a field in the provider spec. The field should be a pointer so it can remain unset, and if dualstack is enabled, the basic SKU should be rejected by validation.

embik:

The public IP address SKU wasn't configurable before. This adds it, plus some validation around the field.

Fixes #1395

**Optional Release Note**:

```release-note
Add new field `publicIPSKU` for Azure provider that configures the public IP address SKU (can be either `basic` or `standard`)
```

### DIFF
--- a/pkg/cloudprovider/provider/azure/provider.go
+++ b/pkg/cloudprovider/provider/azure/provider.go
@@ -24,6 +24,8 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2021-11-01/compute"
 	"github.com/Azure/azure-sdk-for-go/services/network/mgmt/2021-05-01/network"
@@ -102,6 +104,7 @@ type config struct {
 	DataDiskSKU  *compute.StorageAccountTypes
 
 	AssignPublicIP              bool
+	PublicIPSKU                 *network.PublicIPAddressSkuName
 	EnableAcceleratedNetworking *bool
 	Tags                        map[string]string
 }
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -1194,6 +1222,21 @@ func storageTypePtr(storageType string) *compute.StorageAccountTypes {
 	return &storage
 }
 
+func ipSkuPtr(ipSKU string) *network.PublicIPAddressSkuName {
```
PratikDeoghare, line 1225: Not a blocker. 

nit: This looks too complicated for what it does. 

https://pkg.go.dev/strings#Title

It is deprecated but probably okay for us. Or use golang.org/x/text/cases suggested there instead. 

--- 

Maybe we could just refuse improperly typed sku during validation.

PratikDeoghare (approved):

pkg/cloudprovider/provider/azure/provider.go
embik: I didn't want to use the deprecated method because, well, it's deprecated, and I didn't want to introduce an additional package either.

IMHO we shouldn't reject the SKUs without title case because we allow it in KKP as well.

moadqassem (approved):
/approve
/lgtm

PR #1402, opened 2022-08-11
Anexia: extend disk configuration

LittleFox94:

This extends the configuration for disks in provider Anexia. The important bit is the now-configurable disk performance type, but the config was revamped a bit to allow configuring multiple disks in the future.

As a side-effect, this PR also fixes CI for Anexia by disabling OperatingSystemManager for now as it is not yet supported.

In the first commit I did some cleanup of the config structs. This inflates the PR diff a bit, but it's separate commits at least.

The API changes are currently backwards compatible, so no "action required" needed for now.

**Optional Release Note**:

```release-note
Allow configuring the disk performance type for provider Anexia
```

### DIFF
--- a/.prow/provider-anexia.yaml
+++ b/.prow/provider-anexia.yaml
@@ -32,6 +32,10 @@ presubmits:
             - "./hack/ci/run-e2e-tests.sh"
           args:
             - "TestAnexiaProvisioningE2E"
+          env:
+            # OperatingSystemManager does not yet support Anexia
+            - name: OPERATING_SYSTEM_MANAGER
+              value: "false"
           securityContext:
             privileged: true
           resources:
--- a/examples/anexia-machinedeployment.yaml
+++ b/examples/anexia-machinedeployment.yaml
@@ -36,7 +36,15 @@ spec:
             locationID: "<< ANEXIA_LOCATION_ID >>"
             cpus: 2
             memory: 2048
-            diskSize: 60
+
+            # only a single disk is currently supported, but support for multiple disks is planned already
+            disks:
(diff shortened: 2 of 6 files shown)


LittleFox94:
CodeClimate isn't too happy about the new `resolveConfig`, has too many return statements. I think it's ok though, as the reason for this function to exist is resolving the config values and return errors when any are missing - in the end much like `Validate`

LittleFox94:
Hm, e2e test for Anexia fails due to not supporting OperatingSystemManager

what would be needed to support it? Should I add a commit to fixing CI by adding environment variable `OPERATING_SYSTEM_MANAGER=false` to prow config for anexia?

ahmedwaleedmalik:
> what would be needed to support it? Should I add a commit to fixing CI by adding environment variable OPERATING_SYSTEM_MANAGER=false to prow config for anexia?

Yes please, Anexia is not supported with OSM at the moment.

LittleFox94:
all of those failures are timeouts again

LittleFox94:
But since I only touched code for Anexia provider and that test went through, I'd say this can be merged anyway?

ahmedwaleedmalik:
Just ignore those failures. I'll review this PR today hopefully.

LittleFox94:
Alright, thank you :)

ahmedwaleedmalik (approved):
/approve

LittleFox94:
As written in [this comment](https://github.com/kubermatic/machine-controller/pull/1405#issuecomment-1220500780), we need an accompanying change in kubermatic for this - I will prepare a PR later today.

PR #1404, opened 2022-08-15
improve kubeconfig/gcp handling of credentials in e2e tests

xrstf:
This PR ensures that we always inject a base64-encoded kubeconfig in the kubevirt config, even if (in the future) an unencoded kubeconfig was provided in the e2e tests / via env vars. The same careful handling was added to the GCP SA, which also needs to be base64-encoded, which was never mentioned anywhere.

*Nothing changes for the users. This only affects our testing.*

/kind cleanup
```release-note
NONE
```

### DIFF
--- a/test/e2e/provisioning/all_e2e_test.go
+++ b/test/e2e/provisioning/all_e2e_test.go
@@ -720,8 +720,9 @@ func TestGCEProvisioningE2E(t *testing.T) {
 	// Act. GCE does not support CentOS.
 	selector := OsSelector("ubuntu")
 	params := []string{
-		fmt.Sprintf("<< GOOGLE_SERVICE_ACCOUNT >>=%s", googleServiceAccount),
+		fmt.Sprintf("<< GOOGLE_SERVICE_ACCOUNT_BASE64 >>=%s", safeBase64Encoding(googleServiceAccount)),
 	}
+
 	runScenarios(t, selector, params, GCEManifest, fmt.Sprintf("gce-%s", *testRunIdentifier))
 }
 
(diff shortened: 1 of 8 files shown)


xmudrii (approved):
/lgtm
/approve

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -709,7 +721,9 @@ func TestGCEProvisioningE2E(t *testing.T) {
 	selector := OsSelector("ubuntu")
 	params := []string{
 		fmt.Sprintf("<< GOOGLE_SERVICE_ACCOUNT >>=%s", googleServiceAccount),
```
xmudrii, line 723: Do we need this any longer?

ahmedwaleedmalik:
> This PR ensures that we always inject a base64-encoded kubeconfig in the kubevirt config, even if (in the future) an unencoded kubeconfig was provided (which is what 99% of users will have at hand)

This statement is quite misleading. You are not mentioning that your change is only applicable to our "E2E tests". So for a normal user, this will have no effect.

xrstf:
> So for a normal user, this will have no effect.

Yes, and that's a feature. I did not intend to break people's stuff. I only intend to fix our e2e stuff.

xmudrii (approved):
/lgtm
/approve

ahmedwaleedmalik:
> > So for a normal user, this will have no effect.
> 
> Yes, and that's a feature. I did not intend to break people's stuff. I only intend to fix our e2e stuff.

Please update the PR description and be more explicit about it. Right now it's misleading as I just mentioned above.

xmudrii (approved):

ahmedwaleedmalik (approved):
/approve

PR #1422, opened 2022-09-01
Update to Go 1.19

ahmedwaleedmalik:

Fixes #

**What type of PR is this?**

/kind feature

```release-note
Update to Go 1.19
```

**Documentation**:

```documentation
NONE
```


go.mod
```diff
@@ -45,13 +45,13 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 	// Please ensure that you update the image tags in `examples/operating-system-manager.yaml` as well.
 	k8c.io/operating-system-manager v1.0.0
-	k8s.io/api v0.24.2
+	k8s.io/api v0.25.0
```
kron4eg, line 48: is it really necessary to upgrade k8s deps?

ahmedwaleedmalik: Why shouldn't we?

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/metadata/client.go
```diff
@@ -54,7 +55,7 @@ type AuthConfig struct {
 }
 
 type Client interface {
-	GetMachineMetadata() (*MachineMetadata, error)
+	GetMachineMetadata(ctx context.Context) (*MachineMetadata, error)
```
kron4eg, line 58: Please don't break public APIs like that in a "simple go upgrade" PR. At least let's have a dedicated PR (or better not break at all).

go.mod
kron4eg: Because MC doesn't exist in isolation and other projects (who import MC) would be forced to upgrade too, which is not always optimal or wanted.

ahmedwaleedmalik: Well, I don't see any harm in upgrading them as well. Since we need to keep them updated to the latest packages anyways :)

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/metadata/client.go
ahmedwaleedmalik: Good point, basically it was to avoid a linter error. Let me re-evalute this.

go.mod
kron4eg: >we need to keep them updated to the latest packages anyways 

false

ahmedwaleedmalik: So for KKP and KubeOne, we don't want to use the latest deps for k8s API in the master? What's the reason for not wanting to keep them updated?

ahmedwaleedmalik: Anyways, if someone wants to use old deps then they can always pin them. :)

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/metadata/client.go
ahmedwaleedmalik: Updated PTAL, I'll change the API in a follow up PR.

kron4eg (approved):

PR #1424, opened 2022-09-01
Fix to mitigate race condition while configuring kubelet node ips

sachintiptur:
Signed-off-by: Sachin Tiptur <sachin@kubermatic.com>

In case of Dualstack, mitigate the timing issue between interface IPv6 assignment and kubelet service bringup.
This is achieved by adding a check for IPv6 address while configuring env variables for kubelet service. If the IPv6 is not yet assigned, return an error, which in-turn triggers the restart of setup and kubelet service.

Fixes #[10871](https://github.com/kubermatic/kubermatic/issues/10871)

**What type of PR is this?**

Verified for below combinations:

- Equinix - Rockylinux
- Hetzner - Ubuntu

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #1428, opened 2022-09-09
Rework OSM integration into generic external bootstrap "interface"

embik:

This PR is essentially both proposal and code change in one go. Let me explain the reasoning behind this PR first:

Essentially, we have a consumer/producer relationship between MC (as consumer) and OSM (as producer) when it comes to the user-data passed to new machines. However, I believe the consumer/producer relationship bits are a bit all over the place. Some logic exists in MC, some in OSM. The consumer/producer relationship here is very similar to Go interfaces, and a rule of thumb there is that interfaces should not be aware of types that implement them.

The same, IMHO, goes for the MC-OSM relationship: MC should support a standard (or interface, if you want; `pkg/bootstrap` defines this in this PR) that any producer of bootstrapping user-data can "slot" into. But MC should not be aware of the producer component in any way, other than the defined interface (which, in this PR, is a `Secret` created under a certain naming pattern). Theoretically, you should be able to create bootstrapping user-data manually.

The point of all this is that MC should not be aware of OSM or depend on it. If this is accepted, I plan to follow up in OSM with adding some of the logic removed here back into it.

In addition, the current MC code does a referential integrity check on an OSP resource - with all the previous explanation this does not make sense anymore anyway (as it's implementation-specific for OSP), but we do not apply referential integrity checks anywhere else for a specific reason: It's against Kubernetes design principles (you won't find this in HPAs referencing a non-existing Deployment, or RBAC role bindings referencing non-existing roles). So, I propose that the check for OSP is removed and replaced with events on `MachineDeployments`, highlighting that the referenced OSP does not (yet) exist. But that is for the follow-up PR in OSM.

TL;DR: This PR does the following things:

- Remove the code dependency on OSM to stop the dependency loop we have between MC and OSM.
- Remove OSM-specific validation and mutation.
- Document "external bootstrap" mechanism in `pkg/bootstrap` that relies on some third-party component (such as OSM) creating a user-data bootstrap secret (following the same pattern that OSM applies right now).

Note: The OSP annotation magic should move to KKP, I'll follow up on that as well.

Fixes #

**What type of PR is this?**
/kind cleanup

```release-note
Remove dependency on operating-system-manager (OSM) and instead rely on generic external bootstrap secret "interface" (which can be fulfilled by any provider)
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/cmd/machine-controller/main.go
+++ b/cmd/machine-controller/main.go
@@ -41,7 +41,6 @@ import (
 	"github.com/kubermatic/machine-controller/pkg/health"
 	machinesv1alpha1 "github.com/kubermatic/machine-controller/pkg/machines/v1alpha1"
 	"github.com/kubermatic/machine-controller/pkg/node"
-	osmv1alpha1 "k8c.io/operating-system-manager/pkg/crd/osm/v1alpha1"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -72,7 +71,8 @@ var (
 	enableLeaderElection             bool
 	leaderElectionNamespace          string
 
-	useOSM bool
+	useOSM               bool
+	useExternalBootstrap bool
 
 	nodeCSRApprover                   bool
 	nodeHTTPProxy                     string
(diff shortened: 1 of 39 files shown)


ahmedwaleedmalik (approved):
Great job @embik, this brings us a lot closer to what we wanted to achieve; remove this circle of dependency between OSM and MC. 

/approve 💙

PR #1438, opened 2022-09-21
Add enableBootDiagnostics field to Azure cloud spec

### LINKED ISSUE #1291: Support (optional) boot diagnostics for Azure VMs
Sometimes, it's beneficial to have [boot diagnostics](https://docs.microsoft.com/en-us/troubleshoot/azure/virtual-machines/boot-diagnostics) for your Azure VMs. This can help identifying why your nodes are not properly bootstrapping in case you do not have direct network access to the VM or your SSH key is not applied for some unknown reason. You cannot access the serial console without boot diagnostics enabled.

This needs a Storage Account to store boot logs in, but creating the Storage Account would happen in KKP, so we do not need to take care of that in machine-controller.

**Acceptance Criteria**

- Add a new field `enableBootDiagnostics` that enables managed boot diagnostics (default and recommended on Azure).

embik:
Boot diagnostics are nice because you can see the console logs if things go wrong and you cannot SSH into your VM for whatever reason. That can be useful for our development, but I imagine users who face stability issues with their Azure VMs might also want to enable this setting.

Azure [enables this setting by default on VMs created via the portal](https://learn.microsoft.com/en-us/azure/virtual-machines/boot-diagnostics#enable-managed-boot-diagnostics-using-the-azure-portal), so it seems a good idea to at least provide a setting for this.

Fixes #1291

**What type of PR is this?**
/kind feature
/kind api-change

```release-note
Add `enableBootDiagnostics` field to Azure cloud spec to enable boot diagnostics on managed storage account
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/azure/provider.go
+++ b/pkg/cloudprovider/provider/azure/provider.go
@@ -106,6 +106,7 @@ type config struct {
 	AssignPublicIP              bool
 	PublicIPSKU                 *network.PublicIPAddressSkuName
 	EnableAcceleratedNetworking *bool
+	EnableBootDiagnostics       bool
 	Tags                        map[string]string
 }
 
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (approved):
/approve

PR #1445, opened 2022-09-26
add support for topology spread constraint for VMs

sankalp-r:
This PR adds support for `TopologySpreadConstraint` for VMs which will spread user cluster nodes across infra-cluster nodes. And removes the Pod affinity/anti-affinity mechanism.

Fixes https://github.com/kubermatic/kubermatic/issues/10646

**What type of PR is this?**

/kind feature

```release-note
Add support for topology spread constraint for KubeVirt VirtualMachines.
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,7 +57,7 @@ require (
 	k8s.io/klog v1.0.0
 	k8s.io/kubelet v0.24.2
 	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
-	kubevirt.io/api v0.54.0
+	kubevirt.io/api v0.57.1
 	kubevirt.io/containerized-data-importer-api v1.50.0
 	sigs.k8s.io/controller-runtime v0.12.1
 	sigs.k8s.io/yaml v1.3.0
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -86,21 +86,22 @@ func New(configVarResolver *providerconfig.ConfigVarResolver) cloudprovidertypes
 }
 
 type Config struct {
-	Kubeconfig            string
-	RestConfig            *rest.Config
-	DNSConfig             *corev1.PodDNSConfig
-	DNSPolicy             corev1.DNSPolicy
-	CPUs                  string
-	Memory                string
-	Namespace             string
-	OsImage               OSImage
-	StorageClassName      string
-	PVCSize               resource.Quantity
-	FlavorName            string
-	SecondaryDisks        []SecondaryDisks
-	PodAffinityPreset     AffinityType
-	PodAntiAffinityPreset AffinityType
-	NodeAffinityPreset    NodeAffinityPreset
+	Kubeconfig                string
+	RestConfig                *rest.Config
+	DNSConfig                 *corev1.PodDNSConfig
+	DNSPolicy                 corev1.DNSPolicy
+	CPUs                      string
+	Memory                    string
+	Namespace                 string
+	OsImage                   OSImage
+	StorageClassName          string
+	PVCSize                   resource.Quantity
+	FlavorName                string
+	SecondaryDisks            []SecondaryDisks
+	PodAffinityPreset         AffinityType
+	PodAntiAffinityPreset     AffinityType
+	NodeAffinityPreset        NodeAffinityPreset
+	TopologySpreadConstraints []corev1.TopologySpreadConstraint
 }
 
 type AffinityType string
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -57,7 +57,7 @@ require (
 	k8s.io/klog v1.0.0
 	k8s.io/kubelet v0.24.2
 	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
-	kubevirt.io/api v0.54.0
+	kubevirt.io/api v0.57.0
```
mfranczy, line 60: could you import v0.57.1?

pkg/cloudprovider/provider/kubevirt/types/types.go
```diff
@@ -87,6 +90,13 @@ type NodeAffinityPreset struct {
 	Values []providerconfigtypes.ConfigVarString `json:"values,omitempty"`
 }
 
+// TopologySpreadConstraint.
+type TopologySpreadConstraint struct {
```
mfranczy, line 94: Please add comments to fields. I know we didn't do that in the past but it's time to change this.

pkg/cloudprovider/provider/kubevirt/types/types.go
```diff
@@ -75,7 +76,9 @@ type Disk struct {
 
 // Affinity.
 type Affinity struct {
-	PodAffinityPreset     providerconfigtypes.ConfigVarString `json:"podAffinityPreset,omitempty"`
+	// Deprecated
+	PodAffinityPreset providerconfigtypes.ConfigVarString `json:"podAffinityPreset,omitempty"`
```
mfranczy, line 80: please add explanation why it's deprecated now. Should be
`// Deprecated: <description>`

We deprecate that field because KubeVirt supports now topology spread constraints.

mfranczy:
@sankalp-r do you consider this PR ready or are you still working on it?

sankalp-r:
> @sankalp-r do you consider this PR ready or are you still working on it?

PR is ready from my perspective.

mfranczy (approved):
/approve
/lgtm

mfranczy:
@sankalp-r as I understand, prior to this change if somebody uses PodAffinity then later when recreating VM it will have applied topology spread constraints, right? Please don't forget to test this behaviour.

sankalp-r:
> @sankalp-r as I understand, prior to this change if somebody uses PodAffinity then later when recreating VM it will have applied topology spread constraints, right? Please don't forget to test this behaviour that migration is smooth.

@mfranczy Before this change assume if we have a MachineDeployments with PodAffnity, corresponding VMs will also have those.
After this change, the existing MD and corresponding VM stay the same, meaning PodAffnity spec remains in the VM.
But if now we try to scale the existing old MD to more replicas, new VMs will be created and they will not have PodAffnity, but now they will have default topology constraints. The old existing MD object will still persist those Affinity specs until we remove them.

sankalp-r:
> @sankalp-r as I understand, prior to this change if somebody uses PodAffinity then later when recreating VM it will have applied topology spread constraints, right? Please don't forget to test this behaviour that migration is smooth.

@mfranczy Before this change assume if we have a MachineDeployments with PodAffnity, corresponding VMs will also have those.
After this change, the existing MD and corresponding VM stay the same, meaning PodAffnity spec remains in the VM.
But if now we try to scale the existing old MD to more replicas, new VMs will be created and they will not have PodAffnity(because it is ignored), but now they will have default topology constraints. The old existing MD object will still persist those Affinity specs until we remove them.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -845,37 +846,21 @@ func getAffinity(config *Config, matchKey, matchValue string) *corev1.Affinity {
...
+		return config.TopologySpreadConstraints
 	}
-}
-
-func addPrefixToSecondaryDisk(secondaryDisks []SecondaryDisks, prefix string) {
-	for i := range secondaryDisks {
-		secondaryDisks[i].Name = fmt.Sprintf("%s-%s", prefix, secondaryDisks[i].Name)
+	// Return default TopologySpreadConstraint
+	return []corev1.TopologySpreadConstraint{
+		{MaxSkew: 1, TopologyKey: topologyKeyHostname, WhenUnsatisfiable: corev1.ScheduleAnyway, LabelSelector: &metav1.LabelSelector{MatchLabels: matchLabels}},
```
mfranczy, line 864: just a nit, could you fix the formatting of this structure? It will be more readable if it's not a one liner.

ahmedwaleedmalik (approved):
/approve

PR #1450, opened 2022-09-29
Add support for OpenNebula as a cloud provider

nilsding:

This PR adds support for [OpenNebula](https://opennebula.io/) as a cloud provider.  Similar to OpenStack it can be hosted on your own infrastructure.

We're using OpenNebula for many years already, and adding (basic?) support for it seemed to be easy enough.

**What type of PR is this?**
/kind feature

I'm not too fond of GoLang, so if there's anything that's not idiomatic please point it out.

Also I would still like to figure out automated end-to-end tests, any support with that is welcome.  I tested it manually on a small kubernetes cluster (just some basic creation of the VMs) and it seems to work fine so far.

```release-note
Add support for OpenNebula as a cloud provider
```

**Documentation**:

```documentation
TBD
```
(?)


akurz:
Hi, 

we (adidas Runtastic) are testing this provider now for some month in some small user-clusters and it works fine for us. A review and approval would be much aprechiated @xrstf or @ahmedwaleedmalik - Thanks in advance! 

Cheers,
Andreas

test/e2e/provisioning/all_e2e_test.go
```diff
@@ -964,6 +965,43 @@ func TestNutanixProvisioningE2E(t *testing.T) {
...
+	if oneNetwork == "" {
+		oneNetwork = "vnet"
+	}
+
+	params := []string{
+		fmt.Sprintf("<< ONE_ENDPOINT >>=%s", oneEndpoint),
+		fmt.Sprintf("<< ONE_USERNAME >>=%s", oneUsername),
+		fmt.Sprintf("<< ONE_PASSWORD >>=%s", onePassword),
+		fmt.Sprintf("<< YOUR_DATASTORE_NAME >>=%s", oneDatastore),
+		fmt.Sprintf("<< YOUR_NETWORK_NAME >>=%s", oneNetwork),
```
embik, line 998: Would it be an option to also prefix these with `ONE_` instead of `YOUR_`, so we know that these variable belong to OpenNebula setups?

test/e2e/provisioning/testdata/machinedeployment-opennebula.yaml
```diff
@@ -0,0 +1,55 @@
...
+          cloudProviderSpec:
+            endpoint: "<< ONE_ENDPOINT >>"
+            username: "<< ONE_USERNAME >>"
+            password: "<< ONE_PASSWORD >>"
+
+            cpu: 1
+            vcpu: 2
+            memory: 1024
+
+            image: "<< OS_IMAGE >>"
```
embik, line 35: similar to the other comment, can we maybe make sure that this is prefixed to be specific for OpenNebula? e.g. `ONE_OS_IMAGE`? This way we can avoid collision with the OpenStack image reference.

pkg/cloudprovider/provider/opennebula/provider.go
```diff
@@ -0,0 +1,430 @@
...
+	tpl.AddCtx(machineUidContextKey, string(machine.UID))
+	tpl.AddCtx("USER_DATA", base64.StdEncoding.EncodeToString([]byte(userdata)))
+	tpl.AddCtx("USER_DATA_ENCODING", "base64")
+
+	controller := goca.NewController(client)
+
+	// create VM from the generated template above
+	vmID, err := controller.VMs().Create(tpl.String(), false)
+	if err != nil {
+		return nil, err
```
embik, line 217: I think it would be helpful to contextualise the error message returned by the upstream SDK here, e.g. with this:

```suggestion
		return nil, fmt.Errorf("failed to create VM from template: %w", err)
```

pkg/cloudprovider/provider/opennebula/provider.go
```diff
@@ -0,0 +1,430 @@
...
+	}
+
+	vm, err := controller.VM(vmID).Info(false)
+	if err != nil {
+		return nil, err
```
embik, line 222: Same here. I'm not _exactly_ sure what this does, but the error should probably look something like this:

```suggestion
		return nil, fmt.Errorf("failed to fetch VM information: %w", err)
```

pkg/cloudprovider/provider/opennebula/provider.go
```diff
@@ -0,0 +1,430 @@
...
+		return false, cloudprovidererrors.TerminalError{
+			Reason:  common.InvalidConfigurationMachineError,
+			Message: fmt.Sprintf("Failed to parse MachineSpec, due to %v", err),
+		}
+	}
+
+	client := getClient(c)
+	controller := goca.NewController(client)
+
+	vmctrl := controller.VM(instance.(*openNebulaInstance).vm.ID)
```
embik, line 248: I have to say I'm not a huge fan of the type assertions here and later in the code since it can introduce run-time panics. I understand it's required to get the OpenNebula specific information, but e.g. the AWS provider implements a private version of `get` to return the provider-specific instance: https://github.com/kubermatic/machine-controller/blob/df72929d4011f49d39f4f5f78208856ebba2ae99/pkg/cloudprovider/provider/aws/provider.go#L877

Have you considered that approach?

embik (changes requested):
Overall this looks good, just a few comments that I think could be addressed.

.gitignore
```diff
@@ -16,3 +16,4 @@ examples/*.srl
 .vscode
 .gitpod.yml
 cmd/machine-controller/__debug_bin
+!pkg
```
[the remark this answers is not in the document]
nilsding: I left it in, since I mainly work with Ruby and `pkg/` is the default output directory for built gems (packages), so it's in my global gitignore.  Can remove it from here if it's unwanted, but it caused some initial headaches for me figuring out why git didn't want see any changes there :dizzy_face:

pkg/cloudprovider/provider/opennebula/provider.go
nilsding: No, this hasn't come across my mind, will try to adapt this then.

.gitignore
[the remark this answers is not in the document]
embik: Hm. I guess it's fine, I wasn't aware of this being a build directory for gems.

embik:
@nilsding please address the [linter issues](https://public-prow.loodse.com/view/gs/prow-dev-public-data/pr-logs/pull/kubermatic_machine-controller/1450/pull-machine-controller-lint/1653288673640189952) as well. The PR should be good to go afterwards.

embik (approved):
/approve

embik:
Thank you for your contribution, @nilsding!

PR #1459, opened 2022-10-11
Use Calico as CNI for e2e and disable node readiness checks

embik:

This is a bit complicated to explain but essentially the e2e tests running out of kind were only somewhat working, but flannel "covered" that up a bit. The problem is between the kind control plane and the joining nodes; CNI pods cannot talk to the Kubernetes API endpoint as the `Endpoints` for the "kubernetes" `Service` is an internal IP nodes will not be able to reach. This also affects `kube-proxy` never being able to properly start.

For some reason, flannel still marked the nodes as `Ready` (although technically they weren't), same for Cilium, but none _really_ worked for CNI and flannel broke during the CI cluster upgrade (machine-controller-webhook was no longer able to resolve DNS due to service IPs no longer properly routing).

This PR moves to Calico to bring back the kind control plane, but removes the readiness check for a node (since no node can ever really be initialised by CNI in our test setup) and resorts to just making sure a Node object exists for a machine. A follow-up issue was raised at #1462.

Fixes #

**What type of PR is this?**
/kind cleanup
/kind regression

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/.prow/e2e-features.yaml
+++ b/.prow/e2e-features.yaml
@@ -106,7 +106,7 @@ presubmits:
               memory: 7Gi
 
   - name: pull-machine-controller-e2e-deployment-upgrade
-    always_run: true
+    always_run: false
     decorate: true
     clone_uri: "ssh://git@github.com/kubermatic/machine-controller.git"
     labels:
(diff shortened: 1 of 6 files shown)


xmudrii (approved):
/lgtm
/approve

PR #1467, opened 2022-10-24
Extend containerd registry mirror config

LittleFox94:

containerd can be configured for multiple registry mirrors per registry, this extends machine-controller to configure mirrors for registries other than docker.io, reusing the existing command line flag.

**What type of PR is this?**
/kind feature

This should probably be documented, but didn't find where..

```release-note
Allow configuring containerd registry mirrors not only for docker.io
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/containerruntime/config.go
+++ b/pkg/containerruntime/config.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"regexp"
 	"strings"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


pkg/containerruntime/config.go
```diff
@@ -45,27 +46,38 @@ func BuildConfig(opts Opts) (Config, error) {
 		}
 	}
 
-	var registryMirrors []string
+	// we want to match e.g. docker.io=registry.docker-cn.com, having docker.io as the first
```
LittleFox94, line 49: I initially worked with `:` instead if `=`, so `docker.io:registry.docker-cn.com`, which looks a bit nicer to me - but there are already flags with `Key=Value` so matching those is probably best?

ahmedwaleedmalik (approved):
/approve

PR #1470, opened 2022-10-31
add http source for kubevirt VM Disk Image

sankalp-r:
This PR introduces `PrimarDiskImageSource` for the KubeVirt provider. The source can be `HTTP` or `PVC-Clone`.
If the source is `HTTP` with `preAllocatedDataVolume` as false, the VM disk-image will be downloaded from the HTTP source.
If the source is `HTTP` with `preAllocatedDataVolume` as true, the VM disk-image will be downloaded just once to create PreAllocatedDataVolume and VM Disks will be created by cloning over PreAllocatedDataVolume.
If the source is `PVC-Clone`, VM Disks will be created by cloning over custom-local-disks.

Fixes #https://github.com/kubermatic/kubermatic/issues/11234

**What type of PR is this?**

/kind feature

```release-note
Add HTTP source for  KubeVirt VM Disk images which enables caching to avoid downloading images.
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -60,12 +60,18 @@ func init() {
 	}
 }
 
+type imageSource string
+
 const (
 	// topologyKeyHostname defines the topology key for the node hostname.
 	topologyKeyHostname = "kubernetes.io/hostname"
 	// machineDeploymentLabelKey defines the label key used to contains as value the MachineDeployment name
 	// which machine comes from.
 	machineDeploymentLabelKey = "md"
+	// httpSource defines the http source type for VM Disk Image.
+	httpSource imageSource = "http"
+	// pvcSource defines the pvc source type for VM Disk Image.
+	pvcSource imageSource = "pvc"
 )
 
 var supportedOS = map[providerconfigtypes.OperatingSystem]*struct{}{
@@ -93,7 +99,7 @@ type Config struct {
 	CPUs                      string
 	Memory                    string
 	Namespace                 string
-	OsImage                   OSImage
+	OSImageSource             *cdiv1beta1.DataVolumeSource
 	StorageClassName          string
 	PVCSize                   resource.Quantity
 	FlavorName                string
@@ -146,11 +152,6 @@ type SecondaryDisks struct {
 	Name             string
 	Size             resource.Quantity
 	StorageClassName string
-}
-
-type OSImage struct {
-	URL            string
-	DataVolumeName string
 }
 
 type kubeVirtServer struct {
@@ -247,15 +248,12 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
 		return nil, nil, fmt.Errorf(`failed to get value of "memory" field: %w`, err)
 	}
 	config.Namespace = getNamespace()
-	osImage, err := p.configVarResolver.GetConfigVarStringValue(rawConfig.VirtualMachine.Template.PrimaryDisk.OsImage)
+
+	config.OSImageSource, err = p.parseOSImageSource(rawConfig.VirtualMachine.Template.PrimaryDisk, config.Namespace)
 	if err != nil {
-		return nil, nil, fmt.Errorf(`failed to get value of "sourceURL" field: %w`, err)
+		return nil, nil, fmt.Errorf(`failed to get value of "osImageSource" field: %w`, err)
 	}
-	if _, err = url.ParseRequestURI(osImage); err == nil {
-		config.OsImage.URL = osImage
-	} else {
-		config.OsImage.DataVolumeName = osImage
-	}
+
 	pvcSize, err := p.configVarResolver.GetConfigVarStringValue(rawConfig.VirtualMachine.Template.PrimaryDisk.Size)
 	if err != nil {
 		return nil, nil, fmt.Errorf(`failed to get value of "pvcSize" field: %w`, err)
@@ -276,9 +274,6 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
 	// Instancetype and Preference
 	config.Instancetype = rawConfig.VirtualMachine.Instancetype
 	config.Preference = rawConfig.VirtualMachine.Preference
-	if err != nil {
-		return nil, nil, fmt.Errorf(`failed to parse "preference" field: %w`, err)
-	}
 
 	dnsPolicyString, err := p.configVarResolver.GetConfigVarStringValue(rawConfig.VirtualMachine.DNSPolicy)
 	if err != nil {
@@ -375,6 +370,35 @@ func (p *provider) parseTopologySpreadConstraint(topologyConstraints []kubevirtt
 		})
 	}
 	return parsedTopologyConstraints, nil
+}
+
+func (p *provider) parseOSImageSource(primaryDisk kubevirttypes.PrimaryDisk, nameSpace string) (*cdiv1beta1.DataVolumeSource, error) {
+	osImage, err := p.configVarResolver.GetConfigVarStringValue(primaryDisk.OsImage)
+	if err != nil {
+		return nil, fmt.Errorf(`failed to get value of "primaryDisk.osImage" field: %w`, err)
+	}
+	osImageSource, err := p.configVarResolver.GetConfigVarStringValue(primaryDisk.Source)
+	if err != nil {
+		return nil, fmt.Errorf(`failed to get value of "primaryDisk.source" field: %w`, err)
+	}
+	switch imageSource(osImageSource) {
+	case httpSource:
+		return &cdiv1beta1.DataVolumeSource{HTTP: &cdiv1beta1.DataVolumeSourceHTTP{URL: osImage}}, nil
+	case pvcSource:
+		if namespaceAndName := strings.Split(osImage, "/"); len(namespaceAndName) >= 2 {
+			return &cdiv1beta1.DataVolumeSource{PVC: &cdiv1beta1.DataVolumeSourcePVC{Name: namespaceAndName[1], Namespace: namespaceAndName[0]}}, nil
+		}
+		return &cdiv1beta1.DataVolumeSource{PVC: &cdiv1beta1.DataVolumeSourcePVC{Name: osImage, Namespace: nameSpace}}, nil
+	default:
+		// handle old API for backward compatibility.
+		if _, err = url.ParseRequestURI(osImage); err == nil {
+			return &cdiv1beta1.DataVolumeSource{HTTP: &cdiv1beta1.DataVolumeSourceHTTP{URL: osImage}}, nil
+		}
+		if namespaceAndName := strings.Split(osImage, "/"); len(namespaceAndName) >= 2 {
+			return &cdiv1beta1.DataVolumeSource{PVC: &cdiv1beta1.DataVolumeSourcePVC{Name: namespaceAndName[1], Namespace: namespaceAndName[0]}}, nil
+		}
+		return &cdiv1beta1.DataVolumeSource{PVC: &cdiv1beta1.DataVolumeSourcePVC{Name: osImage, Namespace: nameSpace}}, nil
+	}
 }
 
 // getNamespace returns the namespace where the VM is created.
@@ -509,7 +533,11 @@ func (p *provider) MachineMetricsLabels(machine *clusterv1alpha1.Machine) (map[s
 	if err == nil {
 		labels["cpus"] = c.CPUs
 		labels["memoryMIB"] = c.Memory
-		labels["osImage"] = c.OsImage.URL
+		if c.OSImageSource.HTTP != nil {
+			labels["osImage"] = c.OSImageSource.HTTP.URL
+		} else if c.OSImageSource.PVC != nil {
+			labels["osImage"] = c.OSImageSource.PVC.Name
+		}
 	}
 
 	return labels, err
@@ -790,7 +818,6 @@ func getVMVolumes(config *Config, dataVolumeName string, userDataSecretName stri
 }
 
 func getDataVolumeTemplates(config *Config, dataVolumeName string) []kubevirtv1.DataVolumeTemplateSpec {
-	dataVolumeSource := getDataVolumeSource(config.OsImage)
 	pvcRequest := corev1.ResourceList{corev1.ResourceStorage: config.PVCSize}
 	dataVolumeTemplates := []kubevirtv1.DataVolumeTemplateSpec{
 		{
@@ -807,7 +834,7 @@ func getDataVolumeTemplates(config *Config, dataVolumeName string) []kubevirtv1.
 						Requests: pvcRequest,
 					},
 				},
-				Source: dataVolumeSource,
+				Source: config.OSImageSource,
 			},
 		},
 	}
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -93,7 +98,7 @@ type Config struct {
 	CPUs                      string
 	Memory                    string
 	Namespace                 string
-	OsImage                   OSImage
+	OsImage                   *OSImage
```
mfranczy, line 101: Should be OSImage on the left side as well.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -149,8 +154,18 @@ type SecondaryDisks struct {
 }
 
 type OSImage struct {
```
mfranczy, line 156: please add comments to each structure with explanation of the structutre itself and fields

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -149,8 +154,18 @@ type SecondaryDisks struct {
...
-	URL            string
+	HTTP     *HTTPSource
+	PVCClone *PVCSource
+}
+
+type HTTPSource struct {
+	URL                    string
+	PreAllocatedDataVolume bool
```
mfranczy, line 163: maybe let's rename to `CacheImage`

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -149,8 +154,18 @@ type SecondaryDisks struct {
...
+}
+
+type PVCSource struct {
 	DataVolumeName string
+	NameSpace      string
```
mfranczy, line 168: Should be named `Namespace`

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -377,6 +388,40 @@ func (p *provider) parseTopologySpreadConstraint(topologyConstraints []kubevirtt
...
 }
 
+func (p *provider) parseOSImage(primaryDisk kubevirttypes.PrimaryDisk, nameSpace string) (*OSImage, error) {
+	osImage, err := p.configVarResolver.GetConfigVarStringValue(primaryDisk.OsImage)
+	if err != nil {
+		return nil, fmt.Errorf(`failed to get value of "primaryDisk.osImage" field: %w`, err)
+	}
+
+	var parsedOSImage *OSImage
+	// If PrimaryDiskOSSource is specified
```
mfranczy, line 398: I would extend this comment and say that if source is specified then that means we deal with new API

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -377,6 +388,40 @@ func (p *provider) parseTopologySpreadConstraint(topologyConstraints []kubevirtt
...
+			preAllocatedDV, _, err := p.configVarResolver.GetConfigVarBoolValue(primaryDisk.Source.HTTP.PreAllocateDataVolume)
+			if err != nil {
+				return nil, fmt.Errorf(`failed to get value of "preAllocateDataVolume" field: %w`, err)
+			}
+			parsedOSImage = &OSImage{HTTP: &HTTPSource{URL: osImage, PreAllocatedDataVolume: preAllocatedDV}}
+		} else if primaryDisk.Source.PVCClone != nil {
+			parsedOSImage = &OSImage{PVCClone: &PVCSource{DataVolumeName: osImage, NameSpace: nameSpace}}
+		}
+	}
+	// If PrimaryDiskOSSource is not specified
```
mfranczy, line 410: We have to extend that comment and say that we are dealing here with older API and it's a matter of backward compatibility

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -651,13 +709,34 @@ func (p *provider) newVirtualMachine(ctx context.Context, c *Config, pc *provide
...
+		}
+		exist = false
+	}
+
+	// create preAllocated DataVolume, if it doesn't exist.
+	//  preAllocated DataVolume will be created per MachineDeployment.
+	if !exist {
+		return sigClient.Create(ctx, newDataVolume(c, dvName, machine.Name))
+	}
+	// annotate existing DataVolume with VM-name, if already exist.
```
mfranczy, line 733: this comment is not needed, the code is quite clear

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -651,13 +709,34 @@ func (p *provider) newVirtualMachine(ctx context.Context, c *Config, pc *provide
...
+	updatedDV := existingDV.DeepCopy()
```
mfranczy, line 734: any reason for doing a `DeepCopy` in here? Controller runtime get function will return already a deep copy.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -651,13 +709,34 @@ func (p *provider) newVirtualMachine(ctx context.Context, c *Config, pc *provide
...
+			DataVolumeTemplates: getDataVolumeTemplates(c, dataVolumeName, getDataVolumeSource(c, mdName)),
 		},
 	}
 	return virtualMachine, nil
 }
 
-func (p *provider) Cleanup(ctx context.Context, machine *clusterv1alpha1.Machine, _ *cloudprovidertypes.ProviderData) (bool, error) {
+func (p *provider) createPreAllocatedDataVolume(ctx context.Context, c *Config, machine *clusterv1alpha1.Machine, sigClient client.Client, dvName string) error {
+	existingDV := &cdiv1beta1.DataVolume{}
+	exist := true
```
mfranczy, line 720: I think you can get rid of this flag, it's not required. You can check if the object is empty.

pkg/cloudprovider/provider/kubevirt/types/types.go
```diff
@@ -67,6 +67,23 @@ type Template struct {
 type PrimaryDisk struct {
 	Disk
 	OsImage providerconfigtypes.ConfigVarString `json:"osImage,omitempty"`
+	Source  *Source                             `json:"source,omitempty"`
+}
+
+// Source.
+type Source struct {
```
mfranczy, line 74: We need better comments with explanation about the fields.

mfranczy (changes requested):
I did a first review round. I will repeat after new changes.

pkg/cloudprovider/provider/kubevirt/provider.go
mfranczy: Btw why it has to be a pointer now?

pkg/cloudprovider/provider/kubevirt/provider.go
mfranczy: the same with the below comment

pkg/cloudprovider/provider/kubevirt/provider.go
mfranczy: What I mean here is that you can return an error from Get and implement lower a check that if err is not found then create a new data volume.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -66,6 +69,8 @@ const (
 	// machineDeploymentLabelKey defines the label key used to contains as value the MachineDeployment name
 	// which machine comes from.
 	machineDeploymentLabelKey = "md"
+	// preAllocatedDVAnnotation defines the annotation for preAllocated DataVolume for cloning.
+	preAllocatedDVAnnotation = "datavolume.clone.kubevirt.io/machine"
```
mfranczy, line 73: let's change to `disk-clone.machines.cluster.k8s.io/machine-%s`

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -679,9 +759,39 @@ func (p *provider) Cleanup(ctx context.Context, machine *clusterv1alpha1.Machine
 		return true, nil
 	}
 
+	// cleanup preAllocated DataVolume
```
mfranczy, line 762: not needed comment

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -679,9 +759,39 @@ func (p *provider) Cleanup(ctx context.Context, machine *clusterv1alpha1.Machine
...
+			if !kerrors.IsNotFound(err) {
+				return err
+			}
+			return nil
+		}
+
+		annotationKey := fmt.Sprintf("%s-%s", preAllocatedDVAnnotation, machine.Name)
+		if dv.Annotations[annotationKey] == "true" {
+			// If the Machine-Deployment is scaled to zero or deleted or preAllocateDataVolume is disabled, remove preAllocated DataVolume.
+			if len(dv.Annotations) == 1 {
```
mfranczy, line 783: let's keep it safe and change to <= 1

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -679,9 +759,39 @@ func (p *provider) Cleanup(ctx context.Context, machine *clusterv1alpha1.Machine
...
+				return sigClient.Delete(ctx, dv)
+			}
+			// Remove annotation of corresponding VM from preAllocated DataVolume.
+			updatedDv := dv.DeepCopy()
```
mfranczy, line 787: you don't need deep copy

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -93,7 +98,7 @@ type Config struct {
 	CPUs                      string
 	Memory                    string
 	Namespace                 string
-	OsImage                   OSImage
+	OSImage                   *OSImage
```
mfranczy, line 101: Why it has to be a pointer now?

mfranczy, line 101: Why it has to be a pointer now?

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -540,8 +598,19 @@ func (p *provider) Create(ctx context.Context, machine *clusterv1alpha1.Machine,
 
 	userDataSecretName := fmt.Sprintf("userdata-%s-%s", machine.Name, strconv.Itoa(int(time.Now().Unix())))
 
+	var machineDeploymentName string
+	if mdName, err := machineDeploymentNameAndRevisionForMachineGetter(ctx, machine, data.Client)(); err == nil {
+		machineDeploymentName = mdName
+	}
```
mfranczy, line 604: I am not sure if I follow this logic. 
Why it cannot be:
```
mdName, err := machineDeploymentNameAndRevisionForMachineGetter(ctx, machine, data.Client)()
if err != nil {
    return nil, err
}
```

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -651,13 +718,26 @@ func (p *provider) newVirtualMachine(ctx context.Context, c *Config, pc *provide
...
 }
 
-func (p *provider) Cleanup(ctx context.Context, machine *clusterv1alpha1.Machine, _ *cloudprovidertypes.ProviderData) (bool, error) {
+func (p *provider) createPreAllocatedDataVolume(ctx context.Context, c *Config, machine *clusterv1alpha1.Machine, sigClient client.Client, dvName string) error {
+	existingDV := &cdiv1beta1.DataVolume{}
+	if err := sigClient.Get(ctx, types.NamespacedName{Namespace: c.Namespace, Name: dvName}, existingDV); err != nil {
+		if !kerrors.IsNotFound(err) {
+			return err
+		}
+		// create DV if it doesn't exist
```
mfranczy, line 733: I think we don't need comments like that, code is self explanatory

mfranczy (changes requested):
Another review round.

pkg/cloudprovider/provider/kubevirt/provider.go
sankalp-r: Notting specific, just `parseOSImage()` return pointer so I can directly assign it to `config.OSImage`

pkg/cloudprovider/provider/kubevirt/provider.go
sankalp-r: Before this change, we used to create VMs even if getting the MD-name returned error.
So continue with earlier behavior.

pkg/cloudprovider/provider/kubevirt/types/types.go
```diff
@@ -67,6 +67,25 @@ type Template struct {
...
+
+// Source describes thr VM Disk Image source.
+type Source struct {
+	// HTTP represents http source for downloading VMDisk Image.
+	HTTP *HTTPSource `json:"http,omitempty"`
+	// PVCClone represents the pvc source for cloning of VMDisk Image.
+	PVCClone *PVCCloneSource `json:"pvcClone,omitempty"`
+}
+
+// HTTPSource.
```
mfranczy, line 81: > We need better comments with explanation about the fields.
The same for all sources (HTTP and PVCClone)

pkg/cloudprovider/provider/kubevirt/provider.go
mfranczy: so if MD-name returns an error then we have a problem. Based on that you create a data volume.

mfranczy:
I am sorry @sankalp-r I have to block this PR, we have to do another brainstorming session

/hold

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -148,9 +153,28 @@ type SecondaryDisks struct {
...
+	// URL of the http source endpoint.
+	URL string
+	// CacheImage represents whether to use cachedImage or not.
+	CacheImage bool
+}
+
+// PVCSource represents pvc source for cloning VMImage.
+type PVCSource struct {
+	// DataVolumeName represent name of source pvc for cloning.
 	DataVolumeName string
```
mfranczy, line 175: let's change `DataVolumeName` to `Name`

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -66,6 +66,10 @@ const (
 	// machineDeploymentLabelKey defines the label key used to contains as value the MachineDeployment name
 	// which machine comes from.
 	machineDeploymentLabelKey = "md"
+	// httpSource defines the http source type for VM Disk Image.
+	httpSource = "http"
```
mfranczy, line 70: I would suggest to introduce `imageSource` type.
```
type imageSource string

const (
   ...
   httpSource imageSource = "http"
   pvcSource imageSource = "pvc"
)
````

and later in the switch statement you would have

````
switch imageSource(osImageSource) ...
````

File: pkg/cloudprovider/provider/kubevirt/provider.go [hunk 21]
```diff
@@ -377,6 +373,35 @@ func (p *provider) parseTopologySpreadConstraint(topologyConstraints []kubevirtt
...
+	}
+	osImageSource, err := p.configVarResolver.GetConfigVarStringValue(primaryDisk.Source)
+	if err != nil {
+		return nil, fmt.Errorf(`failed to get value of "primaryDisk.source" field: %w`, err)
+	}
+	switch osImageSource {
+	case httpSource:
+		return &cdiv1beta1.DataVolumeSource{HTTP: &cdiv1beta1.DataVolumeSourceHTTP{URL: osImage}}, nil
+	case pvcSource:
+		if nameSpaceAndName := strings.Split(osImage, "/"); len(nameSpaceAndName) >= 2 {
```
mfranczy, line 389: please name it `namespaceAndName` (namespace is one word)

mfranczy (approved):
/unhold
/approve
/lgtm

pkg/cloudprovider/provider/kubevirt/provider_test.go
```diff
@@ -167,7 +168,12 @@ func (k kubevirtProviderSpecConf) rawProviderSpec(t *testing.T) []byte {
 					"osImage": "http://x.y.z.t/ubuntu.img",
 					{{- end }}
 					"size": "10Gi",
+					{{- if .OsImageSource }}
+					"storageClassName": "longhorn",
```
hdurand0710, line 172: As `storageClassName` is set to "longhorn" in the if and the else, it could go out of the if/else condition.

mfranczy: there is https://github.com/kubermatic/machine-controller/pull/1470/files#diff-2e539081e4b1bd387e1af416faf12022e219faba4f706b478b0f25523315a7a4R175

it could be just one if to add `OsImageSource`  true, but the  current state doesn't break the behaviour.

PR #1472, opened 2022-11-02
Add cluster labels to kubevirt vm

dermorz:
For https://github.com/kubermatic/kubermatic/issues/10923 we need to add cluster labels to KubeVirt VMs on creation.

```release-note
Add cluster labels to KubeVirt VM
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -608,6 +608,10 @@ func (p *provider) newVirtualMachine(ctx context.Context, c *Config, pc *provide
 		resourceRequirements.Limits = *requestsAndLimits
 	}
 
+	// Add cluster labels
+	labels["cluster.x-k8s.io/cluster-name"] = c.Namespace
+	labels["cluster.x-k8s.io/role"] = "worker"
+
 	var (
 		dataVolumeName = machine.Name
 		annotations    map[string]string
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -93,6 +93,7 @@ func New(configVarResolver *providerconfig.ConfigVarResolver) cloudprovidertypes
 
 type Config struct {
 	Kubeconfig                string
+	ClusterName               string
 	RestConfig                *rest.Config
 	DNSConfig                 *corev1.PodDNSConfig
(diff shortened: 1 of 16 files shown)


pkg/cloudprovider/provider/kubevirt/provider_test.go
```diff
@@ -192,7 +192,7 @@ func (k kubevirtProviderSpecConf) rawProviderSpec(t *testing.T) []byte {
 
 var (
 	userdata      = "fake-userdata"
-	testNamespace = "test-namespace"
+	testNamespace = "tenant-cluster"
```
hdurand0710, line 195: I find this name somehow confusing.
Maybe more `cluster-xyz` that would be more close to the real names ?

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -599,6 +599,10 @@ func (p *provider) newVirtualMachine(ctx context.Context, c *Config, pc *provide
 		}
 	}
 
+	// Add cluster labels
+	labels["cluster.x-k8s.io/cluster-name"] = c.Namespace
```
hdurand0710, line 603: IIRC, kubevirt ccm 0.58.0 adds the following selector to the infra svc:
```
selector:
  cluster.x-k8s.io/cluster-name: kubernetes
  cluster.x-k8s.io/role: worker
```

cluster-name is the cluster name, it's not `cluster-xyz` (that is the namespace name).
I think with this change it would not work.
Except if something was changed on KKP side to change the cluster name ? Sorry maybe I missed it last week, I was off.

hdurand0710: We first need to change in KKP 2 things before this PR is machine-controller can be merged:
- https://github.com/kubermatic/kubermatic/blob/main/pkg/resources/cloudcontroller.go#L109: add KubeVirt in the list of external CCM
- https://github.com/kubermatic/kubermatic/blob/main/pkg/resources/cloudcontroller/kubevirt.go#L134: the `--cluster-name` flag is set here. All other providers set `data.Cluster().Name`, do we really want to set a specific value for KubeVirt `data.Cluster(). NamespaceName`?

hdurand0710:
Can you please also update `examples/kubevirt-machinedeployment.yaml` with the new parameter ?

hdurand0710:
/unhold 
KKP is merged

PR #1473, opened 2022-11-02
Fix static network rendering for flatcar

### LINKED ISSUE #1468: [vSphere] Impossible to deploy machines without static network
With #1263 by @PratikDeoghare the method to check if a provider spec configures static networking was changed from checking if `.Network` is `nil` to a new `IsStaticIPConfig` method. This was necessary because the Spec now includes the `ipFamily` field which is set regardless of whether the network configuration is static or not.

The replacement was done for all userdata providers except `flatcar` which still has the old check in it's config templates
https://github.com/kubermatic/machine-controller/blob/debfca7b52db265625167e1285c9fa7879c4f2d6/pkg/userdata/flatcar/provider.go#L173

Now, when new Machine Deployments are created with the new `ipFamily` setting set (which KKP does by default), the resulting machines all receive a static (but empty) network configuration. This effectively renders them unusable.

ahmedwaleedmalik:

Fixes a bug where static networking configuration was being wrongfully set for flatcar.

Fixes #1468

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/userdata/flatcar/provider.go
+++ b/pkg/userdata/flatcar/provider.go
@@ -170,7 +170,7 @@ const userDataIgnitionTemplate = `passwd:
         {{end}}
 {{- end }}
 
-{{- if .ProviderSpec.Network }}
+{{- if .ProviderSpec.Network.IsStaticIPConfig }}
 networkd:
   units:
     - name: static-nic.network
@@ -544,7 +544,7 @@ users:
 
 coreos:
   units:
-{{- if .ProviderSpec.Network }}
+{{- if .ProviderSpec.Network.IsStaticIPConfig }}
   - name: static-nic.network
     content: |
       [Match]


pkg/userdata/flatcar/provider.go
```diff
@@ -170,7 +172,7 @@ const userDataIgnitionTemplate = `passwd:
         {{end}}
 {{- end }}
 
-{{- if .ProviderSpec.Network }}
+{{- if .StaticIPConfigProvided }}
```
PratikDeoghare, line 175: You could just `.ProviderSpec.Network.IsStaticIPConfig`

like in https://stackoverflow.com/a/10200273

ahmedwaleedmalik: Forgot about this completely 😁

PratikDeoghare (approved):

PR #1476, opened 2022-11-04
restart containerd for flatcar so the drop in conf will be considered

lucakuendig:
Signed-off-by: Luca Kuendig <luca.kuendig@bedag.ch>

I noticed, that during the first boot of a Flatcar node custom configurations, like containerd custom registry endpoint and auth, will not be used. I think it is because compared to other operatingsystems Flatcar allready has containerd installed and starts it directly after boot, so systemctl enable --now does nothing. By restarting the containerd service instead of systemctl enable (should be enabled by default) The config will be picked up on the first boot.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik:
@lucakuendig the fixture data needs to be updated. Please run the script `hack/update-fixtures.sh`.

ahmedwaleedmalik (approved):
/approve

Thank you for taking care of this. 🚀

PR #1489, opened 2022-11-18
vSphere: only delete tags created by machine-controller

### LINKED ISSUE #1433: [vSphere] Machine Cleanup fails due to inability to remove tags
Since #1213 was merged our `machine-controller` deployment is unable finish machine cleanup and keeps unused and drained machines running indefinitely.

The machine cleanup fails with `failed to delete machine at cloud provider, due to failed to delete tags`. Indeed machine-controller tries to delete some tags that it is not supposed to touch (because they are created and used by our external monitoring system) and fails because it lacks the permissions to remove them.

The assumption seems to be that all tags on the machine were created by machine-controller which I think might not be true in most vSphere deployments since a lot of monitoring/backup/compliance systems are using tags to keep track of vSphere objects.

To fix this this I think this code here should probably check which of the attached tags are actually in the configuration object and only delete those with matching names in the `config` struct. Instead of just trying to delete all tags attached to the machine. https://github.com/kubermatic/machine-controller/blob/19af243aeb7fbacfb2706c5d5b63b59de9796f50/pkg/cloudprovider/provider/vsphere/helper.go#L499-L509

heilerich:
In vSphere machine tags are distinct objects. When machine-controller tags a machine it creates a new tag object and attaches it to the machine. When the machine is deleted, machine-controller deletes all the tag objects attached to the machine in order to not leave any orphaned tags behind.

The current assumption is that all tag objects on the machine are created by machine-controller and only attached to the machine that is being removed. This is probably not true in most vSphere deployments as many external systems interfacing with vSphere such as monitoring and backup tools attach tags to machines. The deletion of these external tag objects can lead to two outcomes:

1. machine-controller does not have permission to delete the tag and will crash. Nodes will not be deleted and remain in an unschedulable state indefinitely.
2. machine-controller deletes the tag object which will remove it from *all* machines it was attached to potentially wreaking havoc on a backup system or similar.

Fixes #1433

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/helper.go
+++ b/pkg/cloudprovider/provider/vsphere/helper.go
@@ -487,6 +487,10 @@ func createAndAttachTags(ctx context.Context, config *Config, vm *object.Virtual
 	return nil
 }
 
+type distinctTag struct {
+	name, categoryID string
+}
+
 func deleteTags(ctx context.Context, config *Config, vm *object.VirtualMachine) error {
 	restAPISession, err := NewRESTSession(ctx, config)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik:
@heilerich with https://github.com/kubermatic/machine-controller/pull/1532, we refactored the way tagging works for vSphere. IMO we don't need this PR anymore as this problem is fixed now.

@moadqassem WDYT?

heilerich:
@ahmedwaleedmalik Thank for your effort in resolving this issue. One question: If I understand your changes correctly, machine-controller will not handle the creation or deletion of tags at all anymore. Is there a replacement for that functionality (yet)? That is, will e.g. KKP handle the creation and removal of orphaned vSphere Tag objects?

If any, could you point me to the relevant changes?

Anyway, I agree that we probably don't need this PR anymore so I will close this for now.

PR #1483, opened 2022-11-15
Anexia Provider: Utilize `Creating` state instead of blocking `Create` call

marioreggiori:
**What this PR does / why we need it**:

The `machine-controller` calls the providers `Create` method sequentially and only starts with the provisioning of another instance after `Create` returned without any errors. This PR changes the

- `Create` method to not block until the VM provisioning has completed 
- `Get` method to check via the instance status whether `InstanceID` and `ProvisioningID` are set

and utilizes the `Creating` state until the VM is ready.

This reduces the provisioning duration of node pools with more than one replica significantly as there are now only a couple of seconds (~10s) between the provisioning starts.

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
Anexia: Refactor code to reduce provisioning time for node pools with multiple replicas.
```

**Documentation**:

```documentation
NONE
```

PR #1509, opened 2022-12-07
[release/v1.54] Anexia Provider: Utilize `Creating` state instead of blocking `Create…

LittleFox94:

Manual cherry-pick of #1483 

**What type of PR is this?**
/kind cleanup

```release-note
Anexia: Refactor code to reduce provisioning time for node pools with multiple replicas.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/anexia/helper_test.go
+++ b/pkg/cloudprovider/provider/anexia/helper_test.go
@@ -18,11 +18,9 @@ package anexia
 
 import (
 	"encoding/json"
-	"net/http"
 	"testing"
 
 	"github.com/gophercloud/gophercloud/testhelper"
-	"go.anx.io/go-anxcloud/pkg/vsphere/search"
 
 	"github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
 	anxtypes "github.com/kubermatic/machine-controller/pkg/cloudprovider/provider/anexia/types"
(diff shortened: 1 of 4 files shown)


LittleFox94:
Since @ahmedwaleedmalik reviewed the original PR, it would probably make sense if you take a look at this one as well - thanks :)

ahmedwaleedmalik (approved):
/approve

LittleFox94:
this should probably also go into `release/v1.55`, what do you think?

ahmedwaleedmalik:
> this should probably also go into `release/v1.55`, what do you think?

In general 1.55 is only used in KubeOne at the moment. I have added the backport though.

LittleFox94:
> > this should probably also go into `release/v1.55`, what do you think?
> 
> In general 1.55 is only used in KubeOne at the moment. I have added the backport though.

just thought it might be weird to have a change in main and a release older than the latest one, but not in the latest release - thanks :)

LittleFox94:
Hi @ahmedwaleedmalik,
since we still use KKP 2.20, we'd need this backported to v1.45 as well - could you cherry-pick this or should we create the PR manually?

ahmedwaleedmalik:
> Hi @ahmedwaleedmalik, since we still use KKP 2.20, we'd need this backported to v1.45 as well - could you cherry-pick this or should we create the PR manually?

I'll take care of that.

PR #1518, opened 2022-12-19
Allow manually configuring containerd version via CLI flag

dajudge:
A [broken containerd release](https://github.com/containerd/containerd/pull/7838) made it impossible for our setup to provision working new nodes. This was a result of `machine-controller` automatically installing the latest `1.6*` release of `containerd` available.

This PR introduces the command line flag `--node-containerd-version` to manually pin the `containerd` version to install to a specific release.

The default for this new flag is empty, so the previously existing behavior remains intact if the flag is not specified.

Implements partially https://github.com/kubermatic/kubeone/issues/2545

**What type of PR is this?**
/kind feature

```release-note
Introduce `--node-conterainerd-version` CLI flag.
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/containerruntime/containerruntime.go
+++ b/pkg/containerruntime/containerruntime.go
@@ -57,6 +57,12 @@ func withSandboxImage(image string) Opt {
 	}
 }
 
+func withContainerdVersion(version string) Opt {
+	return func(cfg *Config) {
+		cfg.ContainerdVersion = version
+	}
+}
+
 func get(containerRuntimeName string, opts ...Opt) Config {
 	cfg := Config{}
 
@@ -88,6 +94,7 @@ type Config struct {
 	SandboxImage         string                `json:",omitempty"`
 	ContainerLogMaxFiles string                `json:",omitempty"`
 	ContainerLogMaxSize  string                `json:",omitempty"`
+	ContainerdVersion    string                `json:",omitempty"`
 }
 
 // AuthConfig is a COPY of github.com/containerd/containerd/pkg/cri/config.AuthConfig.
@@ -130,6 +137,7 @@ func (cfg Config) Engine(kubeletVersion *semver.Version) Engine {
 		registryMirrors:     cfg.RegistryMirrors,
 		sandboxImage:        cfg.SandboxImage,
 		registryCredentials: cfg.RegistryCredentials,
+		version:             cfg.ContainerdVersion,
 	}
 
 	moreThan124, _ := semver.NewConstraint(">= 1.24")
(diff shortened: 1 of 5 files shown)


pkg/containerruntime/containerruntime.go
```diff
@@ -130,6 +137,7 @@ func (cfg Config) Engine(kubeletVersion *semver.Version) Engine {
 		registryMirrors:     cfg.RegistryMirrors,
 		sandboxImage:        cfg.SandboxImage,
 		registryCredentials: cfg.RegistryCredentials,
+		version:             cfg.ContainerdVersion,
```
xmudrii, line 140: I don't see that we use this version anywhere. I think that we might need to add it in two places:

https://github.com/kubermatic/machine-controller/blob/e7e4ba8eb535086374108ed235c2e33bab7f4292/pkg/containerruntime/containerd.go#L67

https://github.com/kubermatic/machine-controller/blob/e7e4ba8eb535086374108ed235c2e33bab7f4292/pkg/containerruntime/docker.go#L88

xmudrii (commented):
Thanks a lot for this contribution, we really appreciate it! :tada:

I have one question, but other than that, LGTM.

pkg/containerruntime/containerruntime.go
dajudge: https://github.com/kubermatic/machine-controller/blob/e7e4ba8eb535086374108ed235c2e33bab7f4292/pkg/containerruntime/containerd.go#L67 gets overwritten by the value passed via CLI (if present) at https://github.com/kubermatic/machine-controller/blob/e7e4ba8eb535086374108ed235c2e33bab7f4292/pkg/containerruntime/containerd.go#L71

As for the docker runtime: I added the same mechanism already in place for containerd.

WDYT?

dajudge:
I also noticed that the containerd version gets overwritten to a legacy version in the case of AmazonLinux at the following places:

https://github.com/AtomicJar/machine-controller/blob/5d54195d47e9e78104f330bc877030b5cefd9740/pkg/containerruntime/docker.go#L98

https://github.com/AtomicJar/machine-controller/blob/5d54195d47e9e78104f330bc877030b5cefd9740/pkg/containerruntime/containerd.go#L77

Maybe it would make sense to move those `if` blocks above the `switch` to be below the latter so the value provided via CLI overrides this fallback? :thinking:

pkg/containerruntime/containerruntime.go
xmudrii: This looks good now!

xmudrii:
> Maybe it would make sense to move those if blocks above the switch to be below the latter so the value provided via CLI overrides this fallback?

I don't have strong opinions. Amazon Linux is a bit specific -- they ship only one specific version of containerd (and other relevant packages). I think, for consistency reasons, it might make sense to do so.

In the meanwhile, I'll trigger tests for this PR.
/test all

dajudge:
> Maybe it would make sense to move those if blocks above the switch to be below the latter so the value provided via CLI overrides this fallback?

To unblock this PR I've put above change on a [separate branch](https://github.com/AtomicJar/machine-controller/tree/cli-over-al2-fallback).

Once this is PR here is on `main` I'd simply open a new PR and we can discuss it in detail there.

WDYT?

xmudrii (approved):
/lgtm
/approve

PR #1528, opened 2023-01-11
KubeVirt export few names as public

hdurand0710:

Make some variable names public, so they will not be re-declared/duplicated in KKP or elsewhere.

Fixes #

**What type of PR is this?**

/kind cleanup

```release-note
KubeVirt make some variable names public for standard and custom images.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -75,10 +75,19 @@ const (
 	httpSource imageSource = "http"
 	// pvcSource defines the pvc source type for VM Disk Image.
 	pvcSource imageSource = "pvc"
-	// kubeVirtImagesNamespace namespace contains globally available custom images and cached standard images.
-	kubeVirtImagesNamespace           = "kubevirt-images"
-	dataVolumeStandardImageAnnotation = "kubevirt-initialization.k8c.io/standard-image"
-	osAnnotationForCustomDisk         = "cdi.kubevirt.io/os-type"
+	// DataVolumeStandardImageAnnotationKey is the annotation key for standard DataVolumes.
+	DataVolumeStandardImageAnnotationKey = "kubevirt-initialization.k8c.io/standard-image"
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -75,10 +75,19 @@ const (
...
-	kubeVirtImagesNamespace           = "kubevirt-images"
-	dataVolumeStandardImageAnnotation = "kubevirt-initialization.k8c.io/standard-image"
-	osAnnotationForCustomDisk         = "cdi.kubevirt.io/os-type"
+	// DataVolumeStandardImageAnnotationKey is the annotation key for standard DataVolumes.
+	DataVolumeStandardImageAnnotationKey = "kubevirt-initialization.k8c.io/standard-image"
+	// DataVolumeStandardImageAnnotationKey is the annotation value for standard DataVolumes.
+	DataVolumeStandardImageAnnotationValue = "true"
+	// DataVolumeCustomDiskOsAnnotationKey is the annotation key for custom DataVolumes.
+	// Annotation value will be the os type.
+	DataVolumeCustomDiskOsAnnotationKey = "cdi.kubevirt.io/os-type"
```
mfranczy, line 84: for sure we have to change this annotation to something different that's not under cdi.
Let's discuss that tomorrow.

hdurand0710: glad to read this. I thought this was agreed with you and was surprised.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -75,10 +75,19 @@ const (
...
+	// KubeVirtImagesNamespace namespace contains globally available custom images and cached standard images.
+	KubeVirtImagesNamespace = "kubevirt-images"
```
mfranczy, line 86: I think we will not need that in machine controller as we pass  namespace/name if source of image is pvc.

PR #1532, opened 2023-01-13
Refactor vsphere tagging mechanism

moadqassem:
When a new tag is being added to a vsphere machine deployment, a new tag is being created(if it didn't exist) and assign it to the machine. Once the machine is cleaned up, machine controller would simply delete that tag. Such a scenario is implemented in different cloud provider. However, vSphere tags are kinda different from a structure perspective, as tags must first belong to a tag category and then they can be assigned. In addition to that, tags can from different categories can be shared across different inventory resources. 

With that being said, if a machine being cleaned up and one tag is assigned to another VM, this tag will be lost. Also, if tags are created manually be users and assign across different resources, machine controller should never delete any resources which is not managed by it. 

This PR changes this behaviour to either attach tags on VMs or detach them in case of machine cleanups instead of deleting or creating those tags. 
  
Fixes #

**What type of PR is this?**
/kind bug
/kind api-change
/kind feature

```release-note
Stop creating and deleting vSphere tags unconditionally in machine controller.
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/helper.go
+++ b/pkg/cloudprovider/provider/vsphere/helper.go
@@ -456,38 +456,24 @@ func resolveResourcePoolRef(ctx context.Context, config *Config, session *Sessio
 	return nil, nil
 }
 
-func createAndAttachTags(ctx context.Context, config *Config, vm *object.VirtualMachine) error {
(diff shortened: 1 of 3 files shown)


ahmedwaleedmalik (approved):
/approve 🚀

PR #1561, opened 2023-02-15
Always use latest OSM image in e2e tests

ahmedwaleedmalik:
The E2E tests that we run in machine-controller should always be in sync with the latest state in OSM. This helps us avoid manually updating the image tag for testing OSM changes.

We also don't need to run tests on Azure unconditionally.

Fixes #

**What type of PR is this?**

/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/.prow/provider-azure.yaml
+++ b/.prow/provider-azure.yaml
@@ -14,7 +14,7 @@
 
 presubmits:
   - name: pull-machine-controller-e2e-azure
-    always_run: true
+    run_if_changed: "(pkg/cloudprovider/provider/azure/|pkg/userdata)"
     decorate: true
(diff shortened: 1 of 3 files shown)


WeirdMachine:
I am not sure about this one as i do not know about the dependencies for osm to machine-controller. Don't you need to test a specific release as well?

ahmedwaleedmalik:
> I am not sure about this one as i do not know about the dependencies for osm to machine-controller. Don't you need to test a specific release as well?

This change is just for the main branch. We will use tags for all the release branches. Right now the main issue is that we are using commit hashes instead of tags in the main branch which is a bit problametic.

PR #1565, opened 2023-02-17
Use AWS spot instances for tests that always run

ahmedwaleedmalik:
We run tests on AWS unconditionally, for E2E/sanity testing, which covers legacy userdata and OSM userdata testing. We should rely on spot instances for these tests for cost-cutting purposes.

A small downside of this is that we'll have to make adjustments in the code if we can't match the spot pricing from AWS. But imo it's still worth it.

Fixes #

**What type of PR is this?**

/kind chore

The code changes are the result of running `gimps` on this repo. I just figured out that we don't have a job to verify imports in this repo. 

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/.prow/provider-aws.yaml
+++ b/.prow/provider-aws.yaml
@@ -14,7 +14,7 @@
 
 presubmits:
   - name: pull-machine-controller-e2e-aws
-    always_run: true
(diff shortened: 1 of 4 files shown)


mfranczy (approved):
/approve
/lgtm

PR #1569, opened 2023-02-21
Skip validation for MachineDeployments and Machines that are marked for deletion

ahmedwaleedmalik:
It's not wise to apply validation webhooks when a resource is marked for deletion. Any misconfiguration might mean that the resource is stuck in deletion forever. 

Fixes #

**What type of PR is this?**

/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


kron4eg:
@ahmedwaleedmalik how to proceed in case when, for example, a Machine is being mutated in a way that loose information prevents its clean deletion (e.g. missing auth information, or other IDs)?

PR #1570, opened 2023-02-21
Increase default timeout for OpenStack provider

ahmedwaleedmalik:
Currently, the timeout for the OpenStack client defaults to 15 seconds which is quite low. Self-hosted OpenStack environments might take more time than usual based on the networking configuration. Since we always authenticate before performing actions it's fine to increase the default timeout a bit.

Fixes #

**What type of PR is this?**

/kind chore

```release-note
Default timeout for OpenStack client has been increased to 300 seconds.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -57,6 +57,7 @@ import (
 const (
 	floatingIPReleaseFinalizer = "kubermatic.io/release-openstack-floating-ip"
 	floatingIPIDAnnotationKey  = "kubermatic.io/release-openstack-floating-ip"
+	clientTimeout              = 1 * time.Minute
 )
 
 // clientGetterFunc returns an OpenStack client.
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -57,6 +57,7 @@ import (
 const (
 	floatingIPReleaseFinalizer = "kubermatic.io/release-openstack-floating-ip"
 	floatingIPIDAnnotationKey  = "kubermatic.io/release-openstack-floating-ip"
+	clientTimeout              = 3 * time.Minute
```
kron4eg, line 60: Probably we should make it configurable.

ahmedwaleedmalik: I wanted that as well but I'm not sure if I'd like to configure such a thing on the MD level :/ Maybe as a flag to machine-controller?!

WeirdMachine: Personally I think 3 mins are also to much, personally I would prefer something around 1 minute with is already an increase by 4x.

WeirdMachine: If we make it configurable this should also apply to all cloud provider api calls which is IMO a bigger change.

ahmedwaleedmalik: 1 minute should be more than enough, now that I think of it.

ahmedwaleedmalik: I've switched this to 1 minute which should be enough. Having it configurable although would be good but it requires too much work since we'd then want something that is also configurable at KKP or KubeOne level. For now, I think setting it to 1 minute should be enough because imo 15 seconds is on the lower side when it comes to OpenStack.

PR #1590, opened 2023-03-13
Validate the AWS VPC has DNS hostnames enabled

xrstf:
If the `enableDnsHostnames` attribute is not enabled on a VPC, new nodes will register themselves using their private IP (e.g. `ip-10-0-2-14`), even though the EC2 instance uses the correct name (e.g. it's named after the Machine object).

As the AWS CCM (in-tree and out-of-tree) tries to find the node based on its name, it will always fail, as there is now EC2 instance with the name "ip-......".

Figuring this out took me 3 hours alone and then 30 seconds with @xmudrii's help. We then thought it might make sense to validate this early and completely refuse to work in incompatible VPCs, hence this PR adds a bit of validation.

**What type of PR is this?**
/kind feature

```release-note
AWS: Machines using a VPC without `enableDnsHostname` enabled will now fail validation, as those machines would not be able to register themselves successfully as nodes in the cluster.
```

**Documentation**:
```documentation
NONE
```


xmudrii (approved):
/lgtm
/approve
/hold
for discussion

ahmedwaleedmalik (commented):
/lgtm

kron4eg:
FYI: https://cloud-provider-aws.sigs.k8s.io/prerequisites/#node-name-conventions

> When IP-based naming is used, the nodes must be named after the instance followed by the regional domain name (ip-xxx-xxx-xxx-xxx.ec2.<region>.internal)

xmudrii (approved):

PR #1591, opened 2023-03-13
adjust VPC to the new one after we recently switched AWS accounts

xrstf:
We recently changed the AWS account in which E2E tests are executed, so we need to adjust the VPC ID as well.

The documentation for the instance profile was wrong, since #412 we do not create it automatically anymore. So I had to manually create it by

* creating an IAM role as shown in the deleted code of  #412
* add the two AWS managed policies to that role
* then, because I could not figure out how to do it via the web console, create the instance profile and attach the role to it:
    ```
    aws iam create-instance-profile --instance-profile-name kubernetes-v1
    aws iam add-role-to-instance-profile --instance-profile-name kubernetes-v1 --role-name kubernetes-v1
    ```

**What type of PR is this?**
/kind failing-test

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/docs/cloud-provider.md
+++ b/docs/cloud-provider.md
@@ -80,8 +80,7 @@ ami: ""
 # When not set a 'kubernetes-v1' security group will get created
 securityGroupIDs:
 - ""
-# name of the instance profile to use.
-# When not set a 'kubernetes-v1' instance profile will get created
+# name of the instance profile to use, required.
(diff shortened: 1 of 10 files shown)


embik (approved):

ahmedwaleedmalik:
@xrstf Next time, please make sure to backport such PRs as well.

PR #1596, opened 2023-03-14
initialize iscsi if cloudprovider is Nutanix

lucakuendig:
In order to be able to use the Nutanix CSI driver you need to create an InitiatorName and start/enable the iscsid.service. This PR will autmatically do both steps for Flatcar OS

**What type of PR is this?**
/kind feature

```release-note
This Change will create the file /etc/iscsi/initiatorname.iscsi, if not present, start and enable the iscsid.service for any new VM, if you use flatcar nodes on Nutanix.
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik:
I couldn't take a look on Friday because we were hit with a CI issue, GitHub key rotation. Let me see what's wrong here. Doesn't seem like something from your end.

xrstf:
/override pull-machine-controller-e2e-vmware-cloud-director

temporarily unavailable

ahmedwaleedmalik (approved):
/approve

PR #1598, opened 2023-03-23
WIP - debug webhooks

xrstf:

Fixes #

**What type of PR is this?**

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


xrstf:
Cause found: Relying on OSM:latest accidentally caused us to use 1.1.3, because 1.1.3's prowjob was run a few seconds after 1.2.1 and so `latest` was not really pointing to the most recent release anymore.

Re-ran the OSM release job, so now latest is good again.

Nothing more to do here.

PR #1606, opened 2023-03-27
Use zap instead of klog

xrstf:
This PR replaces klog with Zap. Why?

* We (as in Kubermatic GmbH) use Zap pretty much everywhere.
* Zap support exists in controller-runtime as well.
* klog is terribly useless for structured/contextual logging.
* I personally just immensely dislike klog. :-D

There were a few log messages on level 3, which is somewhere between INFO and DEBUG. I made a gut choice for all of those, trying to make the logs useful but not too verbose.

klog still remains for the debug HTTP client (as it depends on numeric log levels, something we cannot do with Zap) and for the e2e tests (their logs are kind of boring and don't need Zap that much).

**What type of PR is this?**
/kind feature

```release-note
Logging was replaced with Zap; the klog-related CLI flags like `-logtostderr`, `-v` etc. have been removed; use `-log-debug=true/false` and `-log-format=text/json` instead.
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik:
/hold 
to avoid retries

PR #1607, opened 2023-03-27
Remove custom AMI from rhel os

moadqassem:
Since we have enabled the GA images for RHEL, in addition of having a default RHEL image added as part of the cloud provider controller, we don't need to include a custom iam for RHEL machines. 

Fixes #

**What type of PR is this?**
/kind cleanup

```release-note
None
```

**Documentation**:

```documentation
None
```


ahmedwaleedmalik (approved):
/approve 🚀

Issue #1082, opened 2021-10-28
Google Cloud Platform cloudProviderSpec should also pass the GCP Project ID context


JoaquimFreitas:
If the `cloudProviderSpec ` for the Google Cloud Platform (Provider: `gce`) also pass the GCP Project ID context to be used by the `machine-controller`  - _the same Project ID used in the Terraform Google Provider configuration, that can be used in the output for the Terraform integration of KubeOne_ -  that will allow the use of higher-level Service Accounts on the Google Cloud Platform _(service accounts that have the necessary roles and access to several projects on GCP, this kind of management is allowed on Google Cloud)_, instead of using a specific  service account scoped to a specific project.

Currently, if a higher-level SA is used at GCP, it will work without problems with Terraform, it will be used in the Terraform Integration with KubeOne, but when the final deployment of the `machine-controller` is done and is trying to reconcile the workers pool with the initial instance (node) it will fail with errors like:

`An error of type = InvalidConfiguration, with message = Failed to insert instance: googleapi: Error 400: Invalid value for field 'resource.networkInterfaces[0]': '{ "network": "https://www.googleapis.com/compute/v1/projects/someprojectv/global/networks/somenetwork"...'. Cross-project references for this resource are not allowed., invalid occurred`

PR #1613, opened 2023-04-06
Refactor gcp client creating

moadqassem:
Currently, it is not possible to create gce instances in a different gcp project than the one that is specified in the service token, as we derive the project id from it. This PR changes this behaviour so users can share a service account across different projects and explicitly specify in which project the machine should be created.

Fixes https://github.com/kubermatic/machine-controller/issues/1082

**What type of PR is this?**
/kind feature
/kind api-change

```release-note
None
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/pkg/cloudprovider/provider/gce/config.go
+++ b/pkg/cloudprovider/provider/gce/config.go
@@ -21,14 +21,15 @@ limitations under the License.
 package gce
 
 import (
+	"context"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"strings"
 
-	"golang.org/x/oauth2/google"
-	"golang.org/x/oauth2/jwt"
+	"golang.org/x/oauth2"
+	googleoauth "golang.org/x/oauth2/google"
 	"google.golang.org/api/compute/v1"
 
 	"github.com/kubermatic/machine-controller/pkg/apis/cluster/v1alpha1"
@@ -106,7 +107,6 @@ type config struct {
 	provisioningModel            *string
 	labels                       map[string]string
 	tags                         []string
-	jwtConfig                    *jwt.Config
 	providerConfig               *providerconfigtypes.Config
 	assignPublicIPAddress        bool
 	multizone                    bool
@@ -116,6 +116,12 @@ type config struct {
 	enableNestedVirtualization   bool
 	minCPUPlatform               string
 	guestOSFeatures              []string
+	clientConfig                 *clientConfig
+}
+
+type clientConfig struct {
+	ClientEmail string
+	TokenSource oauth2.TokenSource
 }
 
 // newConfig creates a Provider configuration out of the passed resolver and spec.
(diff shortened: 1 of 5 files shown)


pkg/cloudprovider/provider/gce/types/types.go
```diff
@@ -51,6 +51,7 @@ type CloudProviderSpec struct {
 	EnableNestedVirtualization   providerconfigtypes.ConfigVarBool    `json:"enableNestedVirtualization,omitempty"`
 	MinCPUPlatform               providerconfigtypes.ConfigVarString  `json:"minCPUPlatform,omitempty"`
 	GuestOSFeatures              []string                             `json:"guestOSFeatures,omitempty"`
+	ProjectID                    providerconfigtypes.ConfigVarString  `json:"projectID,omitempty"`
```
xmudrii, line 54: We should add this field to the example manifest: https://github.com/kubermatic/machine-controller/blob/main/examples/gce-machinedeployment.yaml

pkg/cloudprovider/provider/gce/service.go
```diff
@@ -54,7 +55,7 @@ type service struct {
 
 // connectComputeService establishes a service connection to the Compute Engine.
 func connectComputeService(cfg *config) (*service, error) {
-	client := cfg.jwtConfig.Client(context.Background())
+	client := oauth2.NewClient(context.Background(), cfg.clientConfig.TokenSource)
```
xmudrii, line 58: Nit: technically, it should always exist, but `cfg.clientConfig` is a pointer and it can NPE if it's not set. Maybe it's worth checking for that before using `TokenSource`.

pkg/cloudprovider/provider/gce/config.go
```diff
@@ -251,16 +262,28 @@ func (cfg *config) postprocessServiceAccount() error {
 		sa = string(decoded)
 	}
 
+	creds, err := googleoauth.CredentialsFromJSON(context.TODO(), []byte(sa), compute.ComputeScope)
```
xmudrii, line 265: Nit: it might be better to use `context.Background()` here.

moadqassem: Why? This transaction is actually propagated by another caller which is the cloud providers methods. `context.Background()` which is used at main functionality(e.g: machine creating), however, `context.TODO()` is used when the main context isn't propagated to the called functions ;)

pkg/cloudprovider/provider/gce/service.go
moadqassem: As you mentioned, pointer is already initialised before any use of this field thus what would the check help with?

pkg/cloudprovider/provider/gce/config.go
xmudrii: `context.TODO()` is a kinda "marker" to get back to this and put a real context (e.g. pass context down the stack or use `context.Background()`). If we don't plan on refactoring this later to provide already existing context, using `context.Background()` is a semantically better approach. However, I'm not going to block on this, I don't have strong opinions.

pkg/cloudprovider/provider/gce/service.go
xmudrii: The idea is if we eventually call this function before initializing the struct or we introduce some bug, we can log it instead of NPE. No strong opinions here, so we can ignore this.

pkg/cloudprovider/provider/gce/config.go
moadqassem: Well that's the thing, I wanted to come back again and propagate the context through all other internal functions that needed it ;)

pkg/cloudprovider/provider/gce/types/types.go
moadqassem: Done

moadqassem:
@xmudrii  comments addressed, PTAL

xmudrii (approved):

PR #1526, opened 2023-01-10
exact string match fails, use conventional errors.Is instead

furkhat:
**What this PR does / why we need it**:

Fixes a bug when machine-controller fails on 404 when deleting already gone floating-ip or server. 
Example failure
```
failed to delete machine at cloud provider, due to failed to clean up floating ip: failed to delete floating ip a87484bd-4ba6-4c18-8b7e-27b98606d240: Resource not found: [DELETE https://example-cloud.net:9696/v2.0/floatingips/a87484bd-4ba6-4c18-8b7e-27b98606d240], error message: {"NeutronError": {"message": "Floating IP a87484bd-4ba6-4c18-8b7e-27b98606d240 could not be found", "type": "FloatingIPNotFound", "detail": ""}}
```

Exact string match against `"Resource not found"` does not work because gophercloud v1.1.1, as well as previous versions, [return](https://github.com/gophercloud/gophercloud/blob/v1.1.1/errors.go#L193) 
```golang
 fmt.Sprintf(
        "Resource not found: [%s %s], error message: %s",
	e.Method, e.URL, e.Body,
)
``` 
**Which issue(s) this PR fixes**:

Fixes #

**What type of PR is this?**
/kind bug

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
Fix Openstack IP cleanup
```

**Documentation**:

```documentation
NONE
```

PR #1616, opened 2023-04-06
fix not found check: errors.As for type match whereas errors.Is for value equality

furkhat:

My earlier RP #1526 did not fix the problem (details below).

Fixes #1526

**What type of PR is this?** 

/kind bug

It seems I managed to fool myself (and others) by using errors.Is. From errors.Is doc - "An error is considered to match a target if it is equal to that target..." which means even if there is an error with matching type errors.Is will be false unless its the same error.

The errors.Is is good to use with globally defined (using const / var), exported errors whereas in machine controller default-not-found error is a struct that also wraps some request/response information. Hence for different requests default-not-found error's value is different.

```release-note
machine deletion does not get stuck if assigned floating ip is absent (e.g. deleted externally)
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -764,7 +764,7 @@ func (p *provider) Cleanup(ctx context.Context, machine *clusterv1alpha1.Machine
 		return false, osErrorToTerminalError(err, "failed to get compute client")
 	}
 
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve

kron4eg:
oh wow, that error should be a linter

PR #1622, opened 2023-04-25
WIP: Add notion of community providers to machine-controller

embik:
This PR is both a WIP and a RFC, as we want to make clear that some providers (e.g. Vultr) were developed by external contributors and we cannot guarantee stability, functionality or ongoing maintenance for those provider implementations.

Please leave comments if you have any thoughts on the approach in this PR, the documentation or something else.

Fixes #

**What type of PR is this?**
/kind feature

```release-note
Add `--enable-community-providers` flag that needs to be enabled for using some providers not maintained by the machine-controller developers
```

**Documentation**:

```documentation
NONE
```


embik:
As discussed on SIG cluster-management, we won't go ahead with this.

PR #1638, opened 2023-05-11
Remove DisableStrictZoneCheck from AWS 1.27 cloud-config

xrstf:
https://github.com/kubernetes/cloud-provider-aws/pull/573 removed the `DisableStrictZoneCheck` field alltogether from the cloud-config. The new CCM will reject and fail to load an old cloud-config, so we must ensure that the field is not part of the cloud-config at all.

This PR achieves this by only rendering the field if it's enabled, shifting the burden to set it properly onto our consumers.

**What type of PR is this?**
/kind bug

```release-note
For generating cloud-config for Kubernetes 1.27 machines on AWS, make sure to set `DisableStrictZoneCheck` to `false`, as the flag is not supported in 1.27+ anymore.
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

We'll need the same changes in OSM as well @xrstf

xrstf:
/override pull-machine-controller-e2e-aws-spot-instance
/override pull-machine-controller-e2e-aws-spot-instance-legacy-userdata

AWS is out of capacity for spot instances.

PR #1644, opened 2023-05-16
Switch to cert-manager for certificates

ahmedwaleedmalik:
Use cert-manager for certificates for machine-controller webhooks.

Fixes #

**What type of PR is this?**

/kind chore

```release-note
[ACTION REQUIRED] The default values for webhook flags `tls-cert-path` and `tls-key-path` have been changed to `/tmp/cert/tls.crt` and `/tmp/cert/tls.key` respectively. This is a result of updating it to use certificates generated from cert-manager. 
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -63,8 +63,8 @@ func main() {
 		flag.StringVar(&opt.masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	}
 	flag.StringVar(&opt.admissionListenAddress, "listen-address", ":9876", "The address on which the MutatingWebhook will listen on")
-	flag.StringVar(&opt.admissionTLSCertPath, "tls-cert-path", "/tmp/cert/cert.pem", "The path of the TLS cert for the MutatingWebhook")
-	flag.StringVar(&opt.admissionTLSKeyPath, "tls-key-path", "/tmp/cert/key.pem", "The path of the TLS key for the MutatingWebhook")
+	flag.StringVar(&opt.admissionTLSCertPath, "tls-cert-path", "/tmp/cert/tls.crt", "The path of the TLS cert for the MutatingWebhook")
+	flag.StringVar(&opt.admissionTLSKeyPath, "tls-key-path", "/tmp/cert/tls.key", "The path of the TLS key for the MutatingWebhook")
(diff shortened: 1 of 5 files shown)


cmd/webhook/main.go
```diff
@@ -63,8 +63,8 @@ func main() {
 		flag.StringVar(&opt.masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	}
 	flag.StringVar(&opt.admissionListenAddress, "listen-address", ":9876", "The address on which the MutatingWebhook will listen on")
-	flag.StringVar(&opt.admissionTLSCertPath, "tls-cert-path", "/tmp/cert/cert.pem", "The path of the TLS cert for the MutatingWebhook")
-	flag.StringVar(&opt.admissionTLSKeyPath, "tls-key-path", "/tmp/cert/key.pem", "The path of the TLS key for the MutatingWebhook")
+	flag.StringVar(&opt.admissionTLSCertPath, "tls-cert-path", "/tmp/cert/tls.crt", "The path of the TLS cert for the MutatingWebhook")
+	flag.StringVar(&opt.admissionTLSKeyPath, "tls-key-path", "/tmp/cert/tls.key", "The path of the TLS key for the MutatingWebhook")
```
kron4eg, line 67: As both kubeone and KKP rely on default names:
* this will break kubeone https://github.com/kubermatic/kubeone/blob/2fd69d0e9431877bea5bb33b06bd95e287b17012/addons/machinecontroller/webhook.yaml#L6-L10

* and KKP https://github.com/kubermatic/kubermatic/blob/9aad988df4c621558ada17946f4425ca3e3097b6/pkg/resources/resources.go#L164-L167

kron4eg (changes requested):
other then breaking CLI change, it looks good to me

cmd/webhook/main.go
ahmedwaleedmalik: The idea is to use this in machine-controller and update this in KKP and KubeOne instead. Since cert-manager generates `tls.crt` and `tls.key`. We need to treat machine-controller as a standalone tool in this case.

kron4eg: Sure, I meant to point out this, to not forget to update those filenames. Other approach could be — we shell use flags explicitly in the webhook deployment.

kron4eg (approved):
/lgtm
/approve

kron4eg:
@ahmedwaleedmalik please update the release note with something like "BREAKING CLI change" explaining of what happened with those flags.

PR #1645, opened 2023-05-22
Allow external CCMs to handle node objects before MC

mfranczy:
Allow external CCMs to handle failing node objects before MC takes any action.
It prevents race condition between MC and CCM.

The PR introduces a `handleNodeFailuresWithExternalCCM` function that:

>handleNodeFailuresWithExternalCCM reacts to node status discovery of CCM's node lifecycle controller.
>If an instance at cloud provider is not found then it waits till CCM deletes node objects, that allows:
>   - create a new instance at cloud provider
>   - initialize a new node object - the object should not be reused between instance creation
>     for example, instance foo that got deleted and recreated should initialize a completely new node object
>     instead of reusing the old one as it can cause problems to update node's metadata, like IP address.
>
> If node is shut-down it allows MC to react accordingly to specific cloud provider requirements, those are:
>   - wait for node to become online again or
>   - delete a machine which cannot be recovered

Fixes https://github.com/kubermatic/kubermatic/issues/12218

**What type of PR is this?**
/kind bug

```release-note
Allow external CCMs to handle failing node objects before MC.
```

**Documentation**:

```documentation
NONE
```


mfranczy:
/hold

First, I would like to merge this one https://github.com/kubevirt/cloud-provider-kubevirt/pull/243

pkg/controller/machine/controller.go
```diff
@@ -1242,3 +1259,46 @@ func (r *Reconciler) updateNode(ctx context.Context, node *corev1.Node, modifier
...
+
+	_, err := prov.Get(ctx, log, machine, r.providerData)
+	if err != nil {
+		if cloudprovidererrors.IsNotFound(err) {
+			log.Info("The node does not have corresponding instance, waiting for CCM to delete it")
+			return &reconcile.Result{RequeueAfter: deletionRetryWaitPeriod}, nil
+		}
+		return nil, err
+	} else if taintExists(node, taintShutdown) {
+		switch provConfig.CloudProvider {
```
mfranczy, line 1294: We should check other supported cloud providers regarding instance shutdown, if it can be powered on safely later or we should delete machine and recreate.

mfranczy:
How irrational... KubeVirt tests fail. Investigating...

pkg/controller/machine/controller.go
embik: >  if it can be powered on safely later or we should delete machine and recreate.

Was this meant to say "if it cannot be powered on safely later"?

mfranczy: right

mfranczy: > Was this meant to say "if it cannot be powered on safely later"?

Nope, I meant if machine is shutdown and powered on later.

mfranczy:
KubeVirt tests fails becasue of problem with storage class on the infra cluster. Investigating...

pkg/controller/machine/controller.go
mfranczy: AFAIK some CCMs when discover that instance has been shutdown remove nodes object others don't. I remember OpenStack CCM in the past was deleting node object once you powered off the machine. We have to check if this is still a case for providers we support. However, that should be a different PR and issue.

mfranczy:
@kron4eg @embik tests are ok now. I consider this PR to be ready to review.

PR #1647, opened 2023-05-26
Add vmware anti-affinity for machine sets

### LINKED ISSUE #1340: Support for vsphere anti-affinity rule

- Vsphere has a feature called [VM-VM Affinity/Anti-Affinity Rules](https://docs.vmware.com/en/VMware-vSphere/7.0/com.vmware.vsphere.resmgmt.doc/GUID-94FCC204-115A-4918-9533-BFC588338ECB.html?hWord=N4IghgNiBc4GZwJYDtEBcCeIC+Q). Anti Affinity rule spread VMs to different physical hosts which helps prevent single points of failure. 
- The same feature works for control plane components, and here is [PR#2124](https://github.com/kubermatic/kubeone/pull/2124) at kubeone. Based on [terraform](https://registry.terraform.io/providers/hashicorp/vsphere/latest/docs/resources/compute_cluster_vm_anti_affinity_rule) config.
- If we can have this feature n the machine controller for the vsphere provider, it makes a significant difference.

WeirdMachine:

This PR introduces vmware anti affinity rule creation, per machine.

To ensure a concurrent safe usage of the vsphere api, the create or update operation are synced via mutex which is only unlocked if the api returns the desired change or a timeout appears. 

This feature can be enabled/disabled by setting: `vmAntiAffinity: true` in the cloud provider spec.

Required Permissions:

AutoDeploy:

- Rule:
  - Create
  - Delete
  - Edit

Host:
- Inventory:
  -  Modify cluster

Docs required for the updated permissions.

Fixes #1340 

**What type of PR is this?**

/kind feature

```release-note
Add VMs to an Anti-Affinity rule (vSphere only)
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/pull/1455
```

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/provider.go
+++ b/pkg/cloudprovider/provider/vsphere/provider.go
@@ -23,6 +23,7 @@ import (
 	"net/url"
 	"os"
 	"strings"
+	"sync"
 
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
@@ -48,6 +49,7 @@ import (
 
 type provider struct {
 	configVarResolver *providerconfig.ConfigVarResolver
+	mutex             sync.Mutex
 }
 
 // New returns a VSphere provider.
@@ -64,11 +66,13 @@ type Config struct {
 	Password         string
 	VSphereURL       string
 	Datacenter       string
+	Cluster          string
 	Folder           string
 	ResourcePool     string
 	Datastore        string
 	DatastoreCluster string
 	AllowInsecure    bool
+	VMAntiAffinity   bool
 	CPUs             int32
 	MemoryMB         int64
 	DiskSizeGB       *int64
@@ -164,6 +168,11 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
 		return nil, nil, nil, err
 	}
 
+	c.Cluster, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Cluster)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+
 	c.Folder, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Folder)
 	if err != nil {
 		return nil, nil, nil, err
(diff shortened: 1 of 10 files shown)


embik:
What is the reasoning behind reconciling this based on MachineSets?

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -164,6 +168,11 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
 		return nil, nil, nil, err
 	}
 
+	c.Cluster, err = p.configVarResolver.GetConfigVarStringValue(rawConfig.Cluster)
```
embik, line 171: The comment for this field in the raw config type is:

```
// Cluster is a noop field, it's not used anywhere but left here intentionally for backward compatibility purposes
```

Can we update this, since it's not true anymore?

pkg/cloudprovider/provider/vsphere/provider.go
```diff
@@ -64,11 +66,13 @@ type Config struct {
 	Password         string
 	VSphereURL       string
 	Datacenter       string
+	Cluster          string
```
embik, line 69: As far as I understand, this field needs to be non-empty when configuring anti-affinity, so can we maybe add such validation to `func (p *provider) Validate`? At best, even query the cluster for existence as it's done with e.g. the datastore cluster.

embik (changes requested):

pkg/cloudprovider/provider/vsphere/rule.go
```diff
@@ -0,0 +1,196 @@
...
+func (p *provider) createOrUpdateVMAntiAffinityRule(ctx context.Context, session *Session, name string, config *Config) error {
+	p.mutex.Lock()
+	defer p.mutex.Unlock()
+
+	cluster, err := session.Finder.ClusterComputeResource(ctx, config.Cluster)
+	if err != nil {
+		return err
+	}
+
+	vmsInFolder, err := session.Finder.VirtualMachineList(ctx, strings.Join([]string{config.Folder, "*"}, "/"))
```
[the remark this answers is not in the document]
embik: Okay, I didn’t know that - thanks!

embik (approved):
/approve

embik:
> When can we use this machine-controller in KKP v2.23 and v2.22 as well as kubeone 1.6.x?

Since this will require additional permissions and is a new feature, I'm not sure we are planning to backport it as it has the potential to break existing setups. So it would be shipped with KKP v2.23 when it releases.

embik:
@WeirdMachine small request, can we maybe add the additional permissions needed in the release note so it will be visible to end users?

WeirdMachine:
> > When can we use this machine-controller in KKP v2.23 and v2.22 as well as kubeone 1.6.x?
> 
> Since this will require additional permissions and is a new feature, I'm not sure we are planning to backport it as it has the potential to break existing setups. So it would be shipped with KKP v2.23 when it releases.

This feature is not enabled by default and won't be controllable in the dashboard in the next release. 

However it is possible to enable it in the machine deployment itself, depending on the machine controller version
this should be possible in KKP v2.23

PR #1656, opened 2023-06-06
[release/v1.54] pin cluster-exposer to a Go 1.18 compatible release

xrstf:
This ensures we do not accidentally get a more recent version than we can handle, as `k8s.io/kube-openapi` quickly required Go 1.19/generics. See for example the 2 different kube-openapi versions cloned in https://public-prow.loodse.com/view/gs/prow-dev-public-data/pr-logs/pull/kubermatic_machine-controller/1654/pull-machine-controller-e2e-kubevirt/1664813514599960576

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


embik:
cc @mfranczy for awareness.

PR #1682, opened 2023-07-14
Propagate datastore cluster for vSphere to cloud-config

ahmedwaleedmalik:
In case user has specified `DatastoreCluster` for a machine deployment on vSphere, we should use that as the `datastore` in cloud-config so that the disks attached to the VM are placed accordingly.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
vSphere: Fix a bug where datastore cluster value was not being propagated to the cloud-config
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/vsphere/provider.go
+++ b/pkg/cloudprovider/provider/vsphere/provider.go
@@ -618,6 +618,11 @@ func (p *provider) GetCloudConfig(spec clusterv1alpha1.MachineSpec) (config stri
 		workingDir = fmt.Sprintf("/%s/vm", c.Datacenter)
 	}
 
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik:
> pull-machine-controller-e2e-vsphere-datastore-cluster

Fails because `dsc-1` datacenter doesn't exist.

ahmedwaleedmalik:
> > pull-machine-controller-e2e-vsphere-datastore-cluster
> 
> Fails because `dsc-1` datacenter doesn't exist.

@xrstf can you please override this.

xrstf:
/override pull-machine-controller-e2e-vsphere-datastore-cluster

Issue #1619, opened 2023-04-20
Fix VCD tests


xrstf:
VCD e2e tests are currently failing with

> Error: 400: [ a0acf93e-1c9e-4543-87ba-2e9a4585ebfa ] The entity Ref: com.vmware.vcloud.entity.vapp:6602ebbd-4b07-4aad-9909-022fad2e8efe is busy completing an operation VDC_RECOMPOSE_VAPP."}
2023-04-19T18:29:11.592Z	error	machine-controller	machine/controller.go:402	Reconciling failed	{"machine": "kube-system/ubuntu-1-26-1-vcd-1648753846731673600-5889b9cd

See for example #1606

ahmedwaleedmalik:
@csengerszabo maybe after KKP 2.26 release we can get this prioritized somehow. It's not something urgent but would be great if we can get rid of the flakes/failures.

PR #1700, opened 2023-08-16
Update dependencies

ahmedwaleedmalik:
Updating dependencies, notable ones are:
- k8s API to v1.28.0
- controller-runtime to v1.15.1
- AWS SDK to v1.20.1

Fixes #

**What type of PR is this?**

/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik:
> This script was unable to determine the external IP for kube-apiserver.

The failures till now are just flakes. 
/retest

ahmedwaleedmalik:
@xrstf we'll have to override `pull-machine-controller-e2e-vmware-cloud-director`, we are hitting https://github.com/kubermatic/machine-controller/issues/1619. Not sure when we can finally get it fixed though :/

kron4eg:
is this a good idea? it will force the same changes on KKP and KubeOne

ahmedwaleedmalik:
> is this a good idea? it will force the same changes on KKP and KubeOne

I think when KKP and KubeOne start supporting k8s 1.28 then these dependencies should be updated as well. Similarly with other dependencies as well.

Since this change is only for the main branch at the moment; it'll make its way into KKP 2.24 and KubeOne 1.8.

pkg/cloudprovider/provider/openstack/provider.go
```diff
@@ -705,8 +705,8 @@ func waitForPort(instanceLog *zap.SugaredLogger, netClient *gophercloud.ServiceC
 		return port.Status == "ACTIVE", nil
 	}
 
-	if err := wait.Poll(checkPeriod, checkTimeout, portIsReady); err != nil {
```
xrstf, line 708: Please don't introduce new background contexts, but instead add them to the callstacks where needed to hand them down to where they're needed.

ahmedwaleedmalik: Updated

test/e2e/provisioning/deploymentscenario.go
```diff
@@ -34,6 +34,8 @@ func verifyCreateUpdateAndDelete(kubeConfig, manifestPath string, parameters []s
 	if err != nil {
 		return err
 	}
+	ctx := context.Background()
```
ahmedwaleedmalik, line 37: Hacked it a bit here but if I introducing context at a higher level requires a lot of refactoring since our e2e tests just use new context everywhere.

PR #1706, opened 2023-09-16
Fix isssue of creation of multiple vultr instance due to eventual consistency

### LINKED ISSUE #1704: Vultr: Multiple Instances for Single Machine
## Problem
My team uses an older fork of the Vultr cloud provider, and we've now seen an issue on more than one occasion where Vultr tag state gets duplicated. What we tend to see is two Vultr instances, both with a tag for the same Machine UID.

Our best guess is that tags are only eventually consistent in Vultr, and it is possible for reconciliation to result in rare cases where two Vultr instances are tagged for the same machine. This predictably causes a variety of issues.

We are looking at potentially working on a PR to resolve this scenario, but wanted to verify our approach, so would love any feedback on this issue!

## Planned Approach
Since Vultr returns a unique ID on every create call, we were looking to do something like the `anexia` provider, where rather than relying on the tag in vultr, the provider attaches the Vultr UID into the `ProviderStatus` payload and favors that. The tag would still be set in Vultr (which helps with diagnostics) and could still be used as a fallback, which makes for clean migration to the new tracking model over time.

There are some other advantages to this approach, such as making it easier to find the instance Vultr when a `Get` is performed.

2000yeshu:
There has been a bug introduced because of vultr wherein the tags on instances are eventually consistent due to which machine controller is led to believe that the instance was not created as we poll instance based on label and tags. This PR wait for complete creation of the instance so that eventual fetches to get machine are more deterministic
Additionally, there is the support for Vultr bare metal instances
And suppport for specifying VPC v1/v2 for instances has been added in accordance to https://www.vultr.com/api/#tag/instances/operation/create-instance and https://www.vultr.com/api/#tag/baremetal/operation/create-baremetal

Fixes #1704

**What type of PR is this?**

/kind bug
/kind feature

```release-note
- Fixed multiple instance creation issue for vultr by waiting for instance to get started.
- Add support for vultr bare metal instances
- Upgrade govulr to v3
- Add support for vultr VPC v1/v2
- Handle go vultr API response codes 
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/vultr/provider.go
+++ b/pkg/cloudprovider/provider/vultr/provider.go
@@ -18,11 +18,14 @@ package vultr
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
+	"net/http"
 	"strconv"
+	"time"
 
-	"github.com/vultr/govultr/v2"
+	"github.com/vultr/govultr/v3"
 	"go.uber.org/zap"
 	"golang.org/x/oauth2"
 
@@ -38,7 +41,19 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 )
+
+const (
+	createCheckPeriod           = 10 * time.Second
+	createCheckTimeout          = 5 * time.Minute
+	createCheckFailedWaitPeriod = 10 * time.Second
+)
+
+type ValidVPC struct {
+	IsAllValid  bool
+	InvalidVpcs []string
+}
 
 type provider struct {
 	configVarResolver *providerconfig.ConfigVarResolver
@@ -50,11 +65,17 @@ func New(configVarResolver *providerconfig.ConfigVarResolver) cloudprovidertypes
 }
 
 type Config struct {
-	APIKey string
-	Region string
-	Plan   string
-	OsID   string
-	Tags   []string
+	PhysicalMachine bool
+	APIKey          string
+	Region          string
+	Plan            string
+	OsID            string
+	Tags            []string
+	VpcID           []string
+	EnableVPC       bool
+	EnableIPv6      bool
+	EnableVPC2      bool
+	Vpc2ID          []string
 }
 
 func getIDForOS(os providerconfigtypes.OperatingSystem) (int, error) {
@@ -97,6 +118,7 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
 	}
 
 	c := Config{}
+
 	c.APIKey, err = p.configVarResolver.GetConfigVarStringValueOrEnv(rawConfig.APIKey, "VULTR_API_KEY")
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to get the value of \"apiKey\" field, error = %w", err)
@@ -118,12 +140,80 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
 	}
 
 	c.Tags = rawConfig.Tags
+	c.PhysicalMachine = rawConfig.PhysicalMachine
+	c.EnableIPv6 = rawConfig.EnableIPv6
+	c.VpcID = rawConfig.VpcID
+	c.EnableVPC = rawConfig.EnableVPC
+	c.EnableVPC2 = rawConfig.EnableVPC2
+	c.Vpc2ID = rawConfig.Vpc2ID
 
 	return &c, pconfig, err
 }
 
 func (p *provider) AddDefaults(_ *zap.SugaredLogger, spec clusterv1alpha1.MachineSpec) (clusterv1alpha1.MachineSpec, error) {
 	return spec, nil
+}
+
+func (p *provider) validateVpc(ctx context.Context, client *govultr.Client, c *Config, legacyVPC bool) (ValidVPC, error) {
+	validVpc := ValidVPC{IsAllValid: true}
+	accountvpcs := []string{}
+	var requestedvpcs []string
+
+	if legacyVPC {
+		for {
+			vpcs, meta, err := func(ctx context.Context, client *govultr.Client) ([]govultr.VPC, *govultr.Meta, error) {
+				vpcs, meta, resp, err := client.VPC.List(ctx, &govultr.ListOptions{})
+				if err != nil {
+					return nil, nil, vltErrorToTerminalError(resp.StatusCode, err)
+				}
+				defer resp.Body.Close()
+
+				return vpcs, meta, nil
+			}(ctx, client)
+			if err != nil {
+				return validVpc, err
+			}
+			for _, v := range vpcs {
+				accountvpcs = append(accountvpcs, v.ID)
+			}
+			if meta.Links.Next == "" {
+				break
+			}
+		}
+		requestedvpcs = c.VpcID
+	} else {
+		for {
+			vpcs, meta, err := func(ctx context.Context, client *govultr.Client) ([]govultr.VPC2, *govultr.Meta, error) {
+				vpcs, meta, resp, err := client.VPC2.List(ctx, &govultr.ListOptions{})
+				if err != nil {
+					return nil, nil, vltErrorToTerminalError(resp.StatusCode, err)
+				}
+				defer resp.Body.Close()
+
+				return vpcs, meta, nil
+			}(ctx, client)
+			if err != nil {
+				return validVpc, err
+			}
+			for _, v := range vpcs {
+				accountvpcs = append(accountvpcs, v.ID)
+			}
+			if meta.Links.Next == "" {
+				break
+			}
+		}
+		requestedvpcs = c.Vpc2ID
+	}
+	accountvpcsset := sets.New[string](accountvpcs...)
+	// Iterator to provide user the exact mismatches
+	for _, v := range requestedvpcs {
+		if !accountvpcsset.Has(v) {
+			validVpc.IsAllValid = false
+			validVpc.InvalidVpcs = append(validVpc.InvalidVpcs, v)
+		}
+	}
+
+	return validVpc, nil
 }
 
 func (p *provider) Validate(ctx context.Context, _ *zap.SugaredLogger, spec clusterv1alpha1.MachineSpec) error {
@@ -155,12 +245,13 @@ func (p *provider) Validate(ctx context.Context, _ *zap.SugaredLogger, spec clus
 
 	client := getClient(ctx, c.APIKey)
 
-	plans, err := client.Region.Availability(ctx, c.Region, "")
+	plans, resp, err := client.Region.Availability(ctx, c.Region, "")
 
 	// TODO: Validate region separately
 	if err != nil {
-		return fmt.Errorf("invalid/not supported region specified %q: %w", c.Region, err)
+		return err
 	}
+	resp.Body.Close()
 
 	planFound := false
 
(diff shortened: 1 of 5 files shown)


embik:
/ok-to-test

@2000yeshu FYI, the "documentation" block is only meant to provide a link to a https://github.com/kubermatic/docs PR. Looks like what you added should be in the "release notes" block, right above that.

2000yeshu:
Ah, fixed it.

pkg/cloudprovider/provider/vultr/provider.go
```diff
@@ -126,6 +154,55 @@ func (p *provider) AddDefaults(_ *zap.SugaredLogger, spec clusterv1alpha1.Machin
...
+	accountvpcs := []string{}
+	var requestedvpcs []string
+
+	if legacyVPC {
+		for {
+			vpcs, meta, resp, err := client.VPC.List(ctx, &govultr.ListOptions{})
+			if err != nil {
+				return validVpc, vltErrorToTerminalError(resp.StatusCode, err)
+			}
+			defer resp.Body.Close()
```
lanwen, line 168: If I may raise a concern here: usage of `defer` in a loop could have unpredictable side effects, as doesn't close anything until the end of the loop. 

More info: https://blog.learngoprogramming.com/gotchas-of-defer-in-go-1-8d070894cb01

It might be beneficial to do something like this to be on the safe side (with the correct types):

```go
                     vpcs, meta, err := func(ctx, client) (any, any, error) {
				vpcs, meta, resp, err := client.VPC.List(ctx, &govultr.ListOptions{})
				if err != nil {
					return nil, nil, vltErrorToTerminalError(resp.StatusCode, err)
				}
				defer resp.Body.Close()
				
				return vpcs, meta
			}(ctx, client)
```

(could be other variations moving more into function, like continuation condition)

pkg/cloudprovider/provider/vultr/provider.go
```diff
@@ -168,52 +245,168 @@ func (p *provider) Validate(ctx context.Context, _ *zap.SugaredLogger, spec clus
...
+		UserData: base64.StdEncoding.EncodeToString([]byte(userdata)),
+		Tags:     tags,
+
+		EnableIPv6: &c.EnableIPv6,
+		EnableVPC:  &c.EnableVPC,
+		AttachVPC:  c.VpcID,
+		EnableVPC2: &c.EnableVPC2,
+		AttachVPC2: c.Vpc2Id,
+	}
+	instance, resp, err := client.Instance.Create(ctx, &instanceCreateRequest)
```
lanwen, line 379: I might be unfamiliar with vultr API, so it might be a completely silly question, however, shouldn't be response body closed here, as I noticed you do that in some other cases?

pkg/cloudprovider/provider/vultr/provider.go
2000yeshu: Ah, good catch. Fixed it.

pkg/cloudprovider/provider/vultr/provider.go
2000yeshu: You are right, it should be closed. They are not closing it in the underlying function

2000yeshu: Fixed it.

pkg/cloudprovider/provider/vultr/types/types.go
```diff
@@ -22,11 +22,17 @@ import (
...
-	OsID   providerconfigtypes.ConfigVarString `json:"osId"`
-	Tags   []string                            `json:"tags,omitempty"`
+	PhysicalMachine bool                                `json:"physicalMachine,omitempty"`
+	APIKey          providerconfigtypes.ConfigVarString `json:"apiKey,omitempty"`
+	Region          providerconfigtypes.ConfigVarString `json:"region"`
+	Plan            providerconfigtypes.ConfigVarString `json:"plan"`
+	OsID            providerconfigtypes.ConfigVarString `json:"osId"`
+	Tags            []string                            `json:"tags,omitempty"`
+	VpcID           []string                            `json:"vpcId,omitempty"`
+	Vpc2Id          []string                            `json:"vpc2Id,omitempty"`
```
lanwen, line 32:
```suggestion
	Vpc2ID          []string                            `json:"vpc2Id,omitempty"`
```
Just to make sure it wasn't a miss - you seem to follow the uppercase notation for `ID` everywhere, but not in this case, is that intentional?

pkg/cloudprovider/provider/vultr/provider.go
```diff
@@ -50,11 +65,17 @@ func New(configVarResolver *providerconfig.ConfigVarResolver) cloudprovidertypes
...
+	APIKey          string
+	Region          string
+	Plan            string
+	OsID            string
+	Tags            []string
+	VpcID           []string
+	EnableVPC       bool
+	EnableIPv6      bool
+	EnableVPC2      bool
+	Vpc2Id          []string
```
lanwen, line 78:
```suggestion
	Vpc2ID          []string
```

To make sure it was an intentional difference with other cases where `ID` used uppercase. Would it make sense to follow here the same style?

pkg/cloudprovider/provider/vultr/types/types.go
2000yeshu: No actually i prefer snakecase but the lint was failing due to that. So I made this partial change. Will make it consistent.

pkg/cloudprovider/provider/vultr/provider.go
```diff
@@ -155,64 +245,188 @@ func (p *provider) Validate(ctx context.Context, _ *zap.SugaredLogger, spec clus
...
+func (p *provider) waitForInstanceCreation(ctx context.Context, c *Config, instance instance.Instance, machine *clusterv1alpha1.Machine) error {
+	return wait.PollUntilContextTimeout(ctx, createCheckPeriod, createCheckTimeout, false, func(ctx context.Context) (bool, error) {
+		var err error
+		if !c.PhysicalMachine {
+			_, err = p.getVirtualMachine(ctx, c, machine)
+		} else {
+			_, err = p.getPhysicalMachine(ctx, c, machine)
+		}
+
+		if err != nil && err != cloudprovidererrors.ErrInstanceNotFound {
```
realjenius, line 363: When testing this latest push I saw that it skips this entire conditional, then returns true, but then the instance can't be found during reconciliation and it creates another.

I think the `err != cloudprovidererrors.ErrInstanceNotFound` prevents it from going into this retry loop correctly. It's basically the same check as `if cloudprovidererrors.IsNotFound(err)`. I removed the `err != cloudprovidererrors.ErrInstanceNotFound` in my local testbench and it looks to have created a single instance and waited correctly.

pkg/cloudprovider/provider/vultr/provider.go
```diff
@@ -155,64 +245,188 @@ func (p *provider) Validate(ctx context.Context, _ *zap.SugaredLogger, spec clus
...
+				// just that our instance was not found
+				return false, nil
+			}
+			if isTerminalErr, _, _ := cloudprovidererrors.IsTerminalError(err); isTerminalErr {
+				return true, err
+			}
+			// Wait for some time as instance creation is successful
+			// just that we are not able to fetch it
+			time.Sleep(createCheckFailedWaitPeriod)
+			return false, fmt.Errorf("instance %q created but controller failed to fetch instance details", instance.Name())
```
realjenius, line 375: I think if you return an error, the poll loop stops and returns the error - I don't think that's the intention here; it might be better to log the error and return nil instead so it keeps retrying.

2000yeshu: As this error is returned when there is a problem with the vultr client itself and the List() request fails inside getPhysicalMachine() or getVirtualMachine().
I feel that that kind of error should not be ignored and loop should be broken returning the reason as to why the client returned an error. It could happen that if vultr server is down or the client is misconfigured, this wait function might never return. What do you think?

pkg/cloudprovider/provider/vultr/provider.go
2000yeshu: Fixed.

pkg/cloudprovider/provider/vultr/provider.go
realjenius: Yeah, I suppose it could go either way. There is still a possibility that Vultr is temporarily unreachabler due to some API issue (gateway failure, rate limiting, scheduled maintenance) and we could still get into a duplicate instance scenario.

I think the wait function still has a hard timeout of 5 minutes, right? So I think it'd probably eventually bail out either way. Maybe 5 minutes is a bit excessive in this scenario; but at the same time, if the provider gets this far, then the instance create call completed without error, so at least in theory things are most likely out of sync if the code bails out.

I'm less worried about this particular flow, however, as it's definitely an edge case.

I saw some of the other providers go into a "Cleanup" cycle if they fail creation - is that something that would be a better option here? It might leave an orphaned instance in Vultr, but it wouldn't result in a "two instance" reconciliation problem as the machine for that label would be gone, right?

realjenius (approved):

pkg/cloudprovider/provider/vultr/provider.go
2000yeshu: Cleanup cycle when the Instance creation itself fails or the waiting for label fails?
In any case, what kind of metadata will we give the vultr API to delete instance? 
Tags might not have applied to the instance and can't give the label as the controller might have created another instance for the machine.

embik:
Hey @2000yeshu, you'll need to address the [lint job failing](https://public-prow.loodse.com/view/gs/prow-dev-public-data/pr-logs/pull/kubermatic_machine-controller/1706/pull-machine-controller-lint/1704558461922054144):

```
pkg/cloudprovider/provider/vultr/provider.go:365:47: `succesfully` is a misspelling of `successfully` (misspell)
				// Continue the loop as the instances was succesfully fetched
				                                          ^
```

For the other e2e jobs I'm not sure what is going on, we will need to take a look from our side.

pkg/cloudprovider/provider/vultr/provider.go
2000yeshu: An ideal way i could think of is to know whether machine-controller controlled label exists on the machine or not. Currently we can't know whether the UID label is present on machine or not. If we could know that, we can add "wait"s wherever (instance.label == machine.name && UID label DNE) either when creating instance or when getting the instance. If we can't, we(provider) gives error.
As for the cleanup i feel that, maybe we can do it syncronously after the "wait"s and delete the instance if the wait ended with no error and no success. But that would need controller changes.

I think the cleanup could be done here as well, after the wait, let me try that.

2000yeshu: Pushed.

2000yeshu:
> Hey @2000yeshu, you'll need to address the [lint job failing](https://public-prow.loodse.com/view/gs/prow-dev-public-data/pr-logs/pull/kubermatic_machine-controller/1706/pull-machine-controller-lint/1704558461922054144):
> 
> ```
> pkg/cloudprovider/provider/vultr/provider.go:365:47: `succesfully` is a misspelling of `successfully` (misspell)
> 				// Continue the loop as the instances was succesfully fetched
> 				                                          ^
> ```
> 
> For the other e2e jobs I'm not sure what is going on, we will need to take a look from our side.

Done

embik:
/retest

Looking at test history, those tests just randomly flake. Let's hope 1-2 retests bring them in.

2000yeshu:
I saw that they are failing because of "machinedeployment not found". Maybe ignore the `kubectl delete` stderr?

embik:
> I saw that they are failing because of "machinedeployment not found". Maybe ignore the `kubectl delete` stderr?

I believe that the actual error is hidden somewhere. It doesn't even try to create MachineDeployments, so the test fails pretty early on. But this is not related to your PR at all, we need to do some digging on our side why it's so flaky.

embik (approved):
/approve

Thank you for contributing to machine-controller's community providers!

PR #1725, opened 2023-10-23
Drop centos support for Hetzner

ahmedwaleedmalik:
Hetzner has deprecated their CentOS 7 image. CentOS 7 was/is the only release we ever supported and so this now means that there is no CentOS support at all anymore on Hetzner.

Fixes #

**What type of PR is this?**

/kind cleanup

```release-note
Removed support for centos with Hetzner
```

**Documentation**:

```documentation
NONE
```


xmudrii (approved):
/lgtm
/approve

PR #1751, opened 2023-12-19
webhook: refactor to use controller-runtime server

LittleFox94:

Refactor the webhook to use the controller-runtime provided webhook.Server instead of a plain HTTP server. The biggest benefit this brings, is it automatically loading changed certificate files instead of forever using the ones present when the server was started.

**What type of PR is this?**

I'm not quite sure, I'll label this as feature but the webhook server not reloading changed certificates could also be seen as a bug.
/kind feature

This could have been implemented by using the controller-runtime CertWatcher manually, but that would've been more code and aligning machine-controller with the ecosystem-provided standards seems worth it.

```release-note
webhook: changed certificate/key files are now reloaded automatically
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/cmd/webhook/main.go
+++ b/cmd/webhook/main.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
 )
 
 type options struct {
@@ -143,6 +144,13 @@ func main() {
 		NodeFlags:            nodeFlags,
 		Namespace:            opt.namespace,
 		VersionConstraints:   constraint,
+
+		// we could change this to get the CertDir from the configured CertName
+		// and KeyName, but doing so does not bring us any benefits but would
+		// technically break compatibility.
+		CertDir:  "/",
+		CertName: opt.admissionTLSCertPath,
+		KeyName:  opt.admissionTLSKeyPath,
 	}.Build()
 	if err != nil {
 		log.Fatalw("Failed to create admission hook", zap.Error(err))
@@ -150,12 +158,8 @@ func main() {
 
 	log.Infow("Listening", "address", opt.admissionListenAddress)
 
-	if err := srv.ListenAndServeTLS(opt.admissionTLSCertPath, opt.admissionTLSKeyPath); err != nil {
+	serverContext := signals.SetupSignalHandler()
+	if err := srv.Start(serverContext); err != nil {
 		log.Fatalw("Failed to start server", zap.Error(err))
 	}
-	defer func() {
-		if err := srv.Close(); err != nil {
-			log.Fatalw("Failed to shutdown server", zap.Error(err))
-		}
-	}()
 }
--- a/pkg/admission/admission.go
+++ b/pkg/admission/admission.go
@@ -22,9 +22,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"reflect"
-	"time"
+	"strconv"
 
 	"github.com/Masterminds/semver/v3"
 	"go.uber.org/zap"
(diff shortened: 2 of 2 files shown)


cmd/webhook/main.go
```diff
@@ -143,19 +144,25 @@ func main() {
 		NodeFlags:            nodeFlags,
 		Namespace:            opt.namespace,
 		VersionConstraints:   constraint,
+
+		// we could change this to get the CertDir from the configured CertName
+		// and KeyName, but doing so does not bring us any benefits but would
+		// technically break compatibility.
```
LittleFox94, line 150: technically breaks compatibility if we enforce TLS cert and key being in the same directory

cmd/webhook/main.go
```diff
@@ -143,19 +144,25 @@ func main() {
...
+	serverContext, cancelServerContext := context.WithCancel(context.Background())
+	if err := srv.Start(serverContext); err != nil {
 		log.Fatalw("Failed to start server", zap.Error(err))
 	}
 	defer func() {
-		if err := srv.Close(); err != nil {
-			log.Fatalw("Failed to shutdown server", zap.Error(err))
-		}
+		cancelServerContext()
 	}()
```
LittleFox94, line 167: I'm not sure if this is actually still needed, together with the cancel-context above

embik:
thank you for this!

/ok-to-test

LittleFox94:
heya @embik :)
all tests green, if you don't have comments on the two threads I started on the code, wanna press that merge button for me? :)

embik:
> heya @embik :) all tests green, if you don't have comments on the two threads I started on the code, wanna press that merge button for me? :)

Because it's the silent time of the year, I'd like to give colleagues a few more days to voice any second opinions (PR looks good from my perspective though) before merging it if that's okay by you. So ETA would be beginning of January.

cmd/webhook/main.go
embik: I think you can just do

```go
serverContext, cancelServerContext := context.WithCancel(context.Background())
defer cancelServerContext()
if err := srv.Start(serverContext); err != nil {
    log.Fatalw("Failed to start server", zap.Error(err))
}
```

cmd/webhook/main.go
```diff
@@ -143,19 +144,25 @@ func main() {
...
+		KeyName:  opt.admissionTLSKeyPath,
 	}.Build()
 	if err != nil {
 		log.Fatalw("Failed to create admission hook", zap.Error(err))
 	}
 
 	log.Infow("Listening", "address", opt.admissionListenAddress)
 
-	if err := srv.ListenAndServeTLS(opt.admissionTLSCertPath, opt.admissionTLSKeyPath); err != nil {
+	serverContext, cancelServerContext := context.WithCancel(context.Background())
```
embik, line 161: Okay, looks like I do have something to comment on: This shouldn't use `context.Background()` but a context that cancels on SIGINT/SIGTERM so the server shuts down gracefully. Then you can probably drop the `defer cancelServerContext()` as well that I commented on below (don't think we need a `context.WithCancel` then).

You can use  [signals.SetupSignalHandler](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/manager/signals#SetupSignalHandler) for that (like we do over in the [seed-controller-manager](https://github.com/kubermatic/kubermatic/blob/acf679a2dd5572a6a20168c8b6a493f02afb7cb1/cmd/seed-controller-manager/main.go#L109)).

embik (changes requested):

cmd/webhook/main.go
LittleFox94: I was thinking about that, but when I run machine-controller locally on my machine, it already reacts perfectly fine to SIGINT - you sure this is needed?

embik: Sorry for the late reply. I'm not 100% _sure_ it's needed, but it's the proper way to handle signals and might cover more corner cases than the current code. I would prefer it, if only for consistency with the KKP code base.

cmd/webhook/main.go
LittleFox94: not done, according to https://github.com/kubermatic/machine-controller/pull/1751#discussion_r1439205717

cmd/webhook/main.go
LittleFox94: alright, changed code according to this :)

Sorry for my late reply as well, was still off the first week of the year ^^  
Have a good new year btw :)

embik: Happy new year! Hope you got to enjoy the days off :)

embik (approved):
/approve

thank you for the contribution!

PR #1753, opened 2024-01-03
Anexia: Adds field to check if a IP reservation is still valid

89Q12:
Adds a field to check if a IP reservation is still valid, if the IP is no longer valid a new one is requested.

This fixes a bug where we use a IP reservation that expired already for a VM, if the VM provisioning failed, which then leads to a race condition. In which we wait for the VM provisioning to finish, but it can't due the IP being no longer useable, but we also never request a new IP in that case.

**What type of PR is this?**
/kind bug

Please backport to version v1.57

```release-note
Anexia: Fix IP reservation related race condition.
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik (approved):
/approve

PR #1760, opened 2024-02-16
Remove TORCX_BINDIR from containerd binary call

embik:
Same PR as https://github.com/kubermatic/operating-system-manager/pull/357, just for machine-controller. torcx has been removed from Flatcar, we need to remove the hard dependency on it in our containerd unit drop-in. MC does not try to load the torcx environment file, so this looks slightly different from the OSM PR.

Fixes #

**What type of PR is this?**
/kind bug

```release-note
Fix containerd configuration for Flatcar nodes on stable 3815.2.0 or higher
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve 🚀

PR #1772, opened 2024-03-18
Bump dependencies

xrstf:
It's KubeCon time! Most maintainers are at the conference, so it's the perfect time for general housekeeping. This PR updates all direct dependencies.

**What type of PR is this?**
/kind chore

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


pkg/cloudprovider/provider/vmwareclouddirector/helper.go
```diff
@@ -121,9 +120,7 @@ func createVM(client *Client, machine *clusterv1alpha1.Machine, c *Config, org *
 			}
 		}
 		if storageProfile == nil {
-			if err != nil {
-				return fmt.Errorf("failed to get storage profile '%s': %w", *c.StorageProfile, err)
-			}
+			return fmt.Errorf("failed to get storage profile '%s'", *c.StorageProfile)
```
xrstf, line 123: This is an actual change in behaviour. Before this, because of the `err != nil` condition, it was impossible to have an error returned here. But now without the condition, if no storage profile is found, an error will be returned.

ahmedwaleedmalik (approved):
/apporve

PR #1789, opened 2024-04-30
Remove user-data plugins from machine-controller

### LINKED ISSUE #1788: Remove user-data plugins from machine-controller
User-data plugins from machine-controller were deprecated in favor of  [Operating System Manager](https://github.com/kubermatic/operating-system-manager) with https://github.com/kubermatic/machine-controller/pull/1642 over a year ago. 

It's finally time to remove the user-data plugins completely.

ahmedwaleedmalik:
## User data plugins from machine-controller have been removed in favor of OSM.

Historically, machine-controller relied on user data plugins to generate configurations(cloud-config) for worker machines. Each operating system requires its own user-data plugin. These configs are then injected into the worker nodes using provisioning utilities such as [cloud-init](https://cloud-init.io) or [ignition](https://coreos.github.io/ignition). Eventually, the nodes are bootstrapped.

Over time, it has been observed that this workflow has certain limitations.

- Machine Controller expects **ALL** the supported user-data plugins to exist and be ready. User might only be interested in a subset of the available operating systems. For example, user might only want to work with `ubuntu`.
- The user-data plugins have templates defined [in-code](https://github.com/kubermatic/machine-controller/blob/v1.53.0/pkg/userdata/ubuntu/provider.go#L136). Which is not ideal since code changes are required to update those templates. Then those code changes need to become a part of the subsequent releases for machine-controller and KubeOne. So we need a complete release cycle to ship those changes to customers.
- Managing configs for multiple cloud providers, OS flavors and OS versions, adds a lot of complexity and redundancy in machine-controller.
- Since the templates are defined in-code, there is no way for an end user to customize them to suit their use-cases.
- Each cloud provider sets some sort of limits for the size of `user-data`, machine won't be created in case of non-compliance. For example, at the time of writing this, AWS has set a [hard limit of 16KB](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-add-user-data.html).
- Better support for air-gapped environments is required.

**To overcome these limitations; Kubermatic built [Operating System Manager](https://github.com/kubermatic/operating-system-manager)** a few years ago. Please refer to [Operating System Manager](https://docs.kubermatic.com/operatingsystemmanager) for more details.

These plugins were deprecated about a year ago in #1642. 

User data plugins from machine-controller have been removed in favor of OSM.

Fixes #1788

**What type of PR is this?**

/kind cleanup

There will be a follow-up PR with some more cleanup, moving things around from MC to OSM. But since it requires changes in OSM first, we'll do it later in another PR.

```release-note
[ACTION REQUIRED] User data plugins for machine-controller have been removed. [Operating System Manager](https://github.com/kubermatic/operating-system-manager) is the successor of user data plugins. It's responsible for creating and managing the required configurations for worker nodes in a Kubernetes cluster with better modularity and extensibility. Please refer to [Operating System Manager](https://github.com/kubermatic/operating-system-manager#kubermatic-operating-system-manager) for more details.
```

**Documentation**:

```documentation
https://docs.kubermatic.com/operatingsystemmanager
```


ahmedwaleedmalik:
@xrstf pull-machine-controller-e2e-vsphere, pull-machine-controller-e2e-azure-custom-image-reference needs to be overridden. 

Maybe in the future, we can run these tests periodically, once a day, so that we are more aware of issues like https://github.com/kubermatic/operating-system-manager/pull/381 and generally failing tests like pull-machine-controller-e2e-azure-custom-image-reference

kron4eg (commented):
LOVE IT!

small change proposed

kron4eg:
/lgtm cancel

there is failing linter job

ahmedwaleedmalik:
> /lgtm cancel
> 
> there is failing linter job

Should be fixed now, even more cleanup :D

xrstf:
/override pull-machine-controller-e2e-azure 
/override pull-machine-controller-e2e-azure-custom-image-reference

PR #1791, opened 2024-05-04
Delete provSpec redundant check in MC providers

mohamed-rafraf:

No need for this check in the `getConfig()` function
```	
if provSpec.Value == nil {
	return nil, nil, fmt.Errorf("machine.spec.providerconfig.value is nil")
}
```
because it already exists in `providerconfigtypes.GetConfig()` function

Fixes #

**What type of PR is this?**

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/alibaba/provider.go
+++ b/pkg/cloudprovider/provider/alibaba/provider.go
@@ -347,10 +347,6 @@ func (p *provider) SetMetricsForMachines(_ clusterv1alpha1.MachineList) error {
 }
 
 func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *providerconfigtypes.Config, error) {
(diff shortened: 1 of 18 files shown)


ahmedwaleedmalik:
Looks good. Would it be possible for you to do the same for other instances where we also have this redundant check? It should only be required here https://github.com/kubermatic/machine-controller/blob/main/pkg/providerconfig/types/types.go#L386-L388

ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik:
@mohamed-rafraf you'll need overrides for the failing tests. @xrstf is the person to reach for that.

ahmedwaleedmalik:
/hold to avoid retries

@mohamed-rafraf please remove the `hold` when the issue with failing tests is fixed

PR #1810, opened 2024-06-05
prevent forgetting to pin the cluster-exposer

xrstf:
We forgot it again for 1.58 and 1.59 (1.59 didn't hurt yet because KKP main is still on Go 1.22), so I'd say it's time we accept reality and just use a fixed cluster-exposer version. It's not like it changed over the last years..

**What type of PR is this?**
/kind failing-test

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik:
My suggestion is to not build this in machine-controller, etc., at all and move it to a pinned version in the build image. That way, we don't have to modify these scripts needlessly.

xrstf:
Yeah maybe we have to tackle that one finally.

xrstf:
/hold

https://github.com/kubermatic/infra/pull/2710 would make this obsolete

xrstf:
#1813 supersedes this.

PR #1830, opened 2024-07-25
Refactor Tinkerbell provider

moadqassem:
In a previous PR [here](https://github.com/kubermatic/machine-controller/pull/1815),  we started a major refactoring for the tinkerbell provider adding new features and removing deprecated ones. This PR however, address a lot of changes for the original one as the current author is on a long vacation and we need this changes in ASAP. For more details on the functionality and what points this PR amends please take a look at the previous PR and its review.  

Fixes #

**What type of PR is this?**

```release-note
Refactor the tinkerbell provider
```

**Documentation**:

```documentation
None
```


pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/hardware.go
```diff
@@ -18,94 +18,80 @@ package client
...
-	tinkHardware, err := method(ctx, req)
-	if err != nil {
-		if err.Error() == sqlErrorString || err.Error() == sqlErrorStringAlt {
-			return nil, fmt.Errorf("hardware %w", ErrNotFound)
+	// Find the Hardware with the given ID
+	var targetHardware tinkv1alpha1.Hardware
+	for _, hw := range hardwares.Items {
+		if hw.Spec.Metadata.Instance.ID == uid {
+			targetHardware = hw
+			return &targetHardware, nil
```
xrstf, line 92: Wouldn't `return &hw, nil` work just as well? Why a new variable if the loop ends due to the return statement anyway?

moadqassem: That's true, I guess this left by mistake after cleaning up some code. Will take care of that 😉

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
```diff
@@ -20,73 +20,252 @@ import (
...
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// Template client for Tinkerbell.
+type Task struct {
+	Name        string            `json:"name"`
+	WorkerAddr  string            `json:"worker" yaml:"worker"`
```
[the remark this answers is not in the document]
moadqassem: This was left on purpose because of a Tinkerbell issue in the template object. In the API it's called worker however in the Tinkerbell internal parser it still called workeraddr. Tinkerbell has two different templates, the Kubernetes one and the internal one. This is due to historical reasons that tink server used to run gRPC not kube.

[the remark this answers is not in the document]
xrstf: but the tags are the same..

Also can we leave a comment that these types are copied (I presume) from https://github.com/tinkerbell/tink/blob/main/internal/deprecated/workflow/types.go ?

Also that URL makes me wonder if we're building this new implementation based on deprecated types?

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
```diff
@@ -20,73 +20,252 @@ import (
...
-
-		return nil, fmt.Errorf("getting template from Tinkerbell: %w", err)
+func (t *TemplateClient) Delete(ctx context.Context, namespacedName types.NamespacedName) error {
+	template := &tinkv1alpha1.Template{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      namespacedName.Name,
+			Namespace: namespacedName.Namespace,
+		},
+	}
+	// Create the Template object in the Tinkerbell cluster
```
xrstf, line 84: This comment is wrong.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
```diff
@@ -20,73 +20,252 @@ import (
...
+		Actions:    actions,
+	}
+
+	template := Template{
+		Name:          "ubuntu",
+		Version:       "0.1",
+		GlobalTimeout: 1800,
+		Tasks:         []Task{task},
+	}
+	// Marshal task to YAML
```
xrstf, line 148: This comment isn't adding anything and can be removed instead of fixed.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
```diff
@@ -20,73 +20,252 @@ import (
...
-		if err.Error() == sqlErrorString || err.Error() == sqlErrorStringAlt {
-			return fmt.Errorf("template %w", ErrNotFound)
-		}
+}
 
-		return fmt.Errorf("deleting template from Tinkerbell: %w", err)
+func createGrowPartitionAction(destDisk string) Action {
+	return Action{
+		Name:    "grow-partition",
+		Image:   "quay.io/tinkerbell/actions/cexec:latest",
```
xrstf, line 195: This could also be pinned to c5bde803d9f6c90f1a9d5e06930d856d1481854c, which is the same as "latest" at the moment. Or is there a reason not to pin this image?

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/utils.go
```diff
@@ -14,19 +14,26 @@ See the License for the specific language governing permissions and
...
+	length, _ := mask.Size()
 
-// than parsing for these specific error message.
-const (
-	sqlErrorString    = "rpc error: code = Unknown desc = sql: no rows in result set"
-	sqlErrorStringAlt = "rpc error: code = Unknown desc = SELECT: sql: no rows in result set"
-	sqlErrorNotFound  = "rpc error: code = NotFound desc = not found"
-)
+	cidr := ""
+	parts := strings.Split(ip.Address, ".")
```
xrstf, line 32: This will break with IPv6, no?

moadqassem: IPv6 is not supported

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
moadqassem: Why? There is always enhancements that er com into latest and all our tinkerbell actions are using latest,

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
```diff
@@ -18,205 +18,193 @@ package tinkerbell
...
 
-type ClientFactory func() (metadataclient.Client, tinkerbellclient.HardwareClient, tinkerbellclient.TemplateClient, tinkerbellclient.WorkflowClient)
-
 type driver struct {
-	TinkServerAddress string
-	ImageRepoAddress  string
+	ClusterName string
+	OSImageURL  string
+	TinkClient  ctrlruntimeclient.Client
+	//KubeClient     ctrlruntimeclient.Client
```
xrstf, line 46: Is this a leftover?

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
moadqassem: This is part of the leftovers. Will be removed.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
moadqassem: This is part of the leftovers. Will be removed.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
```diff
@@ -18,205 +18,193 @@ package tinkerbell
...
-		ImageRepoAddress:  imageRepoAddress,
-		metadataClient:    mdClient,
-		hardwareClient:    hwClient,
-		templateClient:    tmplClient,
-		workflowClient:    wflClient,
+	d := driver{
+		ClusterName: tinkSpec.ClusterName.Value,
+		TinkClient:  tinkClient,
+		HardwareRef: tinkSpec.HardwareRef,
+		//KubeClient:     k8sClient,
```
xrstf, line 81: Another leftover

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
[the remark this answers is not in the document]
moadqassem: This is not something that we can control, again if we remove that, things work! FYI we are using the latest tink stack and still this is a problem. This is a mitigation for a bug in the tink-contorller:
`,"error":"parsing yaml data: yaml: unmarshal errors:\n  line 6: field workeraddr not found in type workflow.Task",` this is still in the latest releases.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
moadqassem: Probably yeah, it is from the original PR. I just wanted to know what's the idea of this field. But this was at the beginning of my implementation, but since I have refactored quite a lot I can simply take it out.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
```diff
@@ -18,205 +18,193 @@ package tinkerbell
...
+
+		if iface.Netboot != nil && iface.Netboot.AllowWorkflow != nil && iface.Netboot.AllowWorkflow == ptr.Bool(false) {
+			continue
 		}
+
+		allowProvision = true
+	}
+
+	if !allowProvision {
+		log.Infow("server is not allowed to be provisioned; either hardware allowPXE and allowWorkflow is set tp false", "hardware", hardware.Name)
```
xrstf, line 126: tp -> to

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
moadqassem: Probably yeah, it is from the original PR. I just wanted to know what's the idea of this field. But this was at the beginning of my implementation, but since I have refactored quite a lot I can simply take it out.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
```diff
@@ -18,205 +18,193 @@ package tinkerbell
...
+		return nil, nil
```
xrstf, line 127: Returning nil, nil here will make the baremetal provider return a baremetalServer struct with nil in it, which will then later explode because its GetName() functions are not capable of handling nils.

If a server cannot be provisioned, it should IMHO be an error that can bubble up and not hidden in the MC log where few people would ever look (especially not the user who doesn't have access to MC logs in KKP systems).

moadqassem: Good catch! I agree this makes sense

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/workflow.go
```diff
@@ -18,203 +18,89 @@ package client
...
-		}
-
-		if err != nil {
-			return nil, fmt.Errorf("getting workflow event from Tinkerbell: %w", err)
-		}
-
-		result = append(result, e)
+	workflow := &tinkv1alpha1.Workflow{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      workflowName + "-workflow",
```
xrstf, line 56: Please somehow make this single line re-usable. At the moment it's both an implementation detail, but also used by the Tinkerbell driver directly (i.e. it has knowledge it should not have).

Also, is it intended that "-workflow" is appended to the _workflow_ name, or is the parameter for this function misnamed? In the driver, the server's name is used for this parameter.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
```diff
@@ -18,205 +18,193 @@ package tinkerbell
...
+	// Get the hardware object from tinkerbell cluster
+	targetHardware, err := d.HardwareClient.GetHardware(ctx, d.HardwareRef)
+	if err != nil {
+		return err
 	}
 
-	if hw.Hardware.Network == nil {
-		return fmt.Errorf("tinkerbell hardware network configs can not be empty")
+	// Delete the associated Workflow.
+	workflowName := targetHardware.Name + "-workflow" // Assuming workflow names are derived from hardware names
```
xrstf, line 162: (this is the spot I mean regarding leaking implementation logic)

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
```diff
@@ -18,205 +18,193 @@ package tinkerbell
...
+	if err := d.WorkflowClient.DeleteWorkflow(ctx, workflowName, targetHardware.Namespace); err != nil {
+		return fmt.Errorf("failed to delete workflow %s: %w", workflowName, err)
 	}
 
-	if hw.Hardware.Metadata == "" {
-		return fmt.Errorf("tinkerbell hardware metadata can not be empty")
+	// Reset the hardware ID in the tinkerbell cluster.
```
xrstf, line 167: ID _and state_, so it's more obvious that from this point on, the machine is also not Staged anymore.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
```diff
@@ -18,205 +18,193 @@ package tinkerbell
...
 	}
 
 	return nil
 }
 
-func (d *driver) DeprovisionServer(ctx context.Context, uid types.UID) error {
-	if err := d.hardwareClient.Delete(ctx, string(uid)); err != nil {
-		if resourceNotFoundErr(err) {
-			return nil
+func GetConfig(driverConfig tinktypes.TinkerbellPluginSpec, aa func(configVar providerconfigtypes.ConfigVarString, envVarName string) (string, error)) (*tinktypes.Config, error) {
```
xrstf, line 175: What is `aa`?

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/workflow.go
moadqassem: Yes we can but honestly I took over this PR only yesterday and I cannot fix everything in one shot 😅 I can address that in up coming PRs

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
moadqassem: No idea tbh! part of the previous implementation(The PR that I rebased my changes on top, the one mentioned in the description). But I can take a look

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
```diff
@@ -18,205 +18,193 @@ package tinkerbell
...
+	// Kubeconfig was specified directly in the Machine/MachineDeployment CR. In this case we need to ensure that the value is base64 encoded.
+	if driverConfig.Auth.Kubeconfig.Value != "" {
+		val, err := base64.StdEncoding.DecodeString(driverConfig.Auth.Kubeconfig.Value)
+		if err != nil {
+			// An error here means that this is not a valid base64 string
+			// We can be more explicit here with the error for visibility. Webhook will return this error if we hit this scenario.
+			return nil, fmt.Errorf("failed to decode base64 encoded kubeconfig. Expected value is a base64 encoded Kubeconfig in JSON or YAML format: %w", err)
+		}
+		config.Kubeconfig = string(val)
+	} else {
```
xrstf, line 187: Do we really want to support this? Inlining kubeconfigs has never been a pleasant thing in KKP and I thought it's better practice to reference Secrets instead.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/types/hardware.go
```diff
@@ -0,0 +1,69 @@
...
+var _ plugins.Server = &Hardware{}
+
+func (h *Hardware) GetName() string {
+	return h.Name
+}
+
+func (h *Hardware) GetID() string {
+	if h.Spec.Metadata != nil &&
+		h.Spec.Metadata.Instance != nil &&
+		h.Spec.Metadata.Instance.ID != "" {
```
xrstf, line 37: `&& h.Spec.Metadata.Instance.ID != ""` is redundant and can be removed

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
moadqassem: I kinda agree with you tbh, and it is less error prune. Ok I will adjust that to be read from a secret instead of Inlining  it. But can I do it in a follow up PR instead? I kinda need this image now 😂

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/types/hardware.go
```diff
@@ -0,0 +1,69 @@
...
+
+	return ""
+}
+
+func (h *Hardware) GetStatus() string {
+	if h.Status.State != "" {
+		return string(h.Status.State)
+	}
+
+	return "Unknown"
```
xrstf, line 68: To make comparisons easier, couldn't this be one of the constants? Now this function would need to be understood as "returns a proper State type [no matter that it's currently just a string], or the random string 'Unknown'".

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/types/types.go
```diff
@@ -0,0 +1,64 @@
...
+package types
+
+import (
+	providerconfigtypes "github.com/kubermatic/machine-controller/pkg/providerconfig/types"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	Staged         string = "Staged"
```
xrstf, line 26: A dedicated `type HardwareState string` would help identify what these constants are used for.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
xrstf: I guess it's a sideeffect from using providerconfigtypes.ConfigVarString for the field, but I would be totally okay with commenting that inline strings are not supported for this field.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/types/hardware.go
moadqassem: How is this redundant  ? You can have a hardware object without an id.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/types/types.go
moadqassem: Honestly I added that then I took out as I actually see no real value other than keep casting the value when comparing it with the API field. Especially the API doesn't know this `HardwareState ` or isn't considering as an Enum.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/types/hardware.go
xrstf: Yeah, but if the ID is an empty string, you can just return it. Because the default would be to return an empty string anyway. You are adding a bit of unnecessary logic to this function. The function doesn't need to concern itself with _how_ the ID field looks like, it should just return it if instance metadata exists, regardless of its value.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/types/hardware.go
moadqassem: True can be addressed as well

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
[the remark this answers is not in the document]
xrstf: Okay, the mismatch between the Go field "WorkerAddr" and the tag names "worker" are from upstream, fine.

But I still do not understand the tags here.

Upstream uses only YAML tags: https://github.com/tinkerbell/tink/blob/main/internal/deprecated/workflow/types.go#L13-L32

We copied (?) the types `Action` and `Task` from Tinkerbell, changed the YAML tags to JSON, and then added the `Actions` as YAML again to the `Template` struct. And ultimately we render templates as YAML.

So why the JSON fields for some structs?

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/driver.go
moadqassem: Ok found out, will rename after its functionality

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
xrstf: You pinned every single other image in every single other action. Except this one. Surely there had to be a reason. :) When I look at the code, I might now have accidentally broke it by thinking "oh, someone forgot to pin this, let me just fix that". So a comment why this one action is not pinned to a version would be appreciated.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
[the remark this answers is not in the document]
moadqassem: Not really! In the upstream API we also use JSON not only YAML. For example the hardware and template objects are JSON. The workflow type in v1alpha1 is the only thing that uses yaml. And even in the new v1alpha2 we started using JSON as well. And as I said this was mainly for historical reasons of how Tinkerbell object were handled, again Kubernetes was introduced for some time now but we still have some code which still leveraging the previous backend.

pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go
moadqassem: I guess you have overlooked something, while the tags are pinned for the all the images, due the the image repo and path are different 😉 

```
quay.io/tinkerbell/actions/cexec:latest
```

```
quay.io/tinkerbell-actions/writefile:v1.0.0
```

moadqassem: Anyway you really would like to have this in, I have no problem with that 😉. Will pin it down

PR #1831, opened 2024-07-30
Refactor bare metal provider machine cleanup

moadqassem:
In Tinkerbell a workflow brings together [Hardware](https://tinkerbell.org/docs/concepts/hardware) and a [Template](https://tinkerbell.org/docs/concepts/templates) for execution.. The status field of the workflow can be used to determine the current state of the workflow and to identify any issues that may have occurred during the execution of it. Thus, it is important to keep track of those workflows even if they were executed to keep track of the provisioning processes. This PR makes the workflows of a hardware unique and it opens up the possibility to run a workflow multiple times and treat them as an idempotent process which can't be resumed in terms of failures. Those workflows must be deleted in two situations: 

1- The workflow has a pending state however the machine deployment has been deleted. Thus we should get rid of those workflows to prevent triggering them in the future. 
2- The workflow is stuck in a timeout state and it cannot be resumed thus they don't have any real value.

This PR cleans up those workflows based on the criteria above.

Fixes #

**What type of PR is this?**
/kind feature

```release-note
Tinerkebll: cleanup workflows that are either in a pending state or timeout state.
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/hardware.go
+++ b/pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/hardware.go
@@ -62,7 +62,6 @@ func (h *HardwareClient) SetHardwareID(ctx context.Context, hardware *tinkv1alph
 	}
 
 	hardware.Spec.Metadata.Instance.ID = newID
-	// Set the new ID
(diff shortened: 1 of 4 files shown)


xrstf (approved):
/approve

PR #1833, opened 2024-07-31
AWS: Increase HttpPutResponseHopLimit to 3 for EC2 instances

ahmedwaleedmalik:
The latest Ubuntu 22.04 images have IMDSv2 enabled as a default. By default, the limit of hops of PUT requests to metadata service is 2. The instance metadata service is not reachable from the container network in case if further hops are required.

We came across this issue while working with Cilium as CNI. The issue has been documented [here](https://github.com/cilium/cilium/issues/25232).

Although, the issue mentions a workaround to make it work with Cilium. Fixing it just for a single component is not ideal since a tool like `machine-controller` shouldn't judge how the underlying network or programs hosted on the machines provisioned by it are configured. Hence, we went with a more generic fix for this, i.e. to set the [HTTP PUT response hop limit](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceMetadataOptionsRequest.html#:~:text=HttpPutResponseHopLimit) to 3. This ensures a much wider compatibility.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
AWS: Increase HttpPutResponseHopLimit to 3 for EC2 instances
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/aws/provider.go
+++ b/pkg/cloudprovider/provider/aws/provider.go
@@ -61,6 +61,11 @@ const (
 	// Interval and timeout for polling.
 	pollInterval = 2 * time.Second
 	pollTimeout  = 5 * time.Minute
+	// The maximum number of hops that the metadata service can be forwarded to, defaults to 2.
+	// We need to set this to a higher value i.e. 3 to ensure that it is not blocked by extra hops that are introduced either by CNI or other networking components. With lower
+	// limits AWS metadata service is not reachable from the container network in such a scenario.
+	// For example: https://github.com/cilium/cilium/issues/25232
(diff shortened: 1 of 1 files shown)


xmudrii (approved):
/lgtm
/approve

xmudrii:
/hold
Looks like we don't run AWS tests by default in MC, let me quickly check this

alexinfoblox:
@ahmedwaleedmalik, hello
We encountered the same issue when retrieving the session token from the instance metadata. We are also using Cilium as the cluster CNI (no issues were observed with Flannel).

Therefore, we are also considering increasing the hop limit to 3. What potential drawbacks or concerns should we be aware of when increasing it?"

PR #1844, opened 2024-08-16
Fix vSphere E2E tests

xmudrii:
**What this PR does / why we need it**:

Some parameters on the vSphere side are temporarily changed. This PR ensures that the vSphere E2E tests can pass.

**What type of PR is this?**

/kind failing-test

**Does this PR introduce a user-facing change? Then add your Release Note here**:
```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

PR #1847, opened 2024-08-26
vSphere: support for VM Groups

### LINKED ISSUE #1846: vSphere: Support for VM Groups
As a KKP user, I'd like to be able to assign machine deployments to specific VM groups to control their placement in a deterministic manner.

## Acceptance Criteria

- [ ] Assign VM/Host group to VMs

ahmedwaleedmalik:
This PR adds support for adding VMs to pre-defined VM groups. This is useful for VM placement as host rules can be created for VM groups in vSphere.

Fixes #1846

**What type of PR is this?**

/kind feature

```release-note
vSphere: Support for assigning VMs to VM groups. 
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik:
vSphere tests are supposed to be fixed in https://github.com/kubermatic/machine-controller/pull/1844. So we'll need to override them in this PR so that I can continue with the implementation in KKP and dashboard.

pkg/cloudprovider/provider/vsphere/vmgroup.go
```diff
@@ -0,0 +1,121 @@
...
+			vmRefs = append(vmRefs, vm.Reference())
+		}
+	}
+
+	var vmRefsToAdd []types.ManagedObjectReference
+	for _, vm := range vmRefs {
+		for _, existingVM := range vmGroup.Vm {
+			if existingVM.Value == vm.Value {
+				log.Debugf("VM %s already in VM group %s", machine.Name, config.VMGroup)
+				continue
```
mohamed-rafraf, line 61: If we find the VM in the VM group we can break the inner loop instead.

pkg/cloudprovider/provider/vsphere/vmgroup.go
```diff
@@ -0,0 +1,121 @@
...
+			}
+		}
+		vmRefsToAdd = append(vmRefsToAdd, vm)
```
mohamed-rafraf, line 64: In this case you append very VM to VMRefsToAdd regardless of whether it was already in the group or not

ahmedwaleedmalik: Yeah I mistakenly added continue instead of break :)

PR #1854, opened 2024-08-29
Remove Pod Network from a VM

moadqassem:

Fixes #

**What type of PR is this?**

```release-note

```

**Documentation**:

```documentation

```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -652,11 +652,6 @@ func (p *provider) newVirtualMachine(_ context.Context, c *Config, pc *providerc
 		annotations[k] = v
 	}
 
-	defaultBridgeNetwork, err := defaultBridgeNetwork(macAddressGetter)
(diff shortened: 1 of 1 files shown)


cnvergence:
I see this config was added after an issue:
https://github.com/kubermatic/machine-controller/issues/1202

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -788,6 +779,7 @@ func randomMacAddressGetter() (string, error) {
 	return mac.String(), nil
 }
 
+// TODO(mq): refactor networking in VMs
 func defaultBridgeNetwork(macAddressGetter macAddressGetter) (*kubevirtv1.Interface, error) {
```
cnvergence, line 783: needs to be removed, because of the lint

moadqassem:
This PR is not needed for now as this needs to be investigated further,

moadqassem:
This PR is not needed for now as this needs to be investigated further.

PR #1860, opened 2024-09-25
Support topology aware zones and regions in KubeVirt

moadqassem:
Support for the two labels `topology.kubernetes.io/region` and `topology.kubernetes.io/zone` labels where they can deployed on both Node level(The kubelet or the external cloud-controller-manager populates this with the information from the cloud provider) and on PVC level(topology-aware volume provisioners will automatically set node affinity constraints on a PersistentVolume.). 

A zone represents a logical failure domain. It is common for Kubernetes clusters to span multiple zones for increased availability and a region represents a larger domain, made up of one or more zones. 

Fixes #

**What type of PR is this?**
/kind feature

```release-note
region and zone aware labels support
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -977,21 +977,3 @@ func getTopologySpreadConstraints(config *Config, matchLabels map[string]string)
 		},
 	}
 }
-
-//func getStorageTopologies(ctx context.Context, storageClasName string, client client.Client) (map[string]string, error) {
-//	sc := &storagev1.StorageClass{}
-//	if err := client.Get(ctx, types.NamespacedName{Name: storageClasName}, sc); err != nil {
-//		return err, nil
-//	}
-//
-//	topologies := make(map[string]string)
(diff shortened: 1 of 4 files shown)


pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -958,3 +977,21 @@ func getTopologySpreadConstraints(config *Config, matchLabels map[string]string)
 		},
 	}
 }
+
+//func getStorageTopologies(ctx context.Context, storageClasName string, client client.Client) (map[string]string, error) {
```
yaa110, line 981: The fully commented code can be removed.

yaa110 (changes requested):

pkg/cloudprovider/provider/kubevirt/provider.go
moadqassem: Damn! forgot about it :D

moadqassem: Done

yaa110 (approved):

PR #1862, opened 2024-10-04
Support KubeVirt Provider Network

moadqassem:
KubeOvn has some features that can be used along side with KubeVirt. One of those features is subnet and vpc support. This PR supports the KubeOVN VPC and Subnet as a provider network in KubeVirt.

Fixes #

**What type of PR is this?**
/kind feature

```release-note
Support KubeOVN as a provider network in KubeVirt
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -19,6 +19,7 @@ package kubevirt
 import (
 	"context"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
@@ -703,7 +704,7 @@ func (p *provider) newVirtualMachine(c *Config, pc *providerconfigtypes.Config, 
 
 	annotations["kubevirt.io/allow-pod-bridge-network-live-migration"] = "true"
 
-	if c.ProviderNetworkName == "KubeOVN" {
+	if strings.ToLower(c.ProviderNetworkName) == "kubeovn" {
 		if err := setOVNAnnotations(c, annotations); err != nil {
 			return nil, fmt.Errorf("failed to set OVN annotations: %w", err)
 		}
@@ -1079,7 +1080,7 @@ func setOVNAnnotations(c *Config, annotations map[string]string) error {
 
 		firstIP := ipNet.IP.To4()
 		if firstIP == nil {
-			return fmt.Errorf("invalid IPv4 address")
+			return errors.New("invalid IPv4 address")
 		}
 
 		firstIP[3]++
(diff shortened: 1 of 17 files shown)


pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -686,9 +701,14 @@ func (p *provider) newVirtualMachine(c *Config, pc *providerconfigtypes.Config,
 		annotations["kubevirt.io/ignitiondata"] = userdata
 	}
 
-	annotations["ovn.kubernetes.io/allow_live_migration"] = "true"
 	annotations["kubevirt.io/allow-pod-bridge-network-live-migration"] = "true"
 
+	if c.ProviderNetworkName == "KubeOVN" {
```
yaa110, line 706: Wouldn't it be better to compare the strings in lowercase?

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -1042,3 +1062,37 @@ func getStorageTopologies(ctx context.Context, storageClasName string, c *Config
...
+			return fmt.Errorf("invalid IPv4 address")
+		}
+
+		firstIP[3]++
+		subnetGatewayIP = firstIP.String()
+	} else {
+		subnetGatewayIP = c.SubnetGatewayIP
+	}
+
+	annotations["ovn.kubernetes.io/routes"] = fmt.Sprintf(`|
```
yaa110, line 1091: Wouldn't it be better to create a struct and use the `json` package to marshal it into JSON, rather than constructing the JSON string manually?

```go
type Routes struct {
    Gw string `json:"gw"`
}

routes := []Routes {{ Gw: subnetGatewayIP }}
b, _ := json.Marshal(&routes)
annotations["ovn.kubernetes.io/routes"] = string(b)
```

In this case, you won't need to manually handle escapes and multiline strings.

yaa110 (changes requested):

pkg/cloudprovider/provider/kubevirt/provider.go
moadqassem: Well, this is not really a string comparison per se, it is semantically types comparison. However I see your point, I can lowercase it unconditionally

pkg/cloudprovider/provider/kubevirt/provider.go
moadqassem: Actually I tried that at the beginning and I got an error from kubeovn.,

moadqassem: I will double check though but I am pretty sure I got something incorrect

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -1042,3 +1062,37 @@ func getStorageTopologies(ctx context.Context, storageClasName string, c *Config
 
 	return nil
 }
+
+func setOVNAnnotations(c *Config, annotations map[string]string) error {
```
kron4eg, line 1066: Would be cool to see effect of this function materialized in the testdata.

moadqassem: Has been added

yaa110 (approved):

PR #1878, opened 2024-11-17
Support Importing Volume Over HTTP Using Basic Auth

moadqassem:
One way to import a a virtual machine disk, is via an http endpoint that holds a raw image. So far this method didn't support basic auth to interact with the image repo server that holds these raw images. This PR supports sending the basic auth as a header in the CDI DataVolume object by reading the headers either explicitly from the Machine Deployment spec or referring to  a secret that contains those headers.

`ExtraHeaders` field in the MachineDeployment spec is used to read those headers explicitly and send them as part of the importing request.

`ExtraHeadersFromSecret` is used to read those headers from a secret that is being referenced in the MachineDeployment object.

/kind feature

Fixes #

**What type of PR is this?**

This PR has been tested manually and it succeeded.

```release-note
Support `ExtraHeaders  in the KubeVirt Provider 
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -114,6 +114,8 @@ type Config struct {
 	Region                    string
 	Zone                      string
 	EnableNetworkMultiQueue   bool
+	ExtraHeaders              []string
+	ExtraHeadersSecretRef     string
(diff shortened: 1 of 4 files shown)


cnvergence (approved):
/approve
LGTM

PR #1897, opened 2025-01-27
Support kubevirt evection strategy in kv provider

moadqassem:
Currently, the KubeVirt VMs which are created using machine controller have an eviction strategy of `External` . With `EvictionStrategy: External`, kubevirt will create a PDB for the VMI which blocks eviction, and it will also set the `vmi.Status.EvacuationNodeName` on the vmi's status. When a VMI is being evicted, we need a way to have that eviction blocked so the cluster api related controllers can properly drain the k8s node running within the VMI before the VMI is torn down.

However, in some scenarios, users would like to be able to migrate VMs without blocking the migration by machine controller thus we would like to enable the users to decide which eviction strategy they would like to use.

Fixes #

**What type of PR is this?**
/kind feature

```release-note
Support multiple eviction strategy for KubeVirt VMs
```

**Documentation**:

```documentation
None
```


ahmedwaleedmalik (approved):
/approve

PR #1900, opened 2025-02-28
Move APIs and cloud provider types into standalone Go module in ./sdk

xrstf:
To make it easier to use the machine-controller's types and APIs, this PR moves the Kube APIs, cloud provider types, providerconfig and userdata into a new Go module in the `sdk` directory. Consumers can then simply import `k8c.io/machine-controller/sdk` and avoid having to also import heavy cloud provider dependencies.

No changes were made to the actual API definitions.

**What type of PR is this?**
/kind api-change

```release-note
Introduce a new `k8c.io/machine-controller/sdk` Go module that contains the Kube APIs, cloud provider types, providerconfig etc.
```

**Documentation**:
```documentation
NONE
```


sdk/Makefile
```diff
@@ -0,0 +1,18 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+.PHONY: lint
+lint:
+	golangci-lint run --verbose --print-resources-usage ./...
+	golangci-lint run --verbose --print-resources-usage --config ../.golangci.yml ./...
```
xmudrii, line 18: Why do we run golangci-lint two times? Can you leave a comment in the Makefile why do we do this if this is intentional?

xmudrii (approved):
Looks fantastic! :shipit: 
/lgtm
/approve

PR #1906, opened 2025-03-10
Change KubeVirt VM CPU assignment to not overwrite cpu alloc ratio

### LINKED ISSUE #1905: kubevirt cpu allocation ratio is ignored for provisioned virtual machines
Currently the machine-controller sets requests and limits for cpu and memory resources of a kubevirt virtual machine. (https://github.com/kubermatic/machine-controller/blob/main/pkg/cloudprovider/provider/kubevirt/provider.go#L755-#L762) Due to this kubevirt ignores to calculate the [cpu ratio](https://kubevirt.io/user-guide/compute/node_overcommit/#node-cpu-allocation-ratio) for the launcher pod.

To let kubevirt calculate the cpu requests for a launcher pod the configured cpus needs to be set in `spec.domain.cpu` instead of configuring the requested cpus and the limit in `spec.domain.resources`. This leads then to another challenge because then no limit would be set to the launcher pod by default. For that the kubevirt need to have cpu [auto limit feature](https://kubevirt.io/user-guide/compute/resources_requests_and_limits/#auto-cpu-limits) enabled which also requires additional configuration by either [adding labels to namespaces](https://kubevirt.io/user-guide/compute/resources_requests_and_limits/#autocpulimitnamespacelabelselector-configuration) or [applying resource quotas](https://kubevirt.io/user-guide/compute/resources_requests_and_limits/#autoresourcelimitsgate-feature-gate) for namespaces.

It would be nice to be able to use this kubevirt feature and to have a documentation regarding the requirements to use it.

soer3n:
This pr changes the assignment of cpu resources for kubevirt virtual machines to be able to use kubevirt `cpuAllocationRatio` feature by introducing a new field to the providerSpec called vCPUs.

Fixes #1905

**What type of PR is this?**

/kind feature

```release-note
A new field `cloudProviderSpec.virtualMachine.template.vcpus` was added to kubevirt provider. When `vcpus.cores` is set, the cpus for the virtual machine will be configured in `spec.template.spec.domain.cpu` instead of setting resource requests and limits explicit. The webhook validation will fail if both `vcpus` and `cpus` are configured for a machine.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -616,7 +616,7 @@ func (p *provider) Validate(ctx context.Context, _ *zap.SugaredLogger, spec clus
 	// If instancetype is specified, skip CPU and Memory validation.
 	// Values will come from instancetype.
 	if c.Instancetype == nil {
-		if _, err := parseResources(c.CPUs, c.Memory); err != nil {
+		if _, err := parseResources(c.Memory); err != nil {
 			return err
 		}
 	}
@@ -753,7 +753,7 @@ func (p *provider) newVirtualMachine(c *Config, pc *providerconfig.Config, machi
 
 	// if no instancetype, resources are from config.
 	if c.Instancetype == nil {
-		requestsAndLimits, err := parseResources(c.CPUs, c.Memory)
+		requestsAndLimits, err := parseResources(c.Memory)
 		if err != nil {
 			return nil, err
 		}
@@ -801,6 +801,12 @@ func (p *provider) newVirtualMachine(c *Config, pc *providerconfig.Config, machi
 		runStrategy = kubevirtcorev1.RunStrategyAlways
 	}
 
+	cpusAsUint64, err := strconv.ParseUint(c.CPUs, 0, 64)
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse cpu cores: %w", err)
+	}
+
 	virtualMachine := &kubevirtcorev1.VirtualMachine{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      machine.Name,
@@ -822,6 +828,9 @@ func (p *provider) newVirtualMachine(c *Config, pc *providerconfig.Config, machi
 						*kubevirtcorev1.DefaultPodNetwork(),
 					},
 					Domain: kubevirtcorev1.DomainSpec{
+						CPU: &kubevirtcorev1.CPU{
+							Cores: uint32(cpusAsUint64),
+						},
 						Devices: kubevirtcorev1.Devices{
 							Interfaces:                 []kubevirtcorev1.Interface{*defaultBridgeNetwork},
 							Disks:                      getVMDisks(c),
@@ -867,18 +876,13 @@ func (p *provider) Cleanup(ctx context.Context, _ *zap.SugaredLogger, machine *c
 	return false, sigClient.Delete(ctx, vm)
 }
 
-func parseResources(cpus, memory string) (*corev1.ResourceList, error) {
+func parseResources(memory string) (*corev1.ResourceList, error) {
 	memoryResource, err := resource.ParseQuantity(memory)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse memory requests: %w", err)
 	}
-	cpuResource, err := resource.ParseQuantity(cpus)
-	if err != nil {
-		return nil, fmt.Errorf("failed to parse cpu request: %w", err)
-	}
 	return &corev1.ResourceList{
 		corev1.ResourceMemory: memoryResource,
-		corev1.ResourceCPU:    cpuResource,
 	}, nil
 }
 
--- a/pkg/cloudprovider/provider/kubevirt/provider.go
+++ b/pkg/cloudprovider/provider/kubevirt/provider.go
@@ -98,6 +98,7 @@ type Config struct {
 	DNSConfig                 *corev1.PodDNSConfig
 	DNSPolicy                 corev1.DNSPolicy
 	CPUs                      string
+	VCPUsEnabled              bool
 	Memory                    string
 	Namespace                 string
 	OSImageSource             *cdicorev1beta1.DataVolumeSource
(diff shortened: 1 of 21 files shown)


pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -840,6 +840,19 @@ func (p *provider) newVirtualMachine(c *Config, pc *providerconfig.Config, machi
...
+		cpusAsUint64, err := strconv.ParseUint(c.CPUs, 0, 64)
+
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse cpu cores: %w", err)
+		}
+
+		virtualMachine.Spec.Template.Spec.Domain.CPU = &kubevirtcorev1.CPU{
+			Cores: uint32(cpusAsUint64),
+		}
+	}
```
ahmedwaleedmalik, line 854: If this annotation is consumed in machine-controller then this should be configured at `machine.Annotations` not in the spec.

sdk/cloudprovider/kubevirt/types.go
```diff
@@ -25,6 +25,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
+const (
+	UseDedicatedKubevirtVCPUAnnotationKey = "kubermatic.io/use-dedicated-kubevirt-cpu"
```
ahmedwaleedmalik, line 29:
```suggestion
	UseDedicatedKubeVirtVCPUAnnotationKey = "kubermatic.io/use-dedicated-kubevirt-cpu"
```

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -840,6 +840,19 @@ func (p *provider) newVirtualMachine(c *Config, pc *providerconfig.Config, machi
 			DataVolumeTemplates: getDataVolumeTemplates(c, dataVolumeName, dvAnnotations),
 		},
 	}
+
+	if isDedicatedVCPURequested(machine.Spec.Annotations) {
+		cpusAsUint64, err := strconv.ParseUint(c.CPUs, 0, 64)
+
```
ahmedwaleedmalik, line 846: No need for space here

ahmedwaleedmalik (changes requested):

ahmedwaleedmalik:
1. Please update the release note and mention the annotation explicitly that you have introduced
2. This is a bug or a feature request and not a chore
3. I have updated the title. Please follow proper casing.

sdk/cloudprovider/kubevirt/types.go
moadqassem, line 29: I think it is better if you actually use the RawConfig instead of annotations.

soer3n:
@ahmedwaleedmalik I changed the kind to feature and modified the release notes accordingly to the changes i added after discussing the implementation with @moadqassem.

sdk/cloudprovider/kubevirt/types.go
```diff
@@ -70,12 +70,20 @@ type Flavor struct {
...
 	Memory         providerconfig.ConfigVarString `json:"memory,omitempty"`
 	PrimaryDisk    PrimaryDisk                    `json:"primaryDisk,omitempty"`
 	SecondaryDisks []SecondaryDisks               `json:"secondaryDisks,omitempty"`
 }
 
+// VCPUs.
+type VCPUs struct {
+	Cores   int `json:"cores,omitempty"`
+	Sockets int `json:"sockets,omitempty"`
+	Threads int `json:"threads,omitempty"`
```
ahmedwaleedmalik, line 84: Both Sockets and Threads are unused. We should either remove these variables or utilize them at https://github.com/kubermatic/machine-controller/blob/8b52fcc5f41a2d4e5c55b67e51e364e69b50157d/pkg/cloudprovider/provider/kubevirt/provider.go#L893 and https://github.com/kubermatic/machine-controller/blob/8b52fcc5f41a2d4e5c55b67e51e364e69b50157d/pkg/cloudprovider/provider/kubevirt/provider.go#L854-L856

ahmedwaleedmalik (changes requested):

ahmedwaleedmalik:
Looks much better as part of the providerconfig 🚀  

Same feedback as the first point from https://github.com/kubermatic/machine-controller/pull/1906#issuecomment-2717977729, in release note please explicitly mention the "new field" that you have introduced i.e. `cloudProviderSpec.virtualMachine.template.vcpus`

sdk/cloudprovider/kubevirt/types.go
```diff
@@ -70,12 +70,20 @@ type Flavor struct {
 
 // Template.
 type Template struct {
+	VCPUs          VCPUs                          `json:"vcpus,omitempty"`
```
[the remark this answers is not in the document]
soer3n: Added explanation to both fields `VCPUs` amd `CPUs`.

sdk/cloudprovider/kubevirt/types.go
soer3n: I removed the unused values and only left `cores` for now.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -616,8 +625,12 @@ func (p *provider) Validate(ctx context.Context, _ *zap.SugaredLogger, spec clus
...
 	// Values will come from instancetype.
 	if c.Instancetype == nil {
-		if _, err := parseResources(c.CPUs, c.Memory); err != nil {
-			return err
+		if c.Resources == nil {
+			return fmt.Errorf("no memory requests set for the virtual machine")
+		}
+
+		if (c.VCPUs == nil && c.Resources.Cpu() == nil) || (c.VCPUs != nil && c.Resources.Cpu() != nil) {
+			return fmt.Errorf("invalid/no cpus configured. Either vpus or cpus have to be configured.")
```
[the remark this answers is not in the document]
soer3n: Thanks for the suggestion. I separated both conditions and changed the return message to be a bit more clearer.

ahmedwaleedmalik (approved):
/approve

PR #1907, opened 2025-03-11
(chore): update logs: fix incorrect secret name used in log

buraksekili:

Previously, machine-controller was printing an incorrect secret name when the cloud-init secret had not been created yet. This PR updates the log statement to correctly display the secret name that machine-controller is looking for.

Fixes #

**What type of PR is this?**

/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/controller/machine/controller.go
+++ b/pkg/controller/machine/controller.go
@@ -806,16 +806,23 @@ func (r *Reconciler) ensureInstanceExistsForMachine(
 				return nil, fmt.Errorf("failed to find machine's MachineDployment: %w", err)
 			}
 
-			bootstrapSecretName := fmt.Sprintf(bootstrap.CloudConfigSecretNamePattern,
-				referencedMachineDeployment,
-				machine.Namespace,
-				bootstrap.BootstrapCloudConfig)
+			bootstrapSecretKey := types.NamespacedName{
+				Name: fmt.Sprintf(
+					bootstrap.CloudConfigSecretNamePattern,
+					referencedMachineDeployment,
+					machine.Namespace,
+					bootstrap.BootstrapCloudConfig,
+				),
+				Namespace: util.CloudInitNamespace,
+			}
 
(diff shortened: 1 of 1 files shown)


pkg/controller/machine/controller.go
```diff
@@ -806,16 +806,23 @@ func (r *Reconciler) ensureInstanceExistsForMachine(
...
 
 			bootstrapSecret := &corev1.Secret{}
 			if err := r.client.Get(ctx,
-				types.NamespacedName{Name: bootstrapSecretName, Namespace: util.CloudInitNamespace},
-				bootstrapSecret); err != nil {
-				log.Errorw("cloud-init configuration: cloud config is not ready yet", "secret", bootstrap.BootstrapCloudConfig)
+				bootstrapSecretKey,
+				bootstrapSecret,
+			); err != nil {
+				log.Errorw("cloud-init configuration: cloud config is not ready yet", "secret", bootstrapSecretKey.String())
```
ahmedwaleedmalik, line 824: The intention of the log is to print the type of cloud config that is not ready yet which is `bootstrap` config in this case, similar to https://github.com/kubermatic/machine-controller/blob/main/pkg/controller/machine/controller.go#L824. 

I'm indifferent since it's just a log but if you are modifying it then maybe you can modify the error message as well to something like:

```go
log.Errorw(fmt.Sprintf("cloud-init configuration: cloud config is not ready yet. Bootstrap secret %q is unavailable", bootstrapSecretKey.String()), zap.Error(err))

```

buraksekili: I see. If the intention of the log is to print the type of cloud config, then it is working as expected. I'll go ahead and close the PR. When I first saw the log, I initially thought it was referring to the 'secret name' rather than the type - so, it was not clear for me.

Thanks again for the review!

buraksekili:
Closed due to https://github.com/kubermatic/machine-controller/pull/1907#discussion_r1991876438

PR #1912, opened 2025-03-26
fix kubevirt cpu check + update sdk version in go.mod

soer3n:

This pr fixes the validation of kubevirt virtual machines for dedicated cpus. With the current check the validation will fail when vcpus are configured because `c.Resources.Cpu()`  will never be nil because at least the format is set in the struct. 

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -51,7 +51,7 @@ require (
 	gomodules.xyz/jsonpatch/v2 v2.4.0
 	google.golang.org/api v0.197.0
 	gopkg.in/yaml.v3 v3.0.1
-	k8c.io/machine-controller/sdk v0.0.0-00010101000000-000000000000
+	k8c.io/machine-controller/sdk v0.0.0-20250314150330-99a4aa5532ca
 	k8s.io/api v0.31.1
 	k8s.io/apiextensions-apiserver v0.31.1
 	k8s.io/apimachinery v0.31.1
--- a/go.mod
+++ b/go.mod
@@ -51,7 +51,7 @@ require (
 	gomodules.xyz/jsonpatch/v2 v2.4.0
 	google.golang.org/api v0.197.0
 	gopkg.in/yaml.v3 v3.0.1
-	k8c.io/machine-controller/sdk v0.0.0-20250314150330-99a4aa5532ca
+	k8c.io/machine-controller/sdk v0.0.0-00010101000000-000000000000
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -51,7 +51,7 @@ require (
 	gomodules.xyz/jsonpatch/v2 v2.4.0
 	google.golang.org/api v0.197.0
 	gopkg.in/yaml.v3 v3.0.1
-	k8c.io/machine-controller/sdk v0.0.0-00010101000000-000000000000
+	k8c.io/machine-controller/sdk v0.0.0-20250314150330-99a4aa5532ca
```
xrstf, line 54: But the SDK is replaced with ./sdk, why do we need to bump this version?

soer3n: Yes, you are right. I reverted that change and tested it again. Not sure what went wrong when i built the image locally yesterday but the webhook threw an error that the new field is unknown and i did a `go get k8c.io/machine-controller/sdk` which was a mistake.

pkg/cloudprovider/provider/kubevirt/provider.go
```diff
@@ -285,9 +285,11 @@ func (p *provider) getConfig(provSpec clusterv1alpha1.ProviderSpec) (*Config, *p
 		return nil, nil, fmt.Errorf(`failed to get value of "memory" field: %w`, err)
 	}
 
-	config.Resources, config.VCPUs, err = parseResources(cpus, memory, rawConfig.VirtualMachine.Template.VCPUs)
-	if err != nil {
-		return nil, nil, fmt.Errorf(`failed to configure resource requests and limits and vcpus: %w`, err)
+	if rawConfig.VirtualMachine.Instancetype == nil {
```
xrstf, line 288: I understand the 2nd part of this diff based on the PR description, but why this new if? Not saying it's wrong, I just don't have the context.

soer3n: I noticed during further testing that the validation fails for a machine deployment when an instance type is specified. Then neither cpus nor vcpus are specified and `parseResources` function will return an error because the called function `resource.ParseQuantity` does not work with an empty string as an input. The test covergage could be better regarding that. I would like to add some tests in another pr.

PR #1915, opened 2025-04-03
Improve SDK, add readme

xrstf:
This adds a readme to the SDK, plus it improves the dependencies a tiny bit: Previously everyone who consumed the providerconfig package also had to import controller-runtime. With this PR, the ConfigVarResolver is now in its own Go
package so it at least becomes an indirect dependency for consumers.

With this, I also updated its interface slightly.

* There is now a `providerconfig.ConfigVarResolver` interface.
* Consumers should replace `providerconfig.NewConfigVarResolver` with `configvar.NewResolver`
* All getter functions on the provider had their "ConfigVar" part removed, so `GetConfigVarStringValueOrEnv` becomes `GetStringValueOrEnv`. This is less stuttery and less repetetive. Also it's less repetitive and stutters less. There are also fewer repetitions.

**What type of PR is this?**
/kind api-change

No release note because the big SDK hange has not been released yet.
```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


xrstf:
/override pull-machine-controller-e2e-equinix-metal

The Equinix project has recently been disabled, we should also disable the tests in the future.

/override pull-machine-controller-e2e-vmware-cloud-director

This new test environment is not yet properly set up to run tests.

PR #1925, opened 2025-05-30
Handle SDK version tags

xrstf:
If SDK and main module tags are on the same commit, GIT_TAG could be "sdk/v...", which is not semver anymore and would break anything. This PR is the therefore the equivalent of https://github.com/kubermatic/kubermatic/pull/14636, ensuring only main version tags are considered.

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ GOOS ?= $(shell go env GOOS)
 
 export CGO_ENABLED := 0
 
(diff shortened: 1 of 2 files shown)


embik (approved):
/approve

PR #1952, opened 2025-08-04
feat(nutanix): use CachedClient, chore(nutanix): more error log information

adberger:

- Add ErrorDetail & ProgressMessage to non-successful Nutanix API calls
- Switch to CachedClient with `session_auth` to reduce authentication requests to the Nutanix API

**What type of PR is this?**

/kind feature
/kind chore


pkg/cloudprovider/provider/nutanix/client.go
```diff
@@ -63,26 +84,58 @@ func GetClientSet(config *Config) (*ClientSet, error) {
...
+	// Add proxy configuration if provided
 	if config.ProxyURL != "" {
-		credentials.ProxyURL = config.ProxyURL
+		proxyURL, err := url.Parse(config.ProxyURL)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse proxy URL: %w", err)
+		}
+
+		// Create a custom transport with proxy
+		transport := &http.Transport{
```
kron4eg, line 130: Here we need some timeouts, PTAL at [http.DefaultTransport](https://pkg.go.dev/net/http#DefaultTransport) for inspiration.

kron4eg (changes requested):
In general looks good, just some minor nit

pkg/cloudprovider/provider/nutanix/client.go
adberger: Thanks for the hint.  
I copied the defaults and committed the changes, shall I rebase?

PR #1963, opened 2025-09-15
add support for rhel 9 for Azure provider

### LINKED ISSUE #1660: Support RHEL 9 as OS
What it says on the title. I'm not sure exactly what is required since I haven't checked what has changed in RHEL 9, but we should work on supporting it eventually.

This requires support in OSM as well.

rajaSahil:
Add support for rhel 9 for Azure provider.

Fixes #1660

**What type of PR is this?**

```release-note
Add support for rhel 9 for Azure provider.
```

**Documentation**:

```documentation
None
```


pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -149,8 +149,8 @@ var imageReferences = map[providerconfig.OperatingSystem]compute.ImageReference{
 	providerconfig.OperatingSystemRHEL: {
 		Publisher: to.StringPtr("RedHat"),
 		Offer:     to.StringPtr("rhel-byos"),
```
kron4eg, line 151: what if we try to use official azure RHEL images?

PR #2002, opened 2026-02-27
Skip eviction in case when the Node is NotReady

### LINKED ISSUE #1586: Don't try to drain NotReady nodes
When the Node is in NotReady condition, machine-controller should avoid trying to drain it on the deletion.

kron4eg:
When the Node is in NotReady condition, it's useless to try to drain it, causes endless loop of drain retrying of half-dead Node.

Fixes #1586

**What type of PR is this?**
/kind bug

```release-note
Skip eviction in case when the Node is NotReady
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #2007, opened 2026-03-30
Fix: Add missing RBAC to access endpointslices

uwej711:

Change RBAC role to endpointslices.

With https://github.com/kubermatic/machine-controller/pull/1992 the fallback to discover the kubernetes API was changed to use endpointslices but the corresponding role was not changed. This leads to healthz resulting in an error and machine-controller not to be healthy (if you rely on the fallback). 

Fixes #

**What type of PR is this?**

/kind bug

```release-note
Add missing RBAC to access endpointslices
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/examples/machine-controller.yaml
+++ b/examples/machine-controller.yaml
@@ -396,9 +396,9 @@ metadata:
     local-testing: "true"
 rules:
 - apiGroups:
-  - ""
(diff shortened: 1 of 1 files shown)


kron4eg (approved):
/approve

PR #2022, opened 2026-05-20
Mirror Tinkerbell images to quay.io/kubermatic-mirror

### LINKED ISSUE #2018: Update images in Tinkerbell template
### Description

The images we currently use in the Tinkerbell template (e.g. in the ["stream-ubuntu-image" action](https://github.com/kubermatic/machine-controller/blob/7a607d859d8e712366c354c167b101a05221043f/pkg/cloudprovider/provider/baremetal/plugins/tinkerbell/client/template.go#L184)) are quite outdated. It would be good to use the latest images.

### Additional information

/label kind/feature
/label customer-request
Internal reference: 9089

buraksekili:

Tinkerbell provisioning templates reference 7 container images from upstream registries (Docker Hub, quay.io/tinkerbell, ghcr.io). If any of those registries are unavailable, BMaaS provisioning fails. This PR mirrors all 7 images to `quay.io/kubermatic-mirror/images/` pinned by sha256 digest, and updates `template.go` to reference only the mirrored copies.

A `mirror-images.yaml` manifest (co-located with the Go code that consumes it) declares each image's source and digest. The Go client loads it via `go:embed` at package init and resolves image refs through a `mirrorImage()` lookup, making the manifest the single source of truth. `hack/mirror-images.sh` mirrors images via `crane copy`, using `QUAY_IO_USERNAME`/`QUAY_IO_PASSWORD` from the Prow push preset (Vault credentials are read-only and serve as a local-dev fallback only). A presubmit validator rejects malformed manifest entries (non-digest versions, sources with tags).

Fixes #

**What type of PR is this?**
/kind feature

```release-note
Tinkerbell provisioning now pulls all container images from quay.io/kubermatic-mirror instead of upstream registries, eliminating dependency on third-party registry availability.
```

**Documentation**:
```documentation
NONE
```


.prow/postsubmits.yaml
```diff
@@ -45,6 +45,25 @@ postsubmits:
...
 
+  - name: post-machine-controller-mirror-images
+    run_if_changed: "^hack/mirror-images\\.(yaml|sh)$"
+    branches:
+      - ^main$
+    decorate: true
+    clone_uri: "ssh://git@github.com/kubermatic/machine-controller.git"
+    path_alias: k8c.io/machine-controller
+    labels:
+      preset-vault: "true"
```
mgoltzsche, line 56: Unfortunately I realized that the image sync we worked on recently for the app charts (https://github.com/kubermatic/kubermatic/pull/15851 and https://github.com/kubermatic/kubermatic/pull/15839 and https://github.com/kubermatic/kubermatic/pull/15846) this PR is based on is not cutting it since the credentials within vault don't have write permissions to the Kubermatic registry anymore. (Apparently the main branch build of those PRs only succeeded because the image sync was skipped because the image already existed within the registry since somebody with more permissions must have pushed it previously.) I was going to fix that but then forgot about it due to another ticket I was assigned to.

The real quay.io registry write access fix seems to be to
* use the `preset-docker-push-kubermatic: "true"` preset.
* within the `validate-mirror-images.sh` script use the `QUAY_IO_USERNAME` and `QUAY_IO_PASSWORD` env vars that are injected by the preset to log into the registry (see [here](https://github.com/kubermatic/kubermatic/blob/6f8ead31020f30888c841f11a9ec528c72603d9f/hack/ci/release-images.sh#L82) for an example).

```suggestion
      preset-docker-mirror: "true"
      preset-docker-push-kubermatic: "true"
```

hack/mirror-images.sh
```diff
@@ -0,0 +1,133 @@
...
+# --- Registry login ----------------------------------------------------------
+login_registry() {
+  # allow caller to pass credentials directly (e.g. tests)
+  if [ -z "${REGISTRY_USER:-}" ] || [ -z "${REGISTRY_PASSWORD:-}" ]; then
+    login_vault
+    REGISTRY_USER="${REGISTRY_USER:-$(vault kv get -field=username dev/kubermatic-mirror-quay.io)}"
+    REGISTRY_PASSWORD="${REGISTRY_PASSWORD:-$(vault kv get -field=password dev/kubermatic-mirror-quay.io)}"
+  fi
+
+  crane auth login "${REGISTRY_HOST}" --username "${REGISTRY_USER}" --password-stdin <<< "${REGISTRY_PASSWORD}"
```
mgoltzsche: Since the registry credentials within Vault don't provide write access but the other preset does, let's login using the env vars it injects (e.g. see [here](https://github.com/kubermatic/kubermatic/blob/6f8ead31020f30888c841f11a9ec528c72603d9f/hack/ci/release-images.sh#L82)):
```suggestion
  if [ -z "${QUAY_IO_USERNAME:-}" ] || [ -z "${QUAY_IO_PASSWORD:-}" ]; then
    # Use read-only registry credentials from Vault in case the script is run locally for testing purposes.
    # Registry write access is only granted to the prow job which gets the QUAY_IO_* env vars injected by the preset.
    echo "WARNING: Using read-only $REGISTRY_HOST registry credentials from Vault - for local testing purposes only!"
    login_vault
    : "${QUAY_IO_USERNAME:=$(vault kv get -field=username dev/kubermatic-mirror-quay.io)}"
    : "${QUAY_IO_PASSWORD:=$(vault kv get -field=password dev/kubermatic-mirror-quay.io)}"
  fi

  crane auth login "${REGISTRY_HOST}" --username "${QUAY_IO_USERNAME}" --password-stdin <<< "${QUAY_IO_PASSWORD}"
```

mgoltzsche (commented):
I suggested a few changes to make the image push of the prow post-merge job work since the credentials within Vault don't provide write access anymore but read-only access.

Also, it would be great if the image references could be defined in one place as opposed to having to specify them in two files and ensuring that the two are kept in sync over time which is error-prone.
Can the Go code maybe embed the `mirror-images.yaml` and load the image references from there so that they can be maintained within a single YAML file?

.prow/postsubmits.yaml
mgoltzsche: Actually, given that the image should be pushed to `quay.io/kubermatic-mirror/images/*` as opposed to `quay.io/kubermatic/*` you may need to use the `preset-docker-mirror: "true"` preset. I updated my previous suggestion correspondingly now.

buraksekili: 👌 added preset-docker-mirror

hack/mirror-images.sh
buraksekili: done, updated the script. ptal again

buraksekili:
@mgoltzsche i also embedded YAML file as you suggested. let me know what you think

mgoltzsche (approved):
/lgtm
/approve

PR #2023, opened 2026-05-21
fix mirror-images 403 by dropping wrong-org push preset

### LINKED ISSUE #2018: Update images in Tinkerbell template

buraksekili:
Mirror images to `quay.io/kubermatic-mirror` under human-readable tags instead of digest-only destinations, making them browsable in the Quay UI. Adds a `tag` field to the manifest, a preflight verification script that validates tag-to-digest mappings upstream, and anti-tamper guards.

Fixes #

**What type of PR is this?**
/kind feature

The source pull remains digest-pinned; crane copy preserves the digest-addressable manifest so existing digest-based pulls keep working. The Go code in `pkg/mirror/mirror.go` is unaffected - it still constructs `@sha256:...` references.

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/mirror/mirror-images.yaml
+++ b/pkg/mirror/mirror-images.yaml
@@ -12,37 +12,40 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Images mirrored to quay.io/kubermatic-mirror/images/<key>@<version>.
+# Images mirrored to quay.io/kubermatic-mirror/images/<key>:<tag>.
 #
 # - `source`: upstream registry path. Must NOT include a tag or digest (no ":", no "@").
 # - `version`: must be a sha256 digest (sha256:<64-hex>). Tags, semver, "latest" forbidden.
-# - The comment above each entry records the human-readable tag the digest was resolved from
-#   (use `crane digest <source>:<tag>` to resolve). The comment is informational only.
+#   The upstream pull is always by digest (anti-tamper).
+# - `tag`: the human-readable tag that resolves to `version` upstream. The mirrored image
+#   is published under this tag in quay.io/kubermatic-mirror so it is browsable in the
+#   Quay UI. Format is whatever upstream uses (semver, git commit SHA, etc.) -- must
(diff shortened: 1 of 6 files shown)


pkg/mirror/mirror-images.yaml
```diff
@@ -12,37 +12,40 @@
...
+#   match the OCI tag regex [a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}.
 # - The map key (e.g. "tinkerbell/actions/cexec") becomes the destination path segment
 #   under quay.io/kubermatic-mirror/images/.
 
 images:
   alpine:
     source: docker.io/library/alpine
-    # alpine:3.23
+    tag: "3.23"
     version: sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
```
mgoltzsche, line 31: `3.23` is a floating tag which is meant to be used to implicitly use the latest patch release for that minor version.
Therefore the digest will deviate soon.
I understand that pinning the digest is beneficial security-wise but it may become inconsistent with the tag over time and also requires a bit more maintenance effort - both might be an acceptable trade-off.

As long as no digest is removed from the upstream registry the sync should be reproducible, though, given that the digest is used as copy source as opposed to the tag - which seems to be the case.

mgoltzsche (approved):
/lgtm
/approve

mgoltzsche:
cross-referencing https://github.com/kubermatic/kubermatic/pull/15839 and https://github.com/kubermatic/kubermatic/pull/15895

PR #2043, opened 2026-06-17
feat(openstack): add disablePortSecurity to MachineDeployment spec

### LINKED ISSUE #2012: OpenStack provider allow setting disablePortSecurity in MachineDeployment
# Problem

We cannot use a routing based Kubernetes CNI as OpenStack port security does prevent IP and MAC spoofing. Disabling the port security in OpenStack does allow Kubernetes CNI such as Cilium to use native routing.

Kubermatic's `machine-controller` does not support setting `disablePortSecurity`. Please implement this setting, which is already present in newer cluster-api.

# Reference

https://github.com/kubermatic/machine-controller/blob/1302066bdf88bd6bb99cac8525b957591a744be7/pkg/cloudprovider/provider/openstack/provider.go#L96-L99

mvanhorn:
## Summary

Adds a `disablePortSecurity` boolean to the OpenStack MachineDeployment provider spec. When set to `true`, the machine-controller disables port security on the instance's Neutron network ports after the instance comes up, and clears their security groups (OpenStack rejects security groups on a port with port security disabled). Default behavior is unchanged.

## Why this matters

Routing-based Kubernetes CNIs such as Cilium native routing rely on a node forwarding traffic for IP/MAC addresses other than the port's own. OpenStack port security blocks that traffic, so those CNIs cannot run on worker nodes unless port security is turned off on the instance ports. As described in #2012, the OpenStack provider had no way to do this, while cluster-api already exposes the equivalent setting. This change closes that gap so operators can provision routing-CNI-ready nodes directly from a MachineDeployment.

## What this PR does / why we need it

- New `DisablePortSecurity providerconfig.ConfigVarBool` field on `RawConfig`, resolved into `Config.DisablePortSecurity` in `getConfig`, mirroring the existing `trustDevicePath` / `configDrive` bool fields.
- After the server is created, when the flag is set, the provider waits for each attached network port to become active (reusing the existing port-readiness waiter) and updates it via Neutron to set `port_security_enabled=false`, clearing the port's security groups in the same call.
- The update is applied to every network in the `networks` list, not just the primary one, so the setting also takes effect on multi-NIC machines.
- When the field is omitted or `false`, no port update is performed and port security stays enabled.

## Testing

- `go test ./pkg/cloudprovider/provider/openstack/...` passes, including a new table test asserting `disablePortSecurity` defaults to `false` and resolves to `true` when set.
- `go build`, `go vet`, and `gofmt` are clean for the touched packages in both the main and `sdk` modules.

Fixes #2012

**What type of PR is this?**

/kind feature

The field is documented in `examples/openstack-machinedeployment.yaml`.

```release-note
Add a `disablePortSecurity` option to the OpenStack provider spec that disables port security (and clears security groups) on the instance's network ports, enabling routing-based CNIs such as Cilium native routing.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/examples/openstack-machinedeployment.yaml
+++ b/examples/openstack-machinedeployment.yaml
@@ -138,6 +138,10 @@ spec:
                   key: securityGroup
             # The machine won't get a floating ip if you leave this empty
             floatingIpPool: "ext-net"
+            # Optional. If true, disables OpenStack port security on the instance's
+            # network port (and clears its security groups). Required for routing-based
+            # CNIs such as Cilium native routing that need IP/MAC "spoofing".
+            disablePortSecurity: false
             # Only required if there is more than one AZ to choose from
             availabilityZone: ""
             # Only required if there is more than one network available
--- a/examples/openstack-machinedeployment.yaml
+++ b/examples/openstack-machinedeployment.yaml
@@ -139,7 +139,7 @@ spec:
             # The machine won't get a floating ip if you leave this empty
             floatingIpPool: "ext-net"
             # Optional. If true, disables OpenStack port security on the instance's
-            # network port (and clears its security groups). Required for routing-based
+            # network ports (and clears its security groups). Required for routing-based
             # CNIs such as Cilium native routing that need IP/MAC "spoofing".
             disablePortSecurity: false
             # Only required if there is more than one AZ to choose from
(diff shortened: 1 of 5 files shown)


kron4eg:
@mvanhorn thank for the PR! Looks good and solid! Please fix mentioned typos and I'll merge it.

kron4eg:
@mvanhorn ping.

PR #2047, opened 2026-07-01
feat(openstack): propagate instance faults to Machine status on ERROR

### LINKED ISSUE #2040: Propagate OpenStack instance faults to Machine status when server enters ERROR
## Current behavior:

When Nova places a server in ERROR before kubelet registers, the corresponding Machine object may keep an empty `.status`: 

```shell
k get machine -n kube-system 2xh100nvl-6978bf6c49-ccdtj -oyaml | yq .status
{}
```

The consumer components (e.g. kkp-api/dashboard) only see a provisioning `Machine` and the actual provider fault cannot surface unless the user has access to the underlying OpenStack project (not always the case): 

```shell
os server show f472b2c8-fd17-4759-afaf-ef732776d9cc -c status -c fault
+--------+-------------------------------------------------------------------------------------------------------------------------------+
| Field  | Value                                                                                                                         |
+--------+-------------------------------------------------------------------------------------------------------------------------------+
| fault  | {'code': 500, 'created': '2026-06-04T21:29:31Z', 'message': 'No valid host was found. There are not enough hosts available.'} |
| status | ERROR                                                                                                                         |
+--------+-------------------------------------------------------------------------------------------------------------------------------+
```

(... in this case, this error is due to a lack of hosts that can schedule such VM flavour - i.e. with 2xh100 GPUs.)

## Expected behavior:

If the OpenStack server enters `ERROR` state, the machine-controller should extract the instance's `fault` information and persist it to `Machine.status.errorReason` and `errorMessage`, even when no Kubernetes Node exists yet - as it happens, for instance with quota exhaustion: 

https://github.com/kubermatic/machine-controller/blob/3851738a329ed3c6178d2c6c172137d3bd566fd9/sdk/apis/machines/v1alpha1/types.go#L215-L217

The Dashboard can then surface provider-side provisioning failures such as "`No valid host was found. There are not enough hosts available.`" instead of showing an endless provisioning loop:

mvanhorn:
## Summary

Make the OpenStack provider's `Get` return a `TerminalError` (`Reason: CreateError`) when the UID-matched Nova server is in `ERROR` state, so the provider-side fault surfaces on `Machine.status` instead of leaving the machine stuck in an endless "provisioning" state. The message is composed from the server's `fault.message` (falling back to `fault.details`, then a generic string) and includes the server ID.

The existing controller wiring in `ensureInstanceExistsForMachine` already detects terminal errors and writes `Machine.status.errorReason` / `Machine.status.errorMessage`, so no controller change is needed - this mirrors how quota exhaustion already terminates. The ERROR check is skipped when the `Machine` is being deleted (`DeletionTimestamp` set) so `Cleanup` can still retrieve and delete a failed server.

## Motivation

When an OpenStack Nova server enters `ERROR` before the kubelet registers (for example a flavor/GPU shortage that yields "No valid host was found. There are not enough hosts available."), the machine-controller left the `Machine` object's status empty. Consumers (kkp-api / dashboard) only ever saw an endless "provisioning" state with no indication of the underlying cause. The provider-side fault was visible via `openstack server show` but never propagated to the `Machine`, so operators had no signal that the instance had failed and the reconcile loop requeued indefinitely.

## Testing

Added `TestGetInstanceErrorStatus`, which exercises a mocked `/servers/detail` endpoint across three cases: an ERROR server with a populated fault (terminal error carries the fault message and server ID), an ERROR server with an empty fault (non-empty generic terminal message, no panic), and an ERROR server whose `Machine` is being deleted (returns the instance so cleanup can proceed). The change is confined to the OpenStack provider and does not touch the shared `instance.Instance` interface or other providers. `go build`, `go vet`, and `go test` pass for `pkg/cloudprovider/provider/openstack/...`.

Fixes #2040

**What type of PR is this?**
/kind feature

```release-note
OpenStack: instance faults are now propagated to Machine status when a server enters an ERROR state, surfacing the provider-side failure reason instead of leaving the machine stuck in provisioning.
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/openstack/provider.go
+++ b/pkg/cloudprovider/provider/openstack/provider.go
@@ -914,11 +914,28 @@ func (p *provider) Get(_ context.Context, log *zap.SugaredLogger, machine *clust
 
 	for i, s := range allServers {
 		if s.Metadata[machineUIDMetaKey] == string(machine.UID) {
+			if s.Status == "ERROR" && machine.DeletionTimestamp == nil {
+				return nil, cloudprovidererrors.TerminalError{
+					Reason:  common.CreateMachineError,
+					Message: openStackInstanceErrorMessage(s),
+				}
+			}
 			return &osInstance{server: &allServers[i]}, nil
 		}
 	}
 
 	return nil, cloudprovidererrors.ErrInstanceNotFound
+}
+
+func openStackInstanceErrorMessage(s serverWithExt) string {
+	faultMsg := s.Fault.Message
+	if faultMsg == "" {
+		faultMsg = s.Fault.Details
+	}
+	if faultMsg == "" {
+		faultMsg = "instance entered ERROR state"
+	}
+	return fmt.Sprintf("OpenStack instance %s entered ERROR state: %s", s.ID, faultMsg)
 }
 
 func (p *provider) MigrateUID(_ context.Context, log *zap.SugaredLogger, machine *clusterv1alpha1.Machine, newUID types.UID) error {
--- a/pkg/cloudprovider/provider/openstack/provider_test.go
+++ b/pkg/cloudprovider/provider/openstack/provider_test.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 	"text/template"
 	"time"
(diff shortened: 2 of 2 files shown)


kron4eg:
@mvanhorn thank you for your contribution, appreciate it!

PR #2071, opened 2026-08-27
feat: added support for NVMe disk controller for Azure VMs

### LINKED ISSUE #1980: Azure v6-series VMs fail to provision: missing DiskControllerType configuration for NVMe controllers
## Issue

The machine-controller fails to provision Azure v6-series VMs (e.g., Standard_D4s_v6, Standard_D8s_v6) because it does not set the DiskControllerType field during VM creation. Azure v6-series VMs require NVMe disk controllers, but when DiskControllerType is not explicitly specified, Azure defaults to SCSI controllers, which are incompatible with these VM sizes. This results in VM creation failures with disk controller type mismatch errors.

## Customer Impact

This issue is currently blocking deploying worker nodes using Azure v6-series VM sizes. The v6-series offers better performance and cost optimization compared to older generations, but the inability to provision these VMs forces the customer to use older, less efficient VM sizes, impacting their infrastructure modernization plans and potentially increasing operational costs.

`machine-controller {"level":"error","time":"2025-12-09T10:49:25.550Z","logger":"machine-controller","caller":"machine/controller.go:388","msg":"Reconciling failed","machine":"kube-system/dataapp03bdsfurnace8-6c8db78958-rdl57","error":"failed to create machine at cloudprovider, due to trying to create a VM: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code=\"InvalidParameter\" Message=\"The VM size 'Standard_D4s_v6' cannot boot with OS image or disk. Please check that disk controller types supported by the OS image or disk is one of the supported disk controller types for the VM size 'Standard_D4s_v6'. Please query sku api at https://aka.ms/azure-compute-skus to determine supported disk controller types for the VM size.\" Target=\"vmSize\""}`

dharapvj:
With `v6` series of VMs, Azure has remove SCSI disk controller support and it now only provides NVMe disks. If we try to use any v6 or v7 machines e.g. `Standard_E2as_v7` we do not get new machines ready.

This PR adds support to specify NVMe as diskcontroller type which will fix the issue.

Fixes #1980

**What type of PR is this?**

/kind feature
/kind chore

I do not know how to do a real world test on Azure. So hopefully, we have e2e tests which will get fired with this PR to get such testing done.

```release-note
Azure VMs with series v6+ can be created via machine-controller
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/pkg/cloudprovider/provider/azure/provider.go
+++ b/pkg/cloudprovider/provider/azure/provider.go
@@ -22,6 +22,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -102,10 +104,11 @@ type config struct {
 	ImagePlan             *compute.Plan
 	ImageReference        *compute.ImageReference
 
-	OSDiskSize   int32
-	OSDiskSKU    *compute.StorageAccountTypes
-	DataDiskSize int32
-	DataDiskSKU  *compute.StorageAccountTypes
+	OSDiskSize         int32
+	OSDiskSKU          *compute.StorageAccountTypes
+	DataDiskSize       int32
+	DataDiskSKU        *compute.StorageAccountTypes
+	DiskControllerType *compute.DiskControllerTypes
 
 	AssignPublicIP              bool
 	PublicIPSKU                 *network.PublicIPAddressSkuName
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/provider/azure/provider.go
```diff
@@ -1428,6 +1452,34 @@ func skuSupportsGen2(sku compute.ResourceSku) bool {
...
+// v6 and later generation VMs exclusively support NVMe; the DiskControllerTypes capability
+// value is "NVMe" (no "SCSI") for those SKUs.
+func skuRequiresNVMe(sku compute.ResourceSku) bool {
+	if sku.Capabilities == nil {
+		return false
+	}
+	for _, cap := range *sku.Capabilities {
+		if cap.Name != nil && *cap.Name == "DiskControllerTypes" && cap.Value != nil {
+			v := strings.ToLower(*cap.Value)
+			return v == "nvme" || (strings.Contains(v, "nvme") && !strings.Contains(v, "scsi"))
```
kron4eg:
```suggestion
			return strings.Contains(v, "nvme") && !strings.Contains(v, "scsi")
```

if string equals to "nvme" then strings.Contains(v, "nvme") is also true.

kron4eg (approved):
/approve

dharapvj:
@kron4eg / @buraksekili - why do we not have a release/1.66 branch even though we have a release 1.66.0-1.66.2?

kron4eg:
@dharapvj for now v1.66 and main are the same thing

PR #2082, opened 2026-09-14
Add per-Machine eviction timeout override via annotation

### LINKED ISSUE #2070: Allow to override node drain timeout per machine deployment / machine

As a KKP administrator / KKP user cluster owner
I would like to have the ability to override machine-controllers global node eviction timeout (`--skip-eviction-after`) per `MachineDeployment` / `Machine`
in order to have fine-grained control after which time a node / machine is deleted on the `MachineDeployment` / `Machine` level

### Solution details

- machine-controller provides an option to override the globally configured node drain timeout per `MachineDeployment` / `Machine` (e.g. timeout could be specified with an annotation on the `Machine`, propagated down from the `MachineDeployment`)

### Alternative approaches

None

### Use cases

Right now, the machine-controller allows you to configure the node drain timeout only at a global level. However, there may be situations in which you have multiple `MachineDeployments`, each with different requirements for the rollout / deletion process.

Imagine you have two `MachineDeployments`, each intended for a specific type of workload (using taints, labels, and node label selectors), such as stateful and stateless workloads. In this case, the rollout behavior of the `MachineDeployment` for stateless workloads could be configured to be significantly more aggressive than the rollout behavior for stateful workloads.

### Additional information

/label customer-request
Internal reference: 9415

buraksekili:
Adds the `machine-controller.kubermatic.io/skip-eviction-after` annotation on Machines and MachineDeployments so the global `--skip-eviction-after` eviction timeout can be overridden per MachineDeployment / Machine, without changing the inherited CAPI API types.

Fixes #2070

**What type of PR is this?**
/kind feature

No changes to the vendored CAPI types, per review. `shouldEvict` resolves the threshold from the annotation (`time.ParseDuration`), falling back to the global flag when unset or unparseable (invalid values are logged). MachineDeployment object annotations reach MachineSet objects through the existing annotation sync, and the MachineSet controller copies this one annotation onto Machines it creates, so changing the value never rolls existing Machines. `kubermatic.io/skip-eviction` on Nodes keeps its separate skip-entirely meaning.

```release-note
Added the `machine-controller.kubermatic.io/skip-eviction-after` annotation to override the global `--skip-eviction-after` eviction timeout per MachineDeployment or Machine; unset or invalid values fall back to the global flag.
```

**Documentation**:
```documentation
NONE
```


sdk/apis/cluster/v1alpha1/machine_types.go
```diff
@@ -105,6 +105,14 @@ type MachineSpec struct {
...
 	// +optional
 	ProviderID *string `json:"providerID,omitempty"`
+
+	// NodeDrainTimeout is the maximum duration the machine-controller will keep
+	// trying to evict pods from the corresponding node after this Machine is
+	// deleted. When unset or zero, the global --skip-eviction-after flag value
+	// applies. The timer starts at the Machine's deletionTimestamp, matching the
+	// global flag. Note: NodeDrainTimeout is different from `kubectl drain --timeout`.
+	// +optional
+	NodeDrainTimeout *metav1.Duration `json:"nodeDrainTimeout,omitempty"`
```
kron4eg: oh no, please not in the CAPI, we don't change CAPI as we don't own it.

buraksekili: ergh, our fork is out of sync. this field has been in CAPI for a while actually. let me move it to an annotation if that sounds ok to you?

kron4eg (approved):
/approve

PR #2083, opened 2026-09-21
Allow disabling the eviction timeout

### LINKED ISSUE #2069: Allow to disable forceful deletion of machines

As a KKP administrator / KKP user cluster owner
I would like to have the ability to completely disable the machine-controllers node eviction timeout (`--skip-eviction-after`)
in order to have full (manual) control after which time a node / machine is deleted

### Solution details

- machine-controller provides an option to completely disable the forceful deletion of machines (e.g. Cluster API allows you to set [`Machine.spec.nodeDrainTimeout` to 0](https://github.com/kubernetes-sigs/cluster-api/blob/v1.14.0/docs/book/src/tasks/automated-machine-management/machine_deletions.md?plain=1#L17) to specify that the node can be drained without any time limitations)

### Alternative approaches

- set `--skip-eviction-after` to the highest possible value

### Use cases

Right now, the machine-controller will eventually forcefully delete a machine, and the associated node after the configured node drain timeout (`--skip-eviction-after`) expired, and there is no way to disable this behavior. However, there may be situations in which the cluster administrator wants to have full control over the machine deletion process to prevent any involuntary disruptions to workloads running on that machine / node.

### Additional information

/label customer-request
Internal reference: 9415

ahmadhamzh:
Setting `--skip-eviction-after` or the `machine-controller.kubermatic.io/skip-eviction-after` annotation to `0` now disables the eviction timeout, so a Machine is never deleted while Pods are still running on it.

Fixes #2069

**What type of PR is this?**
/kind feature

```release-note
Setting the `machine-controller.kubermatic.io/skip-eviction-after` annotation to `0` now disables the eviction timeout; in standalone deployments `--skip-eviction-after=0` does the  same.
action required: `0` previously skipped eviction immediately and force-deleted the machine.
```

**Documentation**:

```documentation
TBD
```


buraksekili (commented):
went through this in detail. the flip itself looks right and the three new behaviors all have cases in the table. two things worth acting on: the kkp webhooks still reject `0`, so the flag path is unreachable for kkp users until the paired kubermatic change lands, and the readme sentence promises more than the controller keeps. details inline.

cmd/machine-controller/main.go
```diff
@@ -176,7 +176,7 @@ func main() {
 	flag.StringVar(&joinClusterTimeout, "join-cluster-timeout", "", "when set, machines that have an owner and do not join the cluster within the configured duration will be deleted, so the owner re-creates them")
 	flag.StringVar(&bootstrapTokenServiceAccountName, "bootstrap-token-service-account-name", "", "When set use the service account token from this SA as bootstrap token instead of creating a temporary one. Passed in namespace/name format")
 	flag.BoolVar(&profiling, "enable-profiling", false, "when set, enables the endpoints on the http server under /debug/pprof/")
-	flag.DurationVar(&skipEvictionAfter, "skip-eviction-after", 2*time.Hour, "Skips the eviction if a machine is not gone after the specified duration.")
+	flag.DurationVar(&skipEvictionAfter, "skip-eviction-after", 2*time.Hour, "Skips the eviction if a machine is not gone after the specified duration. Set to 0 to disable the timeout and never give up on the eviction.")
```
buraksekili, line 179: one thing on the kkp side: both webhooks still reject the value this pr exists to allow. [pkg/validation/cluster.go#L1426-L1427](https://github.com/kubermatic/kubermatic/blob/main/pkg/validation/cluster.go#L1426-L1427) and [pkg/webhook/seed/validation.go#L248-L250](https://github.com/kubermatic/kubermatic/blob/main/pkg/webhook/seed/validation.go#L248-L250) in kubermatic keep refusing `skipEvictionAfter <= 0` with "must be a positive duration", so for kkp users the flag path can't actually reach this new behavior. the deployment rendering itself would pass `0` through fine (`deployment.go:251-253` appends `-skip-eviction-after`), it's just the validation gating it.

maybe worth pairing this with a small kubermatic pr that allows exactly `0` (still rejecting negatives) and refreshes the guard comments + the crd field descriptions? until that lands the annotation is the only kkp path, so we could also scope the release note's flag sentence to standalone deployments for now.

README.md
```diff
@@ -143,6 +143,8 @@ metadata:
 
 The value must be a valid duration string (for example `30m`, `1h`, `90s`). If the value cannot be parsed, the controller logs it and falls back to the global flag.
 
+Setting the value to `0` disables the timeout entirely, The Machine then stays in `Terminating` until eviction succeeds, so an operator has to unblock it, for example by relaxing the PodDisruptionBudget that rejects the eviction.
```
buraksekili: this sentence promises a bit more than the controller keeps: eviction is still skipped and deletion proceeds when the node is `NotReady`/`Unknown` (`shouldEvict`, around controller.go:578-581) or when no schedulable target exists (around controller.go:613, think single-node cluster). so the machine waits as long as the node is Ready and the pods have somewhere to go, not until eviction succeeds.

maybe reword to something like "with the timeout disabled, the Machine waits for eviction as long as its node is Ready and a schedulable target node exists"? also a small comma splice here ("entirely, The") that could become its own sentence in the same edit.

pkg/controller/machine/controller.go
```diff
@@ -547,7 +548,9 @@ func (r *Reconciler) shouldEvict(ctx context.Context, log *zap.SugaredLogger, ma
 		}
 	}
 
-	if machine.DeletionTimestamp != nil && time.Since(machine.DeletionTimestamp.Time) > threshold {
+	if threshold == 0 {
+		log.Debugw("Eviction timeout is disabled, not skipping eviction", "threshold", threshold)
```
buraksekili, line 556: 🤔 pre-existing thing i bumped into while reading this: `time.ParseDuration` accepts negative values, so an annotation like `-30m` parses successfully, and then every deleting Machine counts as past the threshold in this comparison: eviction is skipped immediately and the node is force-deleted, the opposite of what whoever wrote `-30m` wanted. the invalid-value fallback never fires because the parse succeeds.

maybe a parsed duration below zero could take the same fallback as the unparseable case, keeping the flag value? fine as a follow-up since it's not introduced here.

pkg/controller/machine/controller_test.go
```diff
@@ -439,6 +442,92 @@ func TestControllerShouldEvict(t *testing.T) {
...
+				},
+				Status: clusterv1alpha1.MachineStatus{
+					NodeRef: &corev1.ObjectReference{Name: "existing-node"},
+				},
+			},
+		},
+		{
+			name:              "eviction possible since the global flag disables the timeout",
+			shouldEvict:       true,
+			skipEvictionAfter: ptr.To(time.Duration(0)),
```
buraksekili: nit: the new cases reach for `ptr.To(time.Duration(0))` while the file already has `durationPtr` for the same conversion (controller_test.go:333-335, used a few times above). can we use that here and drop the `k8s.io/utils/ptr` import, or go the other way and replace `durationPtr` with `ptr.To` throughout so one mechanism remains?

pkg/controller/machine/controller_test.go
```diff
@@ -439,6 +442,92 @@ func TestControllerShouldEvict(t *testing.T) {
 				},
 			},
 		},
+		{
+			name:        "eviction possible since the per-machine override disables the timeout",
```
buraksekili, line 445: 👌 nice coverage: annotation-0, flag-0 and annotation-overriding-a-disabled-flag each got a case in the table

cmd/machine-controller/main.go
ahmadhamzh: i'll open the kkp pr to allow 0 in both validators. 
happy to either hold this until that merges, or land it now with the release note scoped to standalone deployments as you suggested — which do you prefer?

buraksekili:
opened a tracking issue for the kkp side: kubermatic/kubermatic#16531 (allow `skipEvictionAfter: 0` in both webhooks, refresh the guard comment and CRD descriptions). the annotation stays the only kkp route until that lands

cmd/machine-controller/main.go
buraksekili: https://github.com/kubermatic/machine-controller/pull/2083#issuecomment-5775201679

buraksekili (approved):
/approve

thanks @ahmadhamzh !

